PR #3, opened 2019-05-03
Add numeric and free text question types

robinmessage:
Also adds constants.units to the store.

What tests should I be doing for this do you think? I'll look at those on Wednesday.


robinmessage:
I need to include a change in this once MathJax is merged.

PR #4, opened 2019-05-07
Login

bh412:
Adds the ability to login with email/password and redirects using history


bh412:
Looks good, thanks for checking!

PR #10, opened 2019-05-15
Account update

bh412:
- Adds name/dob/email/password changing functionality to my account page, with error cases
- Adds reset password request handling, email directs to password choosing page, validation/error cases
- Adds email address alteration handling, email directs to confirmation page, validation/error cases - doesn't check for non-segue account in front end
- Adds some ids for selenium testing


src/IsaacApiTypes.tsx
```diff
@@ -450,6 +450,12 @@ export interface Address {
     country?: string;
 }
 
+export interface userPreferences {
+    BETA_FEATURE?: {};
+    EMAIL_PREFERENCE?: {};
+    SUBJECT_INTEREST?: {};
+}
```
mlt47: API types are auto-generated by the API so any additions will be overwritten when they next get generated.
This should then be moved to APP types (which is not auto-generated).
I've been meaning to rename this file so that it is clearer that it is auto-generated!

src/app/state/actions.tsx
```diff
@@ -1,11 +1,17 @@
...
-import {AuthenticationProvider, ChoiceDTO, QuestionDTO} from "../../IsaacApiTypes";
+import {AuthenticationProvider, ChoiceDTO, QuestionDTO, RegisteredUserDTO} from "../../IsaacApiTypes";
 import {ACTION_TYPE, DOCUMENT_TYPE, TAG_ID} from "../services/constants";
 import {AppState} from "./reducers";
 import history from "../services/history";
 
+
+interface ValidationUser extends RegisteredUserDTO {
+    password: string | null;
+}
```
mlt47: IsaacAppTypes feels like a better home for this, but as it is only being used in this file it doesn't make too much of a difference

src/app/components/pages/MyAccount.tsx
```diff
@@ -1,21 +1,90 @@
...
+    const myUser1 = Object.assign(tempUser, user);
+    const myUser = Object.assign(myUser1, {password: ""});
+    const [isValidEmail, setValidEmail] = useState(true);
+    const [isValidDob, setValidDob] = useState(true);
+    const [isValidPassword, setValidPassword] = useState(true);
+    const [currentPassword, setCurrentPassword] = useState("");
 
-interface AccountPageProps {user: RegisteredUserDTO | null}
+    let today = new Date();
+    let thirteen_years_ago = Date.UTC(today.getFullYear() - 13, today.getMonth(), today.getDate())/1000;
```
mlt47: Better to use camel case for this variable

src/app/components/pages/MyAccount.tsx
```diff
@@ -1,21 +1,90 @@
...
+    user: RegisteredUserDTO | null
+    updateCurrentUser: (
+        params: {registeredUser: validationUser; passwordCurrent: string},
+        currentUser: RegisteredUserDTO
+    ) => void
+    errorMessage: string | null
+}
+
+const AccountPageComponent = ({user, updateCurrentUser, errorMessage}: AccountPageProps) => {
+    const updateDetails = () => console.log("Account updated");
```
mlt47: I don't think this should be needed anymore. It seems to be used as the form's onSubmit action. So try pressing enter after filling in a field in the form, it should actually update the account and not just print to the console that it has.

src/app/components/pages/MyAccount.tsx
```diff
@@ -1,21 +1,90 @@
...
+
+    const emailPreferences = {"NEWS_AND_UPDATES": true, "ASSIGNMENTS": true, "EVENTS": true};
+
+    const tempUser = {};
+    const myUser1 = Object.assign(tempUser, user);
+    const myUser = Object.assign(myUser1, {password: ""});
```
mlt47: These three lines can be replaced with     `const myUser = Object.assign({}, user, {password: ""});` I think.

src/app/components/pages/MyAccount.tsx
```diff
@@ -1,21 +1,90 @@
...
+
+    const validateAndSetPassword = (event: any) => {
+        setValidPassword(
+            (event.target.value == (document.getElementById("password") as HTMLInputElement).value) &&
+                ((document.getElementById("password") as HTMLInputElement).value != undefined) &&
+                ((document.getElementById("password") as HTMLInputElement).value.length > 5)
+        )
+    };
 
     {/• TODO handle #... in with react-router? •/}
```
mlt47: Do you remember what this TODO was referring to?

src/app/components/pages/MyAccount.tsx
```diff
@@ -56,96 +125,151 @@ const AccountPageComponent = ({user}: AccountPageProps) => {
                     </Nav>
```
mlt47: Can use numbers for setting active tab rather than string representations of numbers (allows for tab + 1 etc.)

src/app/components/pages/MyAccount.tsx
```diff
@@ -56,96 +125,151 @@ const AccountPageComponent = ({user}: AccountPageProps) => {
...
                     <TabContent activeTab={activeTab}>
                         <TabPane tabId="0">
+                            <CardBody>
+                                <Form name="userDetails" onSubmit={updateDetails}>
+                                    <Row>
+                                        <Col size={12} md={6}>
```
mlt47: We don't need the size={12} as this is default. It was probably copying my code that caused you to do this!

src/app/components/pages/MyAccount.tsx
```diff
@@ -56,96 +125,151 @@ const AccountPageComponent = ({user}: AccountPageProps) => {
...
+                                            <FormGroup>
+                                                <Label htmlFor="first-name-input">First Name</Label>
+                                                <Input id="first-name-input" type="text" name="givenName"
+                                                       defaultValue={myUser.givenName}
+                                                       onBlur={(e: any) => {myUser.givenName = e.target.value}}
```
mlt47: Will this be set even if you pres enter to submit the value? I'm not sure, but it's worth checking.

src/app/state/actions.tsx
```diff
@@ -17,6 +23,39 @@ export const requestCurrentUser = () => async (dispatch: Dispatch<Action>) => {
...
 };
 
+
+export const requestUserPreferences = () => async (dispatch: Dispatch<Action>) => {
+
+};
+
+export const updateCurrentUser = (params: {registeredUser: ValidationUser; passwordCurrent: string}, currentUser: RegisteredUserDTO) => async (dispatch: Dispatch<Action>) => {
+    if (currentUser.email !== params.registeredUser.email) {
+        let emailChange = window.confirm("You have edited your email address. Your current address will continue to work until you verify your new address by following the verification link sent to it via email. Continue?");
```
mlt47: This is fine for now, but lets add a TODO here to handle the alert ourselves rather than use the window.confirm alert.

src/app/state/actions.tsx
```diff
@@ -17,6 +23,39 @@ export const requestCurrentUser = () => async (dispatch: Dispatch<Action>) => {
...
+        } else {
+            params.registeredUser.email = currentUser.email;
+        }
+    } else {
+        dispatch({type: ACTION_TYPE.USER_DETAILS_UPDATE});
+        try {
+            const currentUser = await api.users.updateCurrent(params);
+            dispatch({type: ACTION_TYPE.USER_DETAILS_UPDATE_SUCCESS});
+            history.push('/');
+            history.go(0);
```
mlt47: Have a go at trying to remove the duplication between here and the code block above

bh412:
Issue on account page refresh, makes api calls but doesn't re-render with user update

src/app/components/handlers/PasswordResetHandler.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+interface PasswordResetHandlerProps {
+    handleResetPassword: (params: {token: string | null; password: string | null}) => void;
+    verifyPasswordReset: (token: string | null) => void;
+    errorMessage: ErrorState;
+}
+
+const ResetPasswordHandlerComponent = ({handleResetPassword, verifyPasswordReset, errorMessage}: PasswordResetHandlerProps) => {
+    const resetToken = window.location.href.substr(window.location.href.lastIndexOf('/') + 1);
```
mlt47: Can you find a way of doing this which is less brittle?

src/app/components/handlers/PasswordResetHandler.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const validateAndSetPassword = (event: any) => {
+        setValidPassword(
+            (event.target.value == (document.getElementById("password") as HTMLInputElement).value) &&
+            ((document.getElementById("password") as HTMLInputElement).value != undefined) &&
+            ((document.getElementById("password") as HTMLInputElement).value.length > 5)
+        )
+    };
+
+    useEffect(() => {
+        setTimeout(function(){verifyPasswordReset(resetToken)},10);
```
mlt47: Why doe this require a setTimeout?

PR #13, opened 2019-05-17
User consistency checker

robinmessage:
Force a logout and clear of state if the user changes (for example, by a logout or login in another tab)


src/app/components/navigation/ConsistencyErrorModal.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import * as RS from "reactstrap";
+
+export const ConsistencyErrorModal = ({consistencyError}: {consistencyError: boolean}) => {
+    return <RS.Modal isOpen={consistencyError}>
```
robinmessage: Once an error occurs, they can't close this modal - does that seem right to you @mlt47 ?

mlt47: Seems sensible

PR #17, opened 2019-05-21
Contact us

bh412:
Adds pre-configureable contact page at /contact using urlParams along with required field star and potential connection error message


src/app/components/pages/Contact.tsx
```diff
@@ -0,0 +1,205 @@
...
+                                    </div>
+                                    <Row>
+                                        <Col size={12} md={6}>
+                                            <span className="form-required color=$secondary">Required field</span>
+                                        </Col>
+                                        <Col size={12} md={6} className="text-right">
+                                            <Button color="secondary" type="submit" onClick={() => {
+                                                setFirstName(() => (document.getElementById("first-name-input") as HTMLInputElement).value);
+                                                setLastName(() => (document.getElementById("last-name-input") as HTMLInputElement).value);
+                                                setEmail(() => (document.getElementById("email-input") as HTMLInputElement).value);
```
mlt47: You shouldn't need to get this information like this, all the information is already in the firstName lastName email state

src/app/components/pages/Contact.tsx
```diff
@@ -0,0 +1,205 @@
...
+                                <CardFooter>
+                                    <div>
+                                        <Alert color="danger" isOpen={errorMessage}>{errorMessage} You can contact us at <a href="mailto:webmaster@isaaccomputerscience.org">webmaster@isaaccomputerscience.org</a></Alert>
+                                    </div>
+                                    <Row>
+                                        <Col size={12} md={6}>
+                                            <span className="form-required color=$secondary">Required field</span>
+                                        </Col>
+                                        <Col size={12} md={6} className="text-right">
+                                            <Button color="secondary" type="submit" onClick={() => {
```
mlt47: You don't actually need this, it is better to just change it into a Input component of type submit with no onClick actions so that what it does is consistent with what is done if you press enter to submit the form.

bh412:
Corrections made

PR #23, opened 2019-05-24
Adds page not found to document/topic request

bh412:


src/app/components/pages/Topic.tsx
```diff
@@ -15,16 +15,16 @@ const stateToProps = (state: AppState, {match: {params: {topicName}}}: any) => (
...
     topicPage: state ? state.currentTopic : null
 });
-const actionsToProps = {fetchTopicDetails};
+const actionsToProps = {fetchTopic};
 
 interface TopicPageProps {
     topicName: TAG_ID;
     topicPage: IsaacTopicSummaryPageDTO | null;
-    fetchTopicDetails: (topicName: TAG_ID) => void;
+    fetchTopic: (documentType: DOCUMENT_TYPE, topicId: TAG_ID) => void;
```
mlt47: We can reduce the code even further by getting fetchDocument here. Then we can delete fetchTopic altogether which would be pretty cool.

mlt47: I'm sorry, I lied! We want the Topic to be stored in a separate place to any document that we load.

PR #25, opened 2019-05-29
Mobile improvements

bh412:
- Closes menu on item click for mobile
- Adds logout option to top of my account page on mobile


src/app/components/navigation/NavigationBar.tsx
```diff
@@ -14,7 +14,7 @@ interface NavigationBarProps {
 }
 
 const NavigationBarComponent = ({user}: NavigationBarProps) => {
-    const [menuOpen, setMenuOpen] = useState(false);
+    const [menuOpen, setMenuOpen] = useState(window.innerWidth>=768);
```
jsharkey13: Are these hard-coded numbers wise? Surely the whole point is that bootstrap is supposed to do all this for us?

bh412: todo added following conversation with Meurig

PR #26, opened 2019-05-29
Sign up

bh412:
Adds the ability to signup
Takes new user to my account page
Takes email/password from login form to register page
Includes error presentation


src/app/components/pages/LogIn.tsx
```diff
@@ -117,8 +125,8 @@ const LogInPageComponent = ({handleProviderLoginRedirect, logInUser, resetPasswo
                                     />
                                 </Col>
                                 <Col sm={6}>
-                                    <Button tag={Link} to="/register" color="primary" className="mb-2" outline block>
-                                        Sign up
+                                    <Button id="sign-up" color="primary" className="mb-2" onClick={(e: any) => signUp(e)} outline block>
```
mlt47: Cool

src/app/components/pages/Registration.tsx
```diff
@@ -13,17 +13,80 @@ import {
...
+    const [isValidEmail, setValidEmail] = useState(true);
+    const [isDobValid, setIsDobValid] = useState(true);
+    const [isValidPassword, setValidPassword] = useState(false);
+    const [currentPassword, setCurrentPassword] = useState("");
+    const [signUpAttempted, setSignUpAttempted] = useState(false);
+
+    const validateAndSetPassword = (password: string) => {
+        setCurrentPassword(password);
+        setValidPassword(
+            (password == (document.getElementById("password") as HTMLInputElement).value) &&
```
mlt47: Try to avoid having to use document.getElementById(....) in react, there is usually a way to useState(...) which is better

src/app/components/pages/Registration.tsx
```diff
@@ -13,17 +13,80 @@ import {
...
+            validatePassword(password)
+        )
+    };
+
+    const attemptSignUp = () => {
+        setSignUpAttempted(true);
+    };
+
+    useEffect(() => {
```
mlt47: useMemo rather than useEffect as in the account page

PR #38, opened 2019-06-03
Logic questions integration

Morpheu5:
- [x] Supports available symbols (letters only)
- [x] Exam boards syntax support
- [x] `/equality`


src/app/components/elements/InequalityModal.tsx
```diff
@@ -93,9 +96,12 @@ export class InequalityModal extends React.Component<InequalityModalProps> {
...
-            console.log("New editor state: ", s);
+            const modal = document.getElementById('inequality-modal');
+            if (modal) {
+                this.setState({ editorState: s });
+                this.props.onEditorStateChange(s);
+                console.log("New editor state: ", s);
+            }
         };
         sketch.onCloseMenus = () => { this.setState({ menuOpen: false }) }; // TODO Maybe nice to have
         sketch.isUserPrivileged = () => true; // TODO Integrate with currentUser object
```
mlt47: Is user privileged important now? We can have access to the user form connecting the component to the Redux state store.

src/app/components/elements/InequalityModal.tsx
```diff
@@ -183,50 +189,54 @@ export class InequalityModal extends React.Component<InequalityModalProps> {
...
+        let menu: JSX.Element =
+        <nav className="inequality-ui">
+            <div className={"inequality-ui menu-bar" + (this.state.menuOpen ? " open" : " closed")}>
+                {/* {this.state.activeMenu == "numbers" && <div className="top-menu numbers">
+                    <ul className="sub-menu">{
+                        "0123456789".split("").map((n, i) => {
+                            return this.menuItem(new MenuItem("Num", { significand: n }, { label: n, texLabel: true}), i);
+                        })
+                    }</ul>
+                </div>} */}
```
mlt47: Might want to clear up commented out code and/or any remaining console logs.

src/app/services/examBoard.ts
```diff
@@ -1,8 +1,12 @@
 import {EXAM_BOARD} from "./constants";
 import {UserPreferencesDTO} from "../../IsaacAppTypes";
+import { user } from "../state/reducers";
 
 export const determineExamBoardFrom = (userPreferences: UserPreferencesDTO | null) => {
-    return (userPreferences && userPreferences.EXAM_BOARD && userPreferences.EXAM_BOARD.AQA) ?
-        EXAM_BOARD.AQA :
-        EXAM_BOARD.OCR
+    if (!userPreferences) {
```
mlt47: Would something like this be clearer?
``` java
    if (userPreferences && userPreferences.EXAM_BOARD && userPreferences.EXAM_BOARD.AQA) {
        return EXAM_BOARD.AQA;
    } else {
        return EXAM_BOARD.OCR;
    }
```

jsharkey13:
@Morpheu5 - If you can make the changes Meurig suggested, and fix the merge conflict, I think this is good to go in.

src/app/components/elements/InequalityModal.tsx
Morpheu5: I don't think it's relevant to CS, it was relevant to Physics for derivatives and differentials, which is why I'm setting it to true — but could have set it to false just as well.

src/app/services/examBoard.ts
Morpheu5: Maybe? As long as it selects OCR by default is nothing else is selected…

Morpheu5:
@jsharkey13 remind me tomorrow if I forget. This should be 15 minutes tops.

PR #41, opened 2019-06-06
Feature/email verification fix

bh412:


src/app/state/actions.tsx
```diff
@@ -204,6 +204,7 @@ export const handleEmailAlter = (params: ({userid: string | null; token: string
     try {
         dispatch({type: ACTION_TYPE.EMAIL_AUTHENTICATION_REQUEST});
         const response = await api.email.verify(params);
+        dispatch(requestCurrentUser() as any)
```
mlt47: I'm pretty sure this is unnecessary, what you need to do is change the reducer to handle your new action because EMAIL_AUTHENTICATION_RESPONSE_SUCCESS already has the new user.

bh412: I've altered the reducer now and taken this out of here. It seems necessary to put it in the place I now call it as otherwise the user is loaded before the email address is verified

PR #46, opened 2019-06-12
Add Nomensa carousel and three fake events to demonstrate it

robinmessage:


robinmessage:
Self-merging on account of it just being an integration of Nomensa's code and @jsharkey13 being happy with it.

PR #47, opened 2019-06-12
Parsons problems

Morpheu5:


jsharkey13:
Otherwise, I've not reviewed the code, but the result looks good to me!

Morpheu5:
> * I'd add a note for users on small screens that if they're dragging using a touchscreen device they should press and hold to move items around. On desktop and mobile, I'd also add some text about dragging the items: "Drag items across to build your answer", maybe in grey in the "Your answer" box - in the same way we say "Click to enter your answer" on symbolic questions?

"Drag items across" text added. Not adding long-tap hint as that's expected behaviour on mobile so users should know that already.

> * Is there a reason it jumps by half the height of the dragged item when it is dropped into the page? The border box is also too short when adding an item at the bottom, but jumps to be the right size later on; I think this is the same problem?

Reason: indentation works against the library. Library expects to drop flush left, indentation breaks this. Not sure about border, but there's hardly anything that can be done to customise how this library works.

> * The accessibility works quite well with a keyboard, but there's no support for indentation. It would be good to add this, but maybe it can wait; so long as it won't require a change of library. In an ideal world, we'd hijack their keyboard sideways arrow event, and increment/decrement the indentation in the range 0 to 3, then pass the event on to the library handling if it would go outside that range?

[TODO]

> * I think we should limit the number of indent levels to 3 (so 4 states, including the no-indent case). Otherwise someone will be tempted to write a question with more, and that would be awful!

Done.

> * Is it possible to highlight the box you're about to drop the item into, in some way? Maybe a slight colour change? Just to make it obvious that that region is good to drop them into.

Not that I could see. [WOULDBENICE]

src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -0,0 +1,230 @@
...
+                    return !found;
+                });
+            }
+            // WARNING: Inverting the order of the arrays breaks this.
+            // TODO: Investigate if there is a method that gives more formal guarantees.
+            let diff = _differenceBy(prevState.availableItems, availableItems, 'id');
+            // This stops re-rendering when availableItems have not changed from one state update to the next.
+            // The set difference is empty if the two sets contain the same elements (by 'id', see above).
+            if (diff.length > 0) {
+                this.setState({ availableItems });
```
mlt47: I don't think available items need to be stored as "state", removing it might reduce the logic in the code and should reduce calls to render.

PR #48, opened 2019-06-12
Add working minimal content errors page

jsharkey13:
Still needs work on table width, sorting by column and may need some React clean-up since I was learning as I went.


src/app/components/pages/AdminContentErrors.tsx
```diff
@@ -0,0 +1,92 @@
...
+};
+const dispatchToProps = {getAdminContentErrors};
+
+interface AdminContentErrorsPageComponentProps {
+    errors: AdminContentErrorsState;
+    getAdminContentErrors: () => void;
+}
+
+const contentErrorDetailsListItem = (errorDetailsListItem: string) => {
+    return <li>{errorDetailsListItem}</li>
```
robinmessage: These ought to have a `key` property in general, but it isn't easy to generate one here, and they aren't gonna change so React won't need the key so it's fine.

PR #52, opened 2019-06-18
Adds scrollable fixed size school search dropdown

bh412:


jsharkey13:
I don't like that this isn't the same width as the text box, although it's quite hard to manage that I think. The scrollbars also look very strange and get in the way.
It's not accessible (it wasn't before this change, either) and you cannot use the keyboard to choose the school.

PR #56, opened 2019-06-19
Feature/figure numbers

daviesian:
Figure numbering now works for everything on the Regression Test Question page. There may be cases I've missed, in which case there's a description of how and where to add them here: https://github.com/isaacphysics/isaac-cs-app/commit/8b31464090fe473a81ef28abc19e1dfbce2db5a1#diff-704c339ab6c0fa2c9160c939c24c8c87R30

I'm reasonably happy with the overall React/Typescript cleanliness here, apart from the slightly icky `doc: any` in my `walk` function. I'd like to make that a `ContentDTO | QuestionDTO | ...`, but that doesn't help - Typescript still doesn't like my attempt to "Walk `doc.hints` if they exist", which is probably fair. Anyway, life's too short. I'm content for this to be merged if it looks sane to everyone else.


src/IsaacAppTypes.tsx
```diff
@@ -326,3 +327,9 @@ export interface ContentErrorsResponse {
     failedFiles: number;
     totalErrors: number;
 }
+
+export interface FigureNumbersById {[figureId: string]: number}
+
+// TODO: Work out how to initialise the context to {} with the right type, instead of these two silly lines.
+let defaultFigureNumbers: FigureNumbersById = {};
+export const FigureNumberingContext = React.createContext( defaultFigureNumbers);
```
robinmessage: `createContext<FigureNumbersById>` should do the trick

PR #62, opened 2019-06-24
Alters button order and url formatting for sharing

bh412:


mlt47:
> If possible, the `user-select: all;` CSS property on the `span` might be nice, but I don't know where the CSS should go @mlt47?

Ah, that's the CSS property I was thinking of earlier. I think adding `user-select-all` as a class to typography.scss would be useful.

PR #64, opened 2019-06-24
Actions, toasts and generic errors

robinmessage:


src/app/state/actions.tsx
```diff
@@ -640,18 +689,24 @@ export const selectGroup = (group: UserGroupDTO | null) => async (dispatch: Disp
...
-    const newGroup = await api.groups.create(groupName);
-    dispatch({type: ACTION_TYPE.GROUPS_CREATE_RESPONSE_SUCCESS, newGroup: newGroup.data});
-    return newGroup.data as AppGroup;
+    try {
+        const newGroup = await api.groups.create(groupName);
+        dispatch({type: ACTION_TYPE.GROUPS_CREATE_RESPONSE_SUCCESS, newGroup: newGroup.data});
+        return newGroup.data as AppGroup;
+    } catch (e) {
+        dispatch(showErrorToastIfNeeded("Creating a Group Failed", e));
+        throw e;
```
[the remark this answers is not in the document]
robinmessage: Yes, as we actually use the return value of the action. Thinking about it, the action could probably check th return value more carefully.

src/app/state/actions.tsx
```diff
@@ -527,10 +554,31 @@ export const deregisterQuestion = (questionId: string) => (dispatch: Dispatch<Ac
...
-    const response = await api.questions.answer(questionId, attempt);
-    dispatch({type: ACTION_TYPE.QUESTION_ATTEMPT_RESPONSE_SUCCESS, questionId, response: response.data});
-    // TODO MT handle response failure with a timed canSubmit
+    try {
+        dispatch({type: ACTION_TYPE.QUESTION_ATTEMPT_REQUEST, questionId, attempt});
+        const response = await api.questions.answer(questionId, attempt);
+        dispatch({type: ACTION_TYPE.QUESTION_ATTEMPT_RESPONSE_SUCCESS, questionId, response: response.data});
+    } catch (e) {
+        if (e.response.status == 429) {
+            const pause = 5 * 60 * 1000;
```
mlt47: After talking this through with James, this is not the correct value for the pause. The pause value is determined by the API and we don't (currently) send that value in the response so I think we revert to not showing the unlock time to the user. We do get the odd email asking "when will I be able to submit again?" but we would probably get even more emails if we show them a time and it still doesn't work for them after that.

robinmessage: When I looked at the backend, it looked like 5 minutes for logged in users, and 15 minutes for anonymous. Or is it more complex than that? I rounded down on the basis that I doubt an anonymous user will wait for even five minutes.

I could duplicate that logic? I suppose the question is, what are we trying to achieve here? Is it even clear to the user why there is a timeout? How are they meant to try again, or are they not meant to try again?

Thinking about it again, perhaps we should just lock the question permanently (in the Redux store so they can still refresh to try again)?

mlt47: Yeah that's a good point.
It is mainly to stop people from writing scripts to guess the answer.
Its second purpose is to stop users from blindly guessing.

Ideally we would give them a warning on the soft limit saying that if they continue to guess they will be timed out. Unfortunately we don't convey the soft limit through the API. If you think it'll take about 30 mins to do, implementing a warning pop-up if they break the soft limit will hopefully be enough of a cue for any legitimate user to stop guessing and be more cautious.

Otherwise we go with what we have here. It is an improvement on Physics.

I don't think we should block the question in Redux because if they are legitimate users who have over-guessed they will probably leave the tab open with their attempt and try again later.

robinmessage: K, I will do that.

Maybe I could change the message to something like "This question is locked until at least HH:MM:SS to prevent guessing?"

mlt47: That sounds good to me

mlt47:
The warning is great

PR #74, opened 2019-06-28
Admin pages: show to staff users

robinmessage:


src/app/components/pages/Admin.tsx
```diff
@@ -95,7 +95,7 @@ const AdminPageComponent = ({user, getContentVersion, setContentVersion, content
                                             <RS.Button
                                                 type="button" className="p-0 border-dark"
                                                 onClick={startVersionUpdate}
-                                                disabled={displayVersion === contentVersion.liveVersion}
+                                                disabled={user.role != "ADMIN" || displayVersion === contentVersion.liveVersion}
```
jsharkey13: Any reason this isn't `isAdmin(user)` ?

robinmessage: That might make sense but the types don't quite line up...

PR #77, opened 2019-06-29
Add the admin stats page

jsharkey13:
A quick working version of the admin stats page, which will be useful soon.

 There's a lot of type information missing because the return type from the API is not a standard type, and the page relies on augmenting the returned object to add TOTAL counts too.
The page format is as in physics, but without the loading of groups and event bookings since they are additional HTTP requests that would have required much larger changes.

Things that might want altering:
 - we augment the returned object, rather than making a new copy.
 - the types for the `AdminStatsResponse` are very lax because `{[role in Role | "TOTAL"]: number}` and suchlike seemed wrong.
 - there's no consistency in navigation buttons and layout between admin pages like there is on physics 
 - there's no events and groups stats, unlike physics, but we don't need them yet


src/app/components/navigation/IsaacApp.tsx
```diff
@@ -36,11 +36,12 @@ import {Header} from "./Header";
 import {Route} from "react-router";
 import {ScrollManager} from "./ScrollManager";
 import {AdminUserManager} from "../pages/AdminUserManager";
+import {AdminStats} from "../pages/AdminStats";
 import {AdminContentErrors} from "../pages/AdminContentErrors";
 import {ActiveModal} from "../elements/ActiveModal";
 import {isAdmin, isStaffUser, isTeacher} from "../../services/user";
 import {Groups} from "../pages/Groups";
-import { Equality } from '../pages/Equality';
```
Morpheu5: 😭

PR #78, opened 2019-07-03
Table overflow

mlt47:
The table in the quick question of `/concepts/sys_bool_table_for_circuit#worked_example` when viewed on a mobile is a good example of why this is needed.

### DIFF
--- a/src/app/components/elements/TrustedHtml.tsx
+++ b/src/app/components/elements/TrustedHtml.tsx
@@ -234,22 +234,25 @@ export function katexify(html: string, userPreferences: UserPreferencesDTO | nul
     return output;
 }
 
-// RegEx replacements to match Latex inspired Isaac Physics functionality
-function bootstrapify(html: string) {
-    // TODO does not work if class was added by content team
-    // TODO need to center table in parent element if possible
-    const regexRules = {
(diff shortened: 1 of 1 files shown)


src/app/components/elements/TrustedHtml.tsx
```diff
@@ -234,22 +234,26 @@ export function katexify(html: string, userPreferences: UserPreferencesDTO | nul
...
+function manipulateHtml(html: string) {
+    // This can't be quick but it is more robust than writing regular expressions...
+    htmlDom.innerHTML = html;
 
+    // Table manipulation - apply bootstrap classes and insert parent div to handle table overflow
+    const tableElements = htmlDom.getElementsByTagName("table");
+    const tableClasses = "table table-bordered w-100 text-center bg-white m-0";
+    for (let i = 0; i < tableElements.length; i++) {
+        const table = tableElements[i];
+        table.setAttribute("class", (table.getAttribute("class") || "") + tableClasses);
```
Morpheu5: Make classes unique, just in case?

mlt47: Ah yes. I had not thought about that. Browser won't care, but I agree that it would be nicer to avoid duplicate classes.

PR #80, opened 2019-07-08
Add keyboard support for indentation in Parsons problems.

Morpheu5:
- `]` increments the indentation of the item being dragged
- `[` decrements the indentation of the item being dragged

IMPORTANT: An item must be currently being dragged (select + `[space]`) for this to work.


Morpheu5:
No worries, my bad, please ignore! :)

bh412:
Works fine and looks ok when you drag with the mouse and use the keys to indent. It doesn't indent if you use tab to select the desired element and then press keys (i.e. it highlights but doesn't select/indent) which Meurig points out would be necessary for non-mouse users

Morpheu5:
That's the way the library works: a highlighted element (with tab) is not
in the `dragging` state until you select it with space. As far as I could
see, you can't even move them around if they aren't in the `dragging` state.

Il ven 12 lug 2019, 14:36 Ben Hanson <notifications@github.com> ha scritto:

bh412:
That makes sense, I'm not too up on non-mouse users, would it be obvious to someone who doesn't use one that they should press the space bar? If so that seems legitimate

Morpheu5:
> That makes sense, I'm not too up on non-mouse users, would it be obvious to someone who doesn't use one that they should press the space bar? If so that seems legitimate

[This document](https://dev.opera.com/articles/accessible-drag-and-drop/) suggests it's a common pattern thing, and in fact I found it mentioned in many other places, including [this post](https://medium.com/salesforce-ux/4-major-patterns-for-accessible-drag-and-drop-1d43f64ebf09). I don't think it's an official standard, but rather a de-facto standard.

PR #81, opened 2019-07-09
Add 'View in the Content Editor' button

Morpheu5:


src/app/components/pages/Concept.tsx
```diff
@@ -43,6 +44,7 @@ const ConceptPageComponent = ({urlConceptId, doc, fetchDoc}: ConceptPageProps) =
                     currentPageTitle={doc.title as string}
                     collectionType={navigation.collectionType}
                 />
+                {ISAAC_ENV != "live" && <EditContentButton canonicalSourceFile={EDITOR_URL + (doc as any)['canonicalSourceFile']} />}
```
bh412: Can this all be moved into the EditContentButton passing through the doc to prevent it being written 3 times?

Morpheu5: Yes, but a bit of an overkill. To actually avoid this being written three times, it would be nicer to pass doc to the page title component and move the edit button inside there, but with @mlt47 we determined that there are only three dynamic content types, and so the redundancy was acceptable.

src/app/services/constants.ts
```diff
@@ -15,6 +15,14 @@ if (document.location.hostname === "localhost") {
...
 export const API_PATH: string = apiPath;
 
+let isaacEnv: string = "dev";
+if (document.location.origin.indexOf("https://staging.isaaccomputerscience.org") > -1) {
+    isaacEnv = "staging";
+} else if (document.location.origin.indexOf("https://isaaccomputerscience.org/") > -1) {
+    isaacEnv = "live";
+}
+export const ISAAC_ENV: string = isaacEnv;
+
```
jsharkey13: This is not the way to do this. 

There's an API endpoint specifically for this sort of thing, and we [load it for the header and add it to rootScope on Isaac Physics](https://github.com/isaacphysics/isaac-app/blob/390f25a8cddfd6f9407e7f9ffc25bafa04b78056/app/js/app/controllers/HeaderControllers.js#L19-L25). You can [see the endpoint in `Api.js`](https://github.com/isaacphysics/isaac-app/blob/390f25a8cddfd6f9407e7f9ffc25bafa04b78056/app/js/app/services/Api.js#L450).

jsharkey13: Unfortunately, it's a runtime thing, not a constant. I don't know how this messes things up; perhaps @mlt47 will have an idea?

src/app/components/elements/EditContentButton.tsx
```diff
@@ -0,0 +1,12 @@
+import React, {ReactElement, useEffect} from "react";
+
+export interface EditContentButtonProps {
+    canonicalSourceFile?: string,
+    className?: string,
+}
+
+export const EditContentButton = ({canonicalSourceFile, className}: EditContentButtonProps) => {
+    return <div>
+        <a href={canonicalSourceFile} className={`btn btn-primary mt-3 ${className ? ` ${className}` : ""}`}>View in the Content Editor</a>
```
jsharkey13: This isn't exactly subtle, is it :P

src/app/services/constants.ts
Morpheu5: Fixed.

src/app/components/elements/EditContentButton.tsx
Morpheu5: Not meant to be :)

jsharkey13:
@mlt47 - is there not an easier / better way of loading this constant from the API? Surely we can load it once into the state store and then use it from there?

mlt47:
> @mlt47 - is there not an easier / better way of loading this constant from the API? Surely we can load it once into the state store and then use it from there?

The action code already does this. The request action just returns if the value is already set.
Perhaps an improvement would be to move the useEffect(()=>{action()}) for requesting the constant into IsaacApp so that it is only tried once, at the time that the react component for isaac app is created. Then you could remove the use effects from question, concept and generic pages.

PR #82, opened 2019-07-10
Fix unmoving carousel item

bh412:
Adds next/previous index methods to enable carousel movement
Defaults to 5 seconds like Isaac Physics


mlt47:
I will look over this and integrate it with the fixes that Nomensa have provided. I will need to check the WCAG stuff to make sure auto-scrolling can be done in an acceptable way for screen-readers because Nomensa explicitly turned it off.

mlt47:
I cannot be sure that auto-progressing carousels act nicely for screen readers so I think we'll leave it off until someone complains and/or we have more experience testing with screen readers.

PR #83, opened 2019-07-16
Add exam board filtering to search results

bh412:
- Moves filterOnExamBoard to `examBoard` service from the `topics` service.
- Filters search results based on user specified exam board
- Provides Anon user choice like topics in the filter section

### DIFF
--- a/src/app/components/pages/Search.tsx
+++ b/src/app/components/pages/Search.tsx
@@ -102,9 +102,9 @@ const SearchPageComponent = (props: SearchPageProps) => {
         return keepElement || isStaffUser;
     };
 
-    const filteredSearchResults = searchResults && searchResults.results && searchResults.results.filter(filterResult);
+    const filteredSearchResults = searchResults && searchResults.results && filterOnExamBoard(searchResults.results.filter(filterResult), examBoard);
 
-    const shortcutSearchResults = filterOnExamBoard(((shortcutResponse || []).concat(filteredSearchResults || [])), examBoard);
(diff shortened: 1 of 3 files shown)


src/app/services/examBoard.ts
```diff
@@ -8,3 +9,8 @@ export const determineExamBoardFrom = (userPreferences?: UserPreferencesDTO | nu
         return EXAM_BOARD.OCR;
     }
 };
+
+export const filterOnExamBoard = (contents: ContentSummaryDTO[], examBoard: EXAM_BOARD) => {
+    console.log(examBoard);
```
Morpheu5: Remove `console.log` maybe?

src/app/components/pages/Search.tsx
```diff
@@ -98,8 +104,7 @@ const SearchPageComponent = (props: SearchPageProps) => {
 
     const filteredSearchResults = searchResults && searchResults.results && searchResults.results.filter(filterResult);
 
-    // const shortcutSearchResults = Object.assign([], shortcutResponse, filteredSearchResults);
-    const shortcutSearchResults = (shortcutResponse || []).concat(filteredSearchResults || []);
+    const shortcutSearchResults = filterOnExamBoard(((shortcutResponse || []).concat(filteredSearchResults || [])), examBoard);
```
mlt47: I would maybe move filterOnExamBoard(...) to wrap searchResults.results.filter(filterResult) above. Could also rename shortcutSearchResults to shortcutsAndFilteredSearchResults, finalSearchResults or something with describes it better.

bh412:
Implemented suggested changes

PR #92, opened 2019-08-07
Inequality UI and usability improvements

Morpheu5:
This should now be working consistently across devices and browsers.

Needs more testing on IE/Edge.


Morpheu5:
This is now fixed in c52f580 by making the moving element transparent to events.

For reference: https://caniuse.com/#feat=pointer-events

There is a [polyfill](https://github.com/kmewhort/pointer_events_polyfill) just in case.

PR #95, opened 2019-08-28
Use exact for all routes to stop substring matchs

Scratchcat1:
All routes in IsaacApp now use exact to stop substring matches.
Previously 'path="/teachers"' would also match the path
"/teachers/non-existant-path"


jsharkey13:
So I was wrong in my advice, and I think we want to add the exact keyword to the ones with variables in the paths too!

PR #97, opened 2019-08-30
Split related content into two columns

Scratchcat1:
Splits into questions and content. May need changes to look better.

### DIFF
--- a/src/app/components/elements/RelatedContent.tsx
+++ b/src/app/components/elements/RelatedContent.tsx
@@ -48,22 +48,36 @@ function getURLForContent(content: ContentSummaryDTO) {
 }
 
 export const RelatedContentComponent = ({content, parentPage, logAction}: RelatedContentProps) => {
-    return <div className="simple-card my-3 my-md-5 p-3 p-md-5 text-wrap">
-        <h2 className="mb-4">Related Content</h2>
-        <div className="d-lg-flex">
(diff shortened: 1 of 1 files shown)


src/app/components/elements/RelatedContent.tsx
```diff
@@ -48,22 +48,36 @@ function getURLForContent(content: ContentSummaryDTO) {
...
+        </ListGroupItem>
+    )
+    return <div className="row">
+        <div className="simple-card my-3 p-3 text-wrap col-lg-5">
+            <h2 className="mb-4">Related questions</h2>
+            <div className="d-lg-flex">
+                <ListGroup className="w-100 mb-lg-3 mr-lg-3 w-lg-50">
+                    {content.filter((contentSummary) => contentSummary.type == DOCUMENT_TYPE.QUESTION).sort((a, b) => {
+                        if (a.level === b.level) return (a.title || '').localeCompare(b.title || '');
+                        return parseInt(a.level || '-1') > parseInt(b.level || '-1') ? 1 : -1;
```
mlt47: Be careful with not returning 0 when a == b. I don't actually think it'll make a difference in this case, but Java would act in odd ways as the natural order would not be consistent with equals. One example where it would cause problems is with storing the values in tree-like data-structure such as in a sorted set.

PR #99, opened 2019-08-30
Feature/assignment count nav bar

Scratchcat1:
Add notification pill for assignements to navbar


Scratchcat1:
- [ ] Appears to crash IE when not logged in

PR #100, opened 2019-08-30
Feature/item question feedback

Scratchcat1:
Add better feedback to parsons questions.

### DIFF
--- a/src/IsaacApiTypes.tsx
+++ b/src/IsaacApiTypes.tsx
@@ -168,6 +168,10 @@ export interface QuantityValidationResponseDTO extends QuestionValidationRespons
     correctUnits?: boolean;
 }
 
+export interface ItemQuestionValidationResponseDTO extends QuestionValidationResponseDTO {
+    incorrectItemIds: [string];
+}
+
 export interface QuestionValidationResponseDTO {
     questionId?: string;
     answer?: ChoiceDTO;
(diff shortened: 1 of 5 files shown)


src/IsaacApiTypes.tsx
```diff
@@ -168,6 +168,10 @@ export interface QuantityValidationResponseDTO extends QuestionValidationRespons
     correctUnits?: boolean;
 }
 
+export interface ItemQuestionValidationResponseDTO extends QuestionValidationResponseDTO {
+    incorrectItemIds: [string];
+}
```
mlt47: IsaacApiTypes actually gets programmatically generated form the java code so any types added manually will get overwritten.
I have not run the type generation for a while so ItemQuestionValidationResponseDTO is not currently present. I'll re-run the generation script and update this branch to make sure it matches/

src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -228,7 +236,9 @@ class IsaacParsonsQuestionComponent extends React.Component<IsaacParsonsQuestion
                                             {(provided: DraggableProvided, snapshot: DraggableStateSnapshot) => {
                                                 return <div
                                                     id={`parsons-item-${item.id}`}
-                                                    className={`parsons-item indent-${item.indentation}`}
+                                                    className={`parsons-item indent-${item.indentation} ` + classnames({
+                                                        "incorrect-item": item.id && validationResponse && validationResponse.incorrectItemIds && validationResponse.incorrectItemIds.indexOf(item.id) > -1
```
mlt47: This is fine, but as we also have a polyfill so that we can safely use use `validationResponse.includes(item.id)` if you would prefer.

jsharkey13:
I don't think we gain enough from this to go forward any longer. Closing, but not deleting the branch in case we review later on!

PR #107, opened 2019-09-13
Allow verification without need to be logged in

Scratchcat1:
Allow the verification link to be used without being logged in.

Success/Error messages will be shown depending on the state of
the user and if the verification fails.

### DIFF
--- a/src/app/components/handlers/EmailAlterHandler.tsx
+++ b/src/app/components/handlers/EmailAlterHandler.tsx
@@ -5,8 +5,7 @@ import {AppState, ErrorState} from "../../state/reducers";
 import {Link} from "react-router-dom";
 import queryString from "query-string";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
-import {connect} from "react-redux";
-import {LoggedInUser} from "../../../IsaacAppTypes";
+import {connect, useSelector} from "react-redux";
 const stateToProps = (state: AppState, {location: {search}}: {location: {search?: string}}) => ({
     errorMessage: state ? state.error : null,
     queryParams: search ? queryString.parse(search) : {}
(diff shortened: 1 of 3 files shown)


src/app/components/handlers/EmailAlterHandler.tsx
```diff
@@ -44,22 +43,21 @@ const EmailAlterHandlerComponent = (props: EmailAlterHandlerProps) => {
                     <CardBody className="m-4">
                         {/* Isaac Physics had a large icon here */}
 
-                        {emailVerificationSuccess ?
+                        {(!errorMessage || emailVerificationSuccess) &&
                             <React.Fragment>
-                                <h3 className="mb-4">Email address verified</h3>
+                                <h3 className="mb-4">{ emailVerificationSuccess ? "Email address verified" : "Email address verification token received. Log in to confirm verification." }</h3>
```
mlt47: There are four cases we need to cover:
1.  Verification success and user is logged in as the user which matches userid query param value.
2.  Verification success and the user is not logged in.
3.  Verification success and user is logged in as a different user to the value in the userid query param value.
4.  Verification failure.

We don't currently cover case (3). Isaac Physics added a warning saying something along the lines of "You are signed in as a different user to the user with the email you have just verified". It is worth adding this case to avoid confusion even though it is probably a rare occurrence.

src/app/components/handlers/EmailAlterHandler.tsx
```diff
@@ -44,33 +41,34 @@ const EmailAlterHandlerComponent = (props: EmailAlterHandlerProps) => {
                     <CardBody className="m-4">
                         {/* Isaac Physics had a large icon here */}
 
-                        {emailVerificationSuccess ?
+                        {(!errorMessage || emailVerificationSuccess) &&
                             <React.Fragment>
-                                <h3 className="mb-4">Email address verified</h3>
+                                <h3 className="mb-4">{successMessage}</h3>
                                 <Button tag={Link} to="/account" color="secondary" block>
                                     Continue to My Account
```
mlt47: In the new case when the user is logged in as a different user, it would be better to hide this button.

PR #109, opened 2019-09-19
Make question buttons underlined when focused

Scratchcat1:
Previously quick question buttons would change in appearance when
focused. The normally used shadow was removed since it would make the
question look horrible.
Instead the button now underlines its text when in focus.


jsharkey13:
Sorry, but this seems to have been superseded by https://github.com/isaacphysics/isaac-cs-app/pull/111 now!

PR #112, opened 2019-09-24
Registration DoB checkbox enabling

bh412:
dateOfBirth is now optional so checkbox disabling needs to check for being undefined not null

### DIFF
--- a/src/app/components/pages/Registration.tsx
+++ b/src/app/components/pages/Registration.tsx
@@ -252,7 +252,9 @@ const RegistrationPageComponent = ({user, updateCurrentUser, errorMessage, userE
                                             checked={confirmedOverThirteen}
                                             required
                                             label="I am at least 13 years old"
-                                            disabled={registrationUser.dateOfBirth !== undefined}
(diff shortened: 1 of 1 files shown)


src/app/components/pages/Registration.tsx
```diff
@@ -252,7 +252,7 @@ const RegistrationPageComponent = ({user, updateCurrentUser, errorMessage, userE
                                             checked={confirmedOverThirteen}
                                             required
                                             label="I am at least 13 years old"
-                                            disabled={registrationUser.dateOfBirth !== null}
+                                            disabled={registrationUser.dateOfBirth !== undefined}
```
mlt47: DateInput sets the value to null on clear so to re-enable the checkbox after the input box is cleared we must also not disable when the date of birth is null.
```suggestion
                                            disabled={registrationUser.dateOfBirth}
```

bh412: Implemented with comment about future restructuring of DoB handling

PR #113, opened 2019-09-27
Adds skip to content for keyboard users

bh412:
- Adds an id to main in IsaacApp so it can be linked to.
- Same font and size as navbar
- scss used to hide off screen unless focussed on, as suggested here: https://webaim.org/techniques/skipnav/

### DIFF
--- a/src/scss/header.scss
+++ b/src/scss/header.scss
@@ -81,12 +81,12 @@ a.skip-main {
 a.skip-main:focus, a.skip-main:active {
   color: #000;
   background-color:#fff;
-  left: -30%;
+  left: 5%;
   top: 60px;
   width: 150px;
   height: auto;
   overflow: auto;
-  margin: 10px 35%;
+  margin: 10px;
   text-align:center;
   font-family: $secondary-font-medium;
   font-size: 1rem;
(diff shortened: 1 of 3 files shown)


src/scss/header.scss
```diff
@@ -69,6 +69,30 @@ header {
...
+}
+a.skip-main:focus, a.skip-main:active {
+  color: #000;
+  background-color:#fff;
+  left: -30%;
+  top: 60px;
+  width: 150px;
+  height: auto;
+  overflow: auto;
+  margin: 10px 35%;
```
jsharkey13: Is there any reason you've used the funky `margin` stuff when you'reabsolutely positioning this? Why have a `left` of `-30%` and then a left margin of `35%`? Surely `left: 5%` would have done?

src/scss/header.scss
```diff
@@ -69,6 +69,30 @@ header {
...
+  height:1px;
+  overflow:hidden;
+  z-index:-999;
+}
+a.skip-main:focus, a.skip-main:active {
+  color: #000;
+  background-color:#fff;
+  left: -30%;
+  top: 60px;
+  width: 150px;
```
jsharkey13: This width may want to be bigger if you accept the change above!
```suggestion
  width: 175px;
```

src/scss/header.scss
bh412: you're correct, will implement

PR #115, opened 2019-10-02
Add accessible alternative text to KaTeX equations

jsharkey13:
This is based on a library that was created by the KaTeX/Khan Academy team, but is quite heavily modified to change the way certain expressions are described. It aims to make them simpler, but may lead to less clear descriptions in obscure cases. (E.g. integrals and sums now get limits descibed as `sum from, ..., to ...., of ....`, and common fractions like a half are now read as `one half`).
Tested with a large sample of physics and CS content!

This may want reviewing if KaTeX/KaTeX#820 or KaTeX/KaTeX#2072 produce anything more useful or general than my edited version of the library.

### DIFF
--- a/src/app/components/elements/TrustedHtml.tsx
+++ b/src/app/components/elements/TrustedHtml.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import katex from "katex";
 import 'katex/dist/contrib/mhchem.js';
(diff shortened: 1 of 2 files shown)


jsharkey13:
KaTeX is MIT, and we have credited the original source of the file in a future-proof manner!

PR #117, opened 2019-10-03
Event leader role

mlt47:
Matching back-end PR exists


src/app/services/user.tsx
```diff
@@ -16,6 +16,10 @@ export function isStaff(user?: LoggedInUser | null) {
     return user ? user.loggedIn && (user.role == "ADMIN" || user.role == "EVENT_MANAGER" || user.role == "CONTENT_EDITOR") : false;
 }
 
+export function isEventsLeader(user?: LoggedInUser | null) {
```
jsharkey13: I'd prefer `isEventLeader`

src/app/services/user.tsx
```diff
@@ -16,6 +16,10 @@ export function isStaff(user?: LoggedInUser | null) {
...
+    return user ? user.loggedIn && user.role == "EVENT_LEADER" : false;
+}
+
 export function isEventsManager(user?: LoggedInUser | null) {
```
jsharkey13: And here `isAdminOrEventManager`?

src/app/components/pages/Search.tsx
```diff
@@ -94,7 +94,7 @@ const SearchPageComponent = (props: SearchPageProps) => {
         doSearch();
     }, [searchFilterProblems, searchFilterConcepts]);
 
-    const isStaffUser = userRole && (userRole == 'ADMIN' || userRole == 'EVENT_MANAGER' || userRole == 'CONTENT_EDITOR' || userRole == 'STAFF');
```
jsharkey13: I know I edited this, is there any reason it can't use the service now?

PR #121, opened 2019-10-08
Prevent internal explosion for invalid type or category

bh412:


jsharkey13:
He's had a look, so we're all good!

PR #122, opened 2019-10-08
Feature/my boards

bh412:
Add a my boards page with table and card view


src/app/components/pages/MyGameboards.tsx
```diff
@@ -0,0 +1,381 @@
...
+                                                    </th>
+                                                    <th></th>
+                                                    <th></th>
+                                                </tr>
+                                            </thead>
+                                            <tbody>
+                                                {boards.boards
+                                                    .filter(board => board.title && board.title.toLowerCase().includes(boardTitleFilter.toLowerCase()))
+                                                    .map(board =>
+                                                        <Board
```
mlt47: This needs a key or react might get confused particularly as we could be reordering these components

src/app/components/pages/MyGameboards.tsx
```diff
@@ -0,0 +1,381 @@
...
+                        </Form>
+                    </Col>
+                </Row>
+                <ShowLoading until={boards}>
+                    {boards && boards.boards && <div>
+                        {boardView == boardViews.card ?
+                            // Card view
+                            <div className="block-grid-xs-1 block-grid-md-2 block-grid-lg-3 my-2">
+                                {boards.boards.map(board => <div key={board.id}>
+                                    <Board
```
mlt47: This needs a key or react might get confused particularly as we could be reordering these components

src/app/components/pages/MyGameboards.tsx
```diff
@@ -0,0 +1,381 @@
...
+                                    </Row>
+                                </CardHeader>
+                                <CardBody id="boards-table">
+                                    <div className="overflow-auto">
+                                        <Table className="mb-0">
+                                            <thead>
+                                                <tr>
+                                                    <th>Completion</th>
+                                                    <th className="pointer-cursor">
+                                                        <span onClick={() => boardOrder == BoardOrder.title ? setBoardOrder(BoardOrder["-title"]) : setBoardOrder(BoardOrder.title)}>
```
mlt47: Spans can't have onClick properties without also receiving a role (for accessibility reasons). Maybe a button would be better? The other headers have the same problem but the linter should pick this up.

src/app/components/pages/MyGameboards.tsx
```diff
@@ -0,0 +1,381 @@
...
+    // @ts-ignore
+    return <Container>
+        <TitleAndBreadcrumb currentPageTitle="My gameboards" intermediateCrumbs={[STUDENTS_CRUMB]} help={pageHelp} />
+        {boards && boards.totalResults == 0 ? <h3 className="text-center mt-4 mb-5">You have no gameboards to view; use one of the options above to find one.</h3> :
+            <React.Fragment>
+                {boards && boards.totalResults > 0 && <h4>You have <strong>{boards.totalResults}</strong> gameboard{boards.totalResults > 1 && "s"} saved...</h4>}
+                {!boards && <h4>You have <Spinner size="sm" /> saved gameboards...</h4>}
+                <Row>
+                    <Col>
+                        <Form inline>
```
mlt47: Forms have default actions on submission you'll need to stop those from happening

src/app/components/pages/MyGameboards.tsx
```diff
@@ -0,0 +1,381 @@
...
+    }, [boards]);
+
+    const pageHelp = <span>
+        A summary of your gameboards
+    </span>;
+
+    // @ts-ignore
+    return <Container>
+        <TitleAndBreadcrumb currentPageTitle="My gameboards" intermediateCrumbs={[STUDENTS_CRUMB]} help={pageHelp} />
+        {boards && boards.totalResults == 0 ? <h3 className="text-center mt-4 mb-5">You have no gameboards to view; use one of the options above to find one.</h3> :
```
mlt47: There are no options above to find gamebaords?

src/app/components/pages/MyGameboards.tsx
```diff
@@ -0,0 +1,381 @@
...
+                                                            groups={groups}
+                                                            selectedBoards={selectedBoards}
+                                                            setSelectedBoards={setSelectedBoards}
+                                                            boardView={boardView}
+                                                            user={user}
+                                                            loadGroups={dispatch(loadGroups)}
+                                                            boards={boards}
+                                                            loadBoards={dispatch(loadBoards)}
+                                                            deleteBoard={dispatch(deleteBoard)}
+                                                        />)
```
mlt47: Just so that we don't forget what we talked about, you can check these to see whether you need to pass them down. Most if not all of the redux actions and state can just be asked for form the Board component.

You probably won't need these after the changes but it is important to notice that `<Component someProp={dispatch(loadGroups)} />` will be executing a dispatch of the function loadGroups on every render, and the value of someProp is whatever dispatch returns.
What you need instead is something like `<Component someProp={(arg) => {dispatch(loadGroups(arg))}} />` so that someProp's value is then a function which can be called to dispatch loadGroups.

PR #125, opened 2019-10-22
Feature/school choice

bh412:


jsharkey13:
I wonder if we can't use the [Async Creatable](https://react-select.com/creatable) from this example. This gets us the advantage of a built-in loading message, and prevents creating a new school whilst the list is loading.

`loadOptions` would need to make the API call, and then in `.then(...)` do the transformation currently done in the `useEffect` watching `schoolSearchResults`. This would return the promise that the async-creatable can then use to determine the loading stuff.

`formatCreateLabel` could be used to change the `Create ...` text to be `Use '...' as your school name`. We may eventually want a warning that these won't count for events etc.

src/app/components/elements/inputs/SchoolInput.tsx
```diff
@@ -46,81 +33,91 @@ export const SchoolInput = ({userToUpdate, setUserToUpdate, submissionAttempted,
...
-        (userToUpdate.schoolOther !== undefined && userToUpdate.schoolOther !== "" && userToUpdate.schoolOther !== NOT_APPLICABLE)
+    const promiseOptions = (inputOptions: string) =>
+        new Promise(resolve => {
+            setTimeout(() => {resolve(api.schools.search(inputOptions).then(({data}) => {
+                let temp: any = [];
+                data && data.length > 0 && data.map((item: any) => (temp.push({value: item, label: item.name + ", " + item.postcode})));
+                return temp;
+            }).catch((response) => {
+                console.error("Error searching for schools. ", response);
+            }))}, 800);
```
jsharkey13: Why do we need this 800ms delay? Is it an attempt at debouncing?
We could really do with a service for debouncing and rate-limiting function calls. I think it's more responsive with it at like 200ms, but maybe that will hammer the API?

PR #126, opened 2019-10-24
Feature/admin user edit

bh412:


Morpheu5:
- Possible bug with timezone saving birth day one day prior to the one set by the user.
- Need to touch up wording when editing someone else's date of birth (You vs The user).

Otherwise good.

src/app/components/pages/MyAccount.tsx
```diff
@@ -148,111 +169,122 @@ const AccountPageComponent = ({user, updateCurrentUser, errorMessage, userAuthSe
...
 
-                        <TabPane tabId={ACCOUNT_TAB.emailpreferences}>
-                            <UserEmailPreference
-                                emailPreferences={emailPreferences} setEmailPreferences={setEmailPreferences}
-                                submissionAttempted={attemptedAccountUpdate}
-                            />
-                        </TabPane>
+                            <TabPane tabId={ACCOUNT_TAB.passwordreset}>
+                                <UserPassword
+                                    currentUserEmail={user && user.email && user.email} userAuthSettings={userAuthSettings}
```
robinmessage: I can't see easily offhand, but is this mean to be `user` or should it be `userToUpdate`?

src/app/state/actions.tsx
```diff
@@ -244,13 +262,26 @@ export const updateCurrentUser = (
...
+            dispatch(showToast({
+                title: "Account settings updated",
+                body: "Your account settings were updated successfully.",
+                color: "success",
+                timeout: 5000,
+                closable: false,
+            }) as any);
+        }
+        if (editingOtherUser) {
+            history.push('/');
```
robinmessage: Why force the user to homepage here? I imagine staying on this page or going back to a list of users or wherever they came from might be useful.

src/app/components/pages/MyAccount.tsx
bh412: you're right, I added a check on userToUpdate if it exists or user if not

src/app/components/pages/MyAccount.tsx
```diff
@@ -44,50 +44,70 @@ import {history} from "../../services/history"
...
-    // Inputs which trigger re-render
-    const [attemptedAccountUpdate, setAttemptedAccountUpdate] = useState(false);
+    useEffect(() => {editingOtherUser && userFind && setUserToEdit(Object.assign({}, userFind))}, [userFind]);
 
     // - Copy of user to store changes before saving
-    const [userToUpdate, setUserToUpdate] = useState(Object.assign({}, user, {password: ""}));
-    useMemo(() => {setUserToUpdate(Object.assign({}, user, {password: ""}))}, [user]);
+    const [userToUpdate, setUserToUpdate] = useState(editingOtherUser && userOfInterest && userFind ? Object.assign({}, userFind, {loggedIn: true, password: ""}) : Object.assign({}, user, {password: ""}));
+
+    useEffect(() => {editingOtherUser && userToEdit && setUserToUpdate(Object.assign({}, userToEdit, {loggedIn: true}))}, [userToEdit]);
```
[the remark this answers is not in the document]
bh412: I think this is from when I was storing the usertoupdate as the user, it's now separated out so usertoupdate shouldn't be set to user any more. If you're editing yourself usertoupdate isn't set hence the logic in the useEffect, if you're editing someone else, it is, so no accidental crossed wires!

src/app/state/actions.tsx
bh412: We're following the behaviour from physics atm, when you click edit on the user manager it opens the account edit page in a new tab, so going back to the user list might be tricky, aside from closing the tab? We ran into user consistency errors if I stayed on the same page, as it saved then nav'd to the /account page as the current admin user (in the background) but the front looked like the edited user still, so if you saved or anything after it threw the consistency error. At the moment I agreed with Andrea to keep the behaviour the same as physics and just go to the homepage

PR #128, opened 2019-10-28
Feature/focus reflow fix

bh412:


src/scss/focus.scss
```diff
@@ -55,15 +55,15 @@
...
 
 .btn-secondary:focus {
-  box-shadow: none;
-  border: 0.2rem solid black !important;
+  box-shadow: 0 0 0 2pt black !important;
   border-radius: 8px;
+  outline: none !important;
 }
 
 .btn-danger:focus {
```
mlt47: Let's make the change here too while we're at it

src/scss/focus.scss
```diff
@@ -78,25 +78,17 @@
...
-  border: 0.2rem solid $secondary;
-  border-radius: 8px;
-}
-
-.btn.btn-sm.a-alt:focus {
-  box-shadow: none;
-  border: 0.2rem solid $focus-blue;
-  border-radius: 5px;
+  border-color: $secondary !important;
 }
```
mlt47: Why has this one been deleted?

src/scss/focus.scss
```diff
@@ -78,25 +78,17 @@
...
+  border-radius: 3px;
+  outline: none !important;
 }
 
 .footerTop a:focus {
-  border: 0.2rem solid white;
+  outline: 0.2rem solid white;
 }
 
 // Close modal button
```
mlt47: I imagine this close button's focus is fine but it is worth a check if it hasn't already been checked

bh412: all good

src/scss/focus.scss
bh412: It didn't appear to affect anything, but I've added it back incase

PR #132, opened 2019-10-31
Restrict quiz access

mlt47:


src/app/components/pages/Quiz.tsx
```diff
@@ -13,19 +13,18 @@ import {ShowLoading} from "../handlers/ShowLoading";
...
 import {AppState} from "../../state/reducers";
 import {ContentBase} from "../../../IsaacApiTypes";
-import {ACCEPTED_QUIZ_IDS, DOCUMENT_TYPE, EDITOR_URL, NOT_FOUND} from "../../services/constants";
+import {ACCEPTED_QUIZ_IDS, DOCUMENT_TYPE, EDITOR_URL} from "../../services/constants";
 import {RelatedContent} from "../elements/RelatedContent";
 import {WithFigureNumbering} from "../elements/WithFigureNumbering";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
 import {EditContentButton} from "../elements/EditContentButton";
-import {questions} from "../../state/selectors";
+import * as selectors from "../../state/selectors";
```
Morpheu5: I feel this is OK in this case, but can we use individual imports instead? I fear the message will pass that this is OK -- and it is, for small "libraries" -- but it can hinder tree-shaking and increase the size of the bundle long term if used too liberally.

mlt47: That makes sense

src/app/components/pages/Question.tsx
```diff
@@ -34,7 +35,9 @@ export const Question = withRouter(({questionIdOverride, match}: QuestionPagePro
 
     const dispatch = useDispatch();
     useEffect(() => {
-        dispatch(fetchDoc(DOCUMENT_TYPE.QUESTION, questionId));
+        if (!ACCEPTED_QUIZ_IDS.includes(questionId)) {
```
jsharkey13: Why do we need this _and_ the `useSelector` above, when in `Quiz.tsx` we only have the `useSelector`?

mlt47: It is here to stop the request being made and so avoiding network traffic inspection. The selector stops it from being shown if it already is in the doc part of the redux store. If you visit the quiz, then change the url to question, it will show the quiz as a question page because the useEffect on the question page would not request the question, and it is that requesting of the question that clears the redux store value.

What we should actually do is add a clean up method to the useEffect which clears the redux store on component dismount. We would then need to make a slight change to stop the loading spinner showing forever... but the other solution came first.

PR #139, opened 2019-12-09
Replace hardcoded summary links with linked gameboard based ones

Scratchcat1:
Previously linked gameboards for a summary page were generated in an unreliable manner.
Instead find the first linked gameboard which matches the user's examboard and link to that.
Don't show the button if no such linked gameboard exists.


Scratchcat1:
[ ] Other things still need to be done

PR #143, opened 2019-12-11
Allow teachers to reserve spaces for batches of students onto events

Morpheu5:
- [x] Teacher will, on the event booking form, see an extra button, “Reserve spaces”.
- [x] This will open a modal containing their groups and the reservations which they have already made.
- [x] The students that they have already reserved will be available from event bookings' reservedBy field.
- [x] From here they can make reservations for their students or manage the reservations they have already made by cancelling them.
  - [x] make reservations
  - [x] manage reservations made
- [x] The limit for how many students they are allowed to reserve will be present from the event object.
  - [x] needs enforcing via the UI
  - [x] needs changes in the editor


src/app/components/elements/modals/ResearvationsModal.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+export const reservationsModal = () => {
+    return {
+        closeAction: () => {store.dispatch(closeActiveModal())},
+        size: 'xl',
+        title: "Groups and Reservations",
+        body: <ReservationsModal />,
+        // af599 TODO: Wouldn't it be nice if we coul move the Reserve button
+        //             here instead of having it in the component? Would it even
+        //             be possible?
```
mlt47: Yeah I think we need to rethink active modal as the current data structure does not allow sharing of state between body and buttons which is a bit silly but I didn't think of that at the start 🙃.

src/app/components/pages/EventDetails.tsx
```diff
@@ -178,7 +180,7 @@ export const EventDetails = ({match: {params: {eventId}}, location: {pathname}}:
 
                                 {/* Options for logged-in users */}
                                 {user && user.loggedIn && <span>
-                                    {event.eventStatus != 'CLOSED' && !event.expired && !bookingFormOpen && !(event.userBooked || event.userOnWaitList) && <RS.Button
+                                    {event.eventStatus != 'CLOSED' && !event.expired && !bookingFormOpen && !(event.userBooked || event.userOnWaitList || event.userBookingStatus !== "RESERVED") && <RS.Button
```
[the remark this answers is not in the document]
mlt47: Not quite what I thought but users cannot now book onto an event unless they have first been reserved a space. If I undo the changes it will work as it should.

src/app/components/pages/EventDetails.tsx
```diff
@@ -188,7 +197,12 @@ export const EventDetails = ({match: {params: {eventId}}, location: {pathname}}:
                                     >
                                         Close booking form
                                     </RS.Button>}
-                                    {event.userBooked && !event.expired && <RS.Button
+                                    {user.role && ['TEACHER', 'EVENT_LEADER', 'EVENT_MANAGER', 'ADMIN'].includes(user.role) && <RS.Button
```
mlt47: We have a utility method in a service for this but we should also only show this button when booking is possible

src/app/components/elements/modals/ResearvationsModal.tsx
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }, [currentGroup]);
+    
+    const eventBookingsForGroup = useSelector((state: AppState) => state && state.eventBookingsForGroup || []);
+
+    useEffect(() => {
+        // af599 TODO:
+        //   Check the API (EventsFacade), it returns the wrong type of users
+        //   (i.e., the one with an email address attached) even when the
+        //   request comes from a TEACHER.
```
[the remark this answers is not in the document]
mlt47: I've checked this, it is ok

src/app/components/elements/modals/ResearvationsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+    useEffect(() => {
+        dispatch(loadGroups(false));
+    }, []);
+
+    useEffect(() => {
+        if (currentGroup && !currentGroup.members) {
+            dispatch(getGroupMembers(currentGroup));
+        } else if (currentGroup && currentGroup.members) {
+            // af599 TODO: Retrieve event status for members maybe?
```
robinmessage: Just flagging this: fix, make a card in Trello, or just leave for the future?

src/app/components/elements/modals/ResearvationsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }, [currentGroup]);
+
+    useEffect(() => {
```
robinmessage: This could be joined up with the effect above, and possibly covers the TODO?

src/app/components/elements/modals/ResearvationsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+    const toggleAllUnbooked = () => {
+        setCheckAllCheckbox(!checkAllCheckbox);
+        let checkboxes = { ...userCheckboxes };
+        for (const id in userCheckboxes) {
+            checkboxes[id] = !checkAllCheckbox;
+        }
+        setUserCheckboxes(checkboxes);
+    };
+
+    const toggleCancelReservationCheckboxeForUser = (userId?: number) => {
```
robinmessage: typo: Checkboxe

src/app/components/elements/modals/ResearvationsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+        }
+        setCancelReservationCheckboxes(checkboxes);
+    };
+
+    const requestReservations = () => {
+        if (selectedEvent && selectedEvent.id && currentGroup && currentGroup.id) {
+            const reservableIds = Object.entries(userCheckboxes).filter(c => c[1]).map(c => parseInt(c[0]));
+            dispatch(reserveUsersOnEvent(selectedEvent.id, reservableIds, currentGroup.id));
+        }
+        setCheckAllCheckbox(false);
```
robinmessage: I think I see why this is here, but the effect earlier also resets this (or it could if it doesn't) so I'm not sure if this is needed here.

src/app/components/elements/modals/ResearvationsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+    };
+
+    const cancelReservations = () => {
+        if (selectedEvent && selectedEvent.id && currentGroup && currentGroup.id) {
+            const cancellableIds = Object.entries(cancelReservationCheckboxes).filter(c => c[1]).map(c => parseInt(c[0]));
+            dispatch(cancelReservationsOnEvent(selectedEvent.id, cancellableIds, currentGroup.id));
+        }
+        setCheckAllCancelReservationsCheckbox(false);
```
robinmessage: Same for this one

jsharkey13:
There are still some issues with the usability of the pages and flow:

 - [ ] There's no indication of the student's email verification status on the group booking page, but we do have that information in the group member list, and those without verified emails cannot be booked. We should indicate this in the front-end by disabling the row or using some warning icon?
 - [ ] The group booking modal opens when the user has no groups, and the dropdown is empty. Could it link to the Manage Groups page and say "Create a group from the [Manage Groups] page to book your students onto an event" and hide the dropdown?
 - [ ] For the group booking modal, the description text is on top of the group selection dropdown for smaller screens. It needs an additional column offset (say `<Col md={{size: 8, offset: 1}}>`). It would also be easier to see and use if the student selection table was in a new row underneath the dropdown and description text.
---
Also:
 - [ ] There's an unused import in `ReservationsModal.tsx`, and the filename (and just the filename) has a spelling mistake in it, too.
 - [ ] There's a merge conflict with `package-lock.json` again :/

Morpheu5:
> * [ ]  There's an unused import in `ReservationsModal.tsx`, and the filename (and just the filename) has a spelling mistake in it, too.

I removed the unused import, but what is the spelling mistake? If that's the 's', then it's intentional, unless it's not conventional, in which case I'm happy to change it.

> * [ ]  There's a merge conflict with `package-lock.json` again :/

Can we not just pretend these don't exist? My strategy is to delete them, `npm i && npm ci`, and then commit whatever is the result.

src/app/components/elements/modals/ResearvationsModal.tsx
Morpheu5: Which effect are you referring to?

src/app/components/elements/modals/ResearvationsModal.tsx
Morpheu5: Fixed in 14227b65

src/app/components/elements/modals/ResearvationsModal.tsx
Morpheu5: Fixed in 14227b65.

src/app/components/elements/modals/ResearvationsModal.tsx
Morpheu5: Fixed in 14227b65.

src/app/components/elements/modals/ResearvationsModal.tsx
robinmessage: The one on `eventBookingsForGroup`, line 56. I could be wrong about the logic though.

src/app/components/elements/modals/ResearvationsModal.tsx
robinmessage: Although actually the effect on line 56 doesn't actually call `setCheckAllCancelReservationsCheckbox(false)`, which I think it probably should?

PR #146, opened 2019-12-20
Integrate google maps with the events page

Scratchcat1:
Add a map to the events page displaying all of the events.
Uses the @react-google-maps/api package, chosen over google-map-react
to be more actively maintained.

Note:
The state variable mapCenter prevents the map from recentering to
the starting location (Cambridge) when ever an info window is hidden.

Currently a request is made each time the events filter changes
in order only show the correct events. Ideally this only needs to be
one request which provides all event locations and to filter locally.


bh412:
https://stackoverflow.com/questions/54555877/can-i-use-overlappingmarkerspiderfier-with-react-google-maps

PR #149, opened 2020-01-09
Feature/news

bh412:
* Adds news in the same way as pods on IP
* Adds actions/reducers to get news pod content from API


src/app/components/elements/cards/NewsCard.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import * as RS from "reactstrap";
+import classnames from "classnames";
+import {Link} from "react-router-dom";
+import {IsaacPodDTO} from "../../../../IsaacApiTypes";
+import {API_PATH} from "../../../services/constants";
+
+export const NewsCard = ({newsItem, pod = false}: {newsItem: IsaacPodDTO; pod?: boolean}) => {
```
jsharkey13: We don't need the `pod` prop for NewsCards, since they are always shown as pods. EventCards needed that for the events page itself.

PR #152, opened 2020-01-16
Free text test page

mlt47:


src/app/state/actions.tsx
```diff
@@ -23,7 +23,7 @@ import {
     AppGroupMembership,
     ATTENDANCE,
     BoardOrder,
-    EmailUserRoles,
+    EmailUserRoles, FreeTextRule,
```
jsharkey13: I think this one and the line beneath will be reordered when someone next formats the file properly. Could they be inline, rather than two-to-a-line? (Ctrl-Alt-O in WebStorm will do this for you.)

src/app/state/reducers.tsx
```diff
@@ -36,7 +36,7 @@ import {
     UserSummaryForAdminUsersDTO,
     UserSummaryWithEmailAddressDTO,
     UserSummaryWithGroupMembershipDTO,
-    GlossaryTermDTO
+    GlossaryTermDTO, TestCaseDTO
```
jsharkey13: Ditto with this being inline, rather than doubled up?

PR #155, opened 2020-01-29
Alter breadcrumb and next content behaviour

bh412:


src/app/services/topics.tsx
```diff
@@ -23,22 +23,26 @@ export const idIsPresent = (id: string, contents: {id?: string}[] | undefined) =
     return contents && !!contents.filter((content) => content.id === id);
 };
 
-export const determineTopicHistory = (currentTopic: CurrentTopicState) => {
+export const determineTopicHistory = (currentTopic: CurrentTopicState, currentDocId: string) => {
     const result: LinkInfo[] = [];
     if (currentTopic && currentTopic != NOT_FOUND && currentTopic.id && currentTopic.title) {
+        const relatedContent: (string | undefined)[] = [];
+        currentTopic.relatedContent && currentTopic.relatedContent.map(content => relatedContent.push(content.id));
```
jsharkey13: I don't like this inline `&&` syntax particularly much; it muddles values and logic together.

src/app/services/topics.tsx
```diff
@@ -23,22 +23,26 @@ export const idIsPresent = (id: string, contents: {id?: string}[] | undefined) =
...
         result.push(ALL_TOPICS_CRUMB);
-        result.push({title: currentTopic.title, to: `/topics/${currentTopic.id.slice("topic_summary_".length)}`})
+        relatedContent.indexOf(currentDocId) > -1 && result.push({title: currentTopic.title, to: `/topics/${currentTopic.id.slice("topic_summary_".length)}`})
```
jsharkey13: They could both easily be `if`s, which I think would read clearer.

src/app/services/topics.tsx
```diff
@@ -23,22 +23,26 @@ export const idIsPresent = (id: string, contents: {id?: string}[] | undefined) =
...
 export const makeAttemptAtTopicHistory = () => {
-    return [ALL_TOPICS_CRUMB, {title: "Topic", to: "/topics/"}]
+    return [ALL_TOPICS_CRUMB]
 };
 
 
 export const determineNextTopicContentLink = (currentTopic: CurrentTopicState | undefined, contentId: string, examBoard: EXAM_BOARD) => {
-    if (currentTopic && currentTopic != NOT_FOUND && currentTopic.relatedContent) {
+    const relatedContent: (string | undefined)[] = [];
+    currentTopic && currentTopic != NOT_FOUND && currentTopic.relatedContent && currentTopic.relatedContent.map(content => relatedContent.push(content.id));
```
jsharkey13: This whole logic could do with being cleaner too, but it works

src/app/services/topics.tsx
bh412: I've changed this for you

PR #156, opened 2020-01-30
Add shorturls to share links

bh412:
See: https://github.com/isaacphysics/isaac-router/commit/b1a98497e299fa3f3621b59068c2b9cca28d765e
for router short urls

### DIFF
--- a/src/app/components/pages/Concept.tsx
+++ b/src/app/components/pages/Concept.tsx
@@ -57,7 +57,7 @@ const ConceptPageComponent = ({urlConceptId, conceptIdOverride, doc, fetchDoc, s
                     <EditContentButton canonicalSourceFile={EDITOR_URL + (doc as ContentBase)['canonicalSourceFile']} />
                     }
                     <div className="question-actions question-actions-leftmost mt-3">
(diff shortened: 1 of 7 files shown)


src/app/components/pages/Concept.tsx
```diff
@@ -57,7 +57,7 @@ const ConceptPageComponent = ({urlConceptId, conceptIdOverride, doc, fetchDoc, s
                     <EditContentButton canonicalSourceFile={EDITOR_URL + (doc as ContentBase)['canonicalSourceFile']} />
                     }
                     <div className="question-actions question-actions-leftmost mt-3">
-                        <ShareLink linkUrl={`${window.location.origin}/concepts/${doc.id}`}/>
+                        <ShareLink linkUrl={segueEnvironment !== "DEV" ? `https://isaaccs.org/c/${doc.id}` : `${window.location.origin}/concepts/${doc.id}`}/>
```
mlt47: We will need to do something different for Physics. Best to do it now in case we forget later.
Perhaps we have a SHORT_URL somewhere which switches depending on SITE_SUBJECT.

PR #162, opened 2020-02-07
Removes afterAuth path on navigation away from /login 

bh412:
Doesn't remove on tab close, but if we did the google login redirect wouldn't direct to the correct place. We do remove from local storage correctly on login/google redirect


bh412:
Doesn't work with google on redirect. A minor issue that doesn't affect people often

PR #169, opened 2020-02-11
Pad requests containing credentials with random data to hamper bicycle attacks

jsharkey13:
See https://arxiv.org/abs/2002.01513 for an explanation of a bicycle attack. The length of the encrypted request gives away information about the length of a user's password; pad with random data up to the next power of two to avoid this, with a minumim padded length of 256 to better protect users with weaker passwords.
There are two vulnerable Isaac endpoints; `/resetpassword/[token]` which has only the password in the request body, and `/authenticate` which contains email and password. Both endpoints expect a `Map<String, String>` and so will cope with the additional data with no required changes.
One further endpoint does accept passwords; the update user endpoint. But this also has the user object, and user preferences, and is thus much less sensitive to the length of the user's password.


src/app/services/credentialPadding.ts
```diff
@@ -0,0 +1,41 @@
...
+    // Pad to the next power of two above the body size, but at least to 2^8 = 256 bytes:
+    return Math.pow(2, Math.max(Math.ceil(Math.log(bodyLength)/Math.log(2)), 8)) - bodyLength;
+}
+
+function generatePadding(paddingLength: number) {
+    return Array(paddingLength).fill(0).map(() => Math.random().toString(36).charAt(2)).join('');
+}
+
+export function securePadCredentials(credentials: {email: string; password: string}) {
+    let paddedCredentials: PaddedCredentials = Object.assign({_randomPadding: ""}, credentials);
```
mlt47: `: PaddedCredentials` would  be inferred here so it is not necessary (well the name of the type would not be inferred but the shape of it would which is all that is needed in a structural type system like this rather than a nominal type system like Java).

PR #173, opened 2020-02-17
Feature/modal fixing

bh412:
Allows 'ActiveModals' to be closed by pressing escape, or clicking off the modal. Watches for this and implements closeActiveModal at the right time, to prevent activeModals stacking with duplicate keys. Stops 'ActiveModals' having a horizontal scroll when not overflowing


src/app/components/elements/modals/ActiveModal.tsx
```diff
@@ -1,14 +1,29 @@
...
+    const [isOpen, setIsOpen] = useState(true);
+
+    const toggle = () => {
+        const isNowOpen = !isOpen;
+        setIsOpen(isNowOpen);
+    };
+
+    useEffect(() => {
+        if (!isOpen) {
+            store.dispatch(closeActiveModal());
```
mlt47: Probably better to `const dispatch  = useDispatch()` over importing store.

PR #182, opened 2020-03-10
Concepts page

robinmessage:
There is a bit of copy-pasted code from Search but I wasn't sure if it was worth extracting out the common bits at the moment.


jsharkey13:
Compare new on the left, vs old on the right:
image

I think we need to decide how much of the old layout we want before the beta, and how much can wait. I think our existing React class for concepts doesn't cope so well with physics concept pages, and the concepts page has always displayed them differently so we may need a new class here anyhow.

jsharkey13:
This still needs work, but it's good enough for the beta site. Merging!

PR #186, opened 2020-03-13
Physics: disable required fields modal unless on the CS site

robinmessage:


mlt47:
A slight extension on this, that I've just realised is needed, is that the account page also requires fields for submission.
It was my suggestion to just not include the middleware for Physics but I think to do it properly we will need to, instead, alter the validation service to require different fields for Physics and CS (it then doesn't really matter if the middleware is included or not).
I'll close the pull request, fix the task description and put it back on the queue for whoever to pick up. Sorry!

PR #187, opened 2020-03-13
Physics: event card images should be rectangular

robinmessage:


robinmessage:
> Looks good, works fine differentiating between phy/cs - Would you be able to get rid of the white margin at the top of the events cards on physics, with the rectangular images, to match what they look like at the moment? Just a tiny change

I've fixed this now @bh412

PR #189, opened 2020-03-16
Feature/hexagon pages

bh412:


src/app/components/pages/Alevel.tsx
```diff
@@ -0,0 +1,125 @@
+import React from "react";
+import {Col, Container, Row} from "reactstrap";
+import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
+
+export const Alevel = () => {
+    return<Container>
```
robinmessage: Nitpicking, but should have a space here between `return` and `<Container>`

src/app/components/pages/Alevel.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <Row className="pb-4">
+            <Col>
+                <TitleAndBreadcrumb currentPageTitle={"A Level Resources"} breadcrumbTitleOverride="A Level Resources" />
```
robinmessage: `breadcrumbTitleOverride` overrides `currentPageTitle`, so there is no need to use override if they are the same.

src/app/components/pages/TeacherFeatures.tsx
```diff
@@ -0,0 +1,108 @@
...
+import classNames from "classnames";
+
+export const TeacherFeatures = () => {
+
+    const user = useSelector((state: AppState) => state && state.user);
+
+    return<Container>
+        <Row className="pb-4">
+            <Col>
+                <TitleAndBreadcrumb currentPageTitle={"Teacher Features"} breadcrumbTitleOverride="Teacher Features" />
```
robinmessage: And here.

src/app/components/pages/TeacherFeatures.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    <div className="hexagon">
+                        <img className="hexagon-field" src="/assets/teacher_features_sprite.svg#teacher-hat" alt="Isaac hexagon"></img>
+                        <div className="hexagon-title">
+                            1. You are now a teacher!
+                        </div>
+                    </div>
+                </Col>
+            }
+            <Col md="auto">
+                <a href="groups" className={classNames({"hexagon": true, "disabled": (isStudent(user) || !isLoggedIn(user))})}>
```
robinmessage: You could extract `(isStudent(user) || !isLoggedIn(user))`  into a variable, and maybe the whole `className` for each type of thing that repeats.

src/app/components/pages/TeacherFeatures.tsx
```diff
@@ -0,0 +1,98 @@
...
+        </Row>
+        <Row className="teacher-feature-body justify-content-md-center">
+            {isStudent(user) || !isLoggedIn(user) ?
+                <Col md="auto">
+                    <Hexagon link={"/teacher_account_request"}
+                        imageSrc={"/assets/teacher_features_sprite.svg#teacher-hat"}
+                        title={"1. Tell us you are a teacher"}/>
+                </Col> :
+                <Col md="auto">
+                    <div className="hexagon">
```
robinmessage: If Hexagon took an optional `className`, these could be `Hexagon`s too. Not essential but would be nice.

PR #193, opened 2020-03-24
Inequality: show python editor and preview

robinmessage:
Andrea and I have tested this and it seems to work well.


Morpheu5:
This should be good to go.

jsharkey13:
Some things so far:
 - LGTM has some warnings that could do with addressing.
 - At the moment, for text entry, text entered by the user seems to be replaced once parsing has finished. This removes our ability to do research and improve based on what the user types. It is also confusing for the user, if they pause briefly, to have everything they have typed rewritten underneath them? We did used to set `pythonExpression` to be the parsed value, though as this does. I also think we used to send `"textEntry":true,"userInput":".........."` in the `value`  of the answer request.
 - Some symbols do not parse correctly, and others align terribly:
image

Morpheu5:
> * LGTM has some warnings that could do with addressing.

Most of the time, LGTM is as useful as that weird 60yo down the street who just learned about TikTok and is yelling at the screen to get off their lawn. I suggest we take it with a pinch of salt.

> * At the moment, for text entry, text entered by the user seems to be replaced once parsing has finished.

That I told Robin about but I guess he forgot and then I forgot so I'll try to fix it now.

> * Some symbols do not parse correctly, and others align terribly:
>   image

This branch and the branch where I fixed this were out of sync. I just synced them and it should now work.

jsharkey13:
I believe that this PR is entirely contained inside https://github.com/isaacphysics/isaac-cs-app/pull/205 now, and so we do not need this separate one. I believe GitHub will actually mark it merged automagically, so I'll leave it open for now.

PR #200, opened 2020-04-01
Alters the layout of the related questions and concepts

bh412:
Concepts before questions to match physics
Sensible font size
Icons
Level indicator


src/scss/phy/footer.scss
```diff
@@ -3,3 +3,22 @@
...
+  background: url(/assets/phy/icon-books.png);
+}
+
+.related-title {
+  height: 65px;
+  padding-left: 75px;
+  padding-top: 23px;
+  background-repeat: no-repeat;
+  &:after
+  {
```
jsharkey13, line 21: Can we avoid this old SCSS nastiness and have indentation, and brackets on the same lines?
```scss
    &:after {
        clear: both;
    }
```

PR #203, opened 2020-04-02
Implement CS proposal 4-7

bh412:


bh412:
Added more spacing and set the links to be site specific

bh412:
I've just been copying they physics behaviour - happy to change it if requested!

PR #204, opened 2020-04-02
Swap useMemo for useEffect (first pass)

Morpheu5:
There will be other uses in other branches. Somewhere, it makes sense to useMemo. None of the cases I removed here did make sense.


src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -82,13 +82,13 @@ export const QuestionSearchModal = ({originalSelectedQuestions, setOriginalSelec
...
-    useMemo(() => {
+    useEffect(() => {
         setSearchExamBoards([examBoardTagMap[examBoard]].filter(tag => !!tag));
-    }, [user]);
+    }, [user, examBoard]);
 
     useEffect(() => {
         searchDebounce(searchQuestionName, searchTopics, searchLevels, searchExamBoards, searchBook, searchFastTrack, 0);
-    },[searchQuestionName, searchTopics, searchLevels, searchExamBoards, searchBook, searchFastTrack]);
+    },[searchDebounce, searchQuestionName, searchTopics, searchLevels, searchExamBoards, searchBook, searchFastTrack]);
```
Morpheu5, line 91: I hate exhaustive dependencies like this, but apparently it's OK to include functions if they are [referentially equal](https://pauldrummond.dev/posts/useeffect-and-exhaustive-deps/).

PR #206, opened 2020-04-03
Adds subject colour and white text for correct on phy

bh412:
Piggybacks subject check in phy/subjects.scss
Corrects the correct colour (for phy/maths/chem) and makes the text white


bh412:
Added the font and well done/try again

PR #207, opened 2020-04-03
Move examboard tab special case to the IsaacTabs level removing…

mlt47:
… unnecessary padding

### DIFF
--- a/src/app/components/content/IsaacTabs.tsx
+++ b/src/app/components/content/IsaacTabs.tsx
@@ -1,9 +1,12 @@
-import React, {ReactElement, useEffect} from "react";
+import React, {ReactElement} from "react";
 import {Tabs} from "../elements/Tabs";
 import {ContentDTO} from "../../../IsaacApiTypes";
 import {IsaacContent} from "./IsaacContent";
 import {useCurrentExamBoard} from "../../services/examBoard";
 import {SITE, SITE_SUBJECT} from "../../services/siteConstants";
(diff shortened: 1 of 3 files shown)


src/app/components/content/IsaacTabs.tsx
```diff
@@ -11,18 +11,24 @@ interface IsaacTabsProps {
...
         tabTitlesToContent[tabTitle] = <IsaacContent doc={child} />;
     });
 
+    // EXAM BOARD Special Case
+    const examBoardFilter = useCurrentExamBoard();
+    const tabTitles = Object.keys(tabTitlesToContent);
+    const specialCaseExamBoardTab = tabTitles.includes("AQA") && tabTitles.includes("OCR") && tabTitles.length === 2;
+    if (SITE_SUBJECT === SITE.CS && specialCaseExamBoardTab) {
+        // return <IsaacContent doc={tabTitlesToContent[examBoardFilter as any]} />
+        return <div>{tabTitlesToContent[examBoardFilter]}</div>
```
mlt47, line 28: Ah yes, we've already put the child doc into an IsaacContent block on line 19, oops!
Safe to delete the comment on line 27.
We could also delete the div and just return tabTitlesToContent[examBoardFilter] I think?

Morpheu5: I just added the expand button so the outer `div` needs to stay.

Morpheu5:
Laura approves.

PR #208, opened 2020-04-06
Event: gmail export button

robinmessage:


src/app/components/pages/EventDetails.tsx
```diff
@@ -15,7 +15,23 @@ import * as persistence from "../../services/localStorage";
 import {KEY} from "../../services/localStorage";
 import {history} from "../../services/history";
 import {atLeastOne, validateBookingSubmission, zeroOrLess} from "../../services/validation";
-import {isTeacher} from "../../services/user";
+import {isStaff, isTeacher} from "../../services/user";
+
+function formatDate(date: Date|number) {
```
Morpheu5, line 20: Given this is bound to happen in other parts of the site, can we not use a library for this? I used moment.js in the past with some success, but there are others.

robinmessage: We could, I wasn't sure about bringing in a library for one function, but I'm happy to.

robinmessage:
@mlt47 I've included moment.js now.

PR #211, opened 2020-04-09
Implement 15-17 in the booking proposal

bh412:
- Change date/time to use 24 hour time
- Remove booking deadline from page
- Change booking form open wording


src/app/components/pages/EventDetails.tsx
```diff
@@ -122,13 +122,6 @@ export const EventDetails = ({match: {params: {eventId}}, location: {pathname}}:
                                             {user && user.loggedIn && user.email && event.userOnWaitList && <span> - You are on the waiting list for this event.</span>}
                                         </td>
                                     </tr>}
-                                    {event.bookingDeadline && <tr>
```
jsharkey13, line 125: I'd just have commented this block out for now; we're going to want it for physics, even if for whatever reason they don't want it for CS.

src/app/components/elements/DateString.ts
```diff
@@ -1,14 +1,14 @@
...
 // 22/01/2020
 export const NUMERIC_DATE = new Intl.DateTimeFormat(dateLocale);
 // Wed, 22 Jan 2020
 export const FRIENDLY_DATE = new Intl.DateTimeFormat(dateLocale, {weekday: "short", day: "numeric", month: "short", year: "numeric"});
 // 22/01/2020, 13:00:00
 export const NUMERIC_DATE_AND_TIME = new Intl.DateTimeFormat(dateLocale, {year: "numeric", month: "numeric", day: "numeric", hour: "numeric", minute: "numeric", second: "numeric", hour12: false});
-// Wed, 22 Jan 2020, 1:00 pm
-export const FRIENDLY_DATE_AND_TIME = new Intl.DateTimeFormat(dateLocale, {weekday: "short", day: "numeric", month: "short", year: "numeric", hour: "numeric", minute: "2-digit", hour12: true});
+// Wed, 22 Jan 2020, 13:00
+export const FRIENDLY_DATE_AND_TIME = new Intl.DateTimeFormat(dateLocale, {weekday: "short", day: "numeric", month: "short", year: "numeric", hour: "numeric", minute: "2-digit", hour12: false});
```
jsharkey13, line 11: Is this ok for all the other places these constants are used? If so, just mark this resolved.

bh412: we don't use it anywhere else at the moment

PR #216, opened 2020-04-16
Search results and concepts page: coloured and formatted nicely

robinmessage:


src/scss/common/elements.scss
```diff
@@ -2,8 +2,9 @@
 .content-summary-link .content-summary-link-title {
   background: none;
   border: 0;
-  font-family: $secondary-font;
-  font-weight: 900;
+  .text-secondary {
+    font-family: $secondary-font-bold;
```
jsharkey13, line 6: I can see that the `font-weight` wasn't being used before, but I actually quite like it without the bold and think perhaps it looks too intense on the physics version. 
Does anyone else have an opinion?  It's using semi-bold on the Angular physics app, but we don't use that font style in React.

jsharkey13: Turns out there's a `-semi-bold` too, which is closest to what it used to be. Seems sensible to compromise to that!

PR #218, opened 2020-04-16
Inequality improvements - April 2020

Morpheu5:
* 93cd22c1 - Fix an issue with reloading the previous question attempt in inequality
* 9c0797ae - Enable leaving inequality by pressing [esc]ape
* e6ba5a90 - Only update the text-entry box if the graphical editor is visible
* b1e10deb - Restore inequality mode selection for /equality


bh412:
When you enter something in the text entry, on an untouched question, you don't get the option to submit the answer until you put it into they symbolic editor. The /equality page works but doesn't have a background, and doesn't have text entry. Pressing escape works! When you refresh/return, it remembers the symbolic but not the text entry, and doesn't mark as correct, but gives option to check answer again to mark correct

PR #222, opened 2020-04-21
Feature/role modify

bh412:
Allows admins to modify user roles differently to event managers (the page access verification prevents anyone else being able to change roles)

### DIFF
--- a/src/app/components/pages/AdminUserManager.tsx
+++ b/src/app/components/pages/AdminUserManager.tsx
@@ -43,6 +43,11 @@ const AdminUserManagerComponent = ({currentUser, adminUserSearch, adminModifyUse
     });
     const [selectedUserIds, setSelectedUserIds] = useState<number[]>([]);
     const userIdToSchoolMapping = useSelector((state: AppState) => state && state.userSchoolLookup);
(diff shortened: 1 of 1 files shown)


src/app/components/pages/AdminUserManager.tsx
```diff
@@ -222,7 +229,7 @@ const AdminUserManagerComponent = ({adminUserSearch, adminModifyUserRoles, admin
                             <RS.DropdownToggle caret disabled={userUpdating} color="primary" outline>Modify Role</RS.DropdownToggle>
                             <RS.DropdownMenu>
                                 <RS.DropdownItem header>Promote or demote selected users to:</RS.DropdownItem>
-                                {["STUDENT", "TEACHER", "EVENT_LEADER"].map(role =>
+                                {(promotableRoles).map(role =>
```
Morpheu5, line 232: Doesn't this allow ADMINs to demote other ADMINs?

jsharkey13: No, there's no actual security worth doing in the frontend. The API will only allow you to modify people whose role is strictly below yours, but you can promote those below you to any role up to your own as a staff user. So it doesn't actually matter what this says, except perhaps to avoid confusion - but since promote and demote are mixed, we need all the roles.

PR #225, opened 2020-04-24
Feature/related natural sort

bh412:
- Defaults related content to naturally order by title, and if there is not title then falls back to previous sort by level
- Adds natural sort my groups and search, already worked in my boards.


src/app/components/pages/Search.tsx
```diff
@@ -83,7 +83,10 @@ export const Search = withRouter((props: {history: History; location: Location})
     const filteredSearchResults = searchResults && searchResults.results &&
         filterOnExamBoard(searchResults.results.filter((result) => searchResultIsPublic(result, user)), examBoard);
 
-    const shortcutAndFilteredSearchResults = (shortcutResponse || []).concat(filteredSearchResults || []);
+    const shortcutAndFilteredSearchResults = (shortcutResponse || []).concat(filteredSearchResults || []).sort((a, b) => {
+        if (a.title && b.title) return (a.title.localeCompare(b.title, undefined, { numeric: true, sensitivity: 'base' }));
```
Morpheu5, line 87: Using `options` may not work on some browsers (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

jsharkey13: It works on Safari and IE11 when I test it, and those are the two most non-compliant browsers out there, so I think we're good. Works in Firefox and Chrome fine too.

PR #226, opened 2020-04-28
Add ie unsupported pages with webpack changes

bh412:
Adds static unsupported pages which we will use for Internet Explorer


public/unsupported_browser.phy.html
```diff
@@ -0,0 +1,100 @@
...
+            top: 0px;
+            right: 30px;
+        }
+
+    </style>
+</head>
+<body>
+<div id="panel">
+    <h1>Internet Explorer unsupported</h1>
+    <h2>Isaac Physics no longer supports internet explorer, Microsoft recommends using
```
Morpheu5, line 80: We may want to capitalise Internet Explorer. Same in the other file.

jsharkey13:
I have updated the formatting and added the nginx config. I have commented it out for now, since the consensus is that we need a modal warning people of IE's impending block.

PR #227, opened 2020-04-29
Refactor use of ShareLink and make more accessible

mlt47:


src/app/components/elements/ShareLink.tsx
```diff
@@ -24,28 +23,29 @@ export const ShareLink = (props: {linkUrl: string}) => {
...
+            }
+        }
+    }, [showShareLink]);
+
     return <React.Fragment>
         <button className="share-link-icon btn-action" onClick={() => toggleShareLink()} aria-label="Get share link"/>
-        <div className={classnames({"share-link": true, "d-block": showShareLink})}><div ref={shareLink}>{shareUrl}</div></div>
+        <div
+            className={`share-link ${showShareLink ? "d-block" : ""}`}
+            style={{width: Math.min((shareUrl.length + 1), 20) * 8.5}}
```
jsharkey13, line 48: Why is this `Math.min(...)`? Surely it ought to be `Math.max(...)`, since we always want it to be at least 20px wide and up to the length of the URL? I suppose we may also want to limit it to some maximum width too.
Also, why `8.5`?

mlt47: Ah yes, I had it without a min earlier and 8.5 seemed to fit the text best, but then while testing having really long links causes everything to be a mess (can't remember the conditions particularly, probably the pages with gameboard cards, maybe on mobile). Perhaps we can impose a max width more explicitly than what I've done here. We are probably unlikely to have URLs which are less than 20 characters.

mlt47: but yeah 8.5 was pretty arbitrary

PR #231, opened 2020-05-05
Feature/gameboard coloured hexagons

bh412:
Implements logic to either show CS hexagon on Isaac CS, or choose most occurring -> least occurring subject order on Isaac Physics

### DIFF
--- a/src/app/services/gameboards.tsx
+++ b/src/app/services/gameboards.tsx
@@ -4,6 +4,7 @@ import {NOT_FOUND} from "./constants";
 import React from "react";
 import countBy from "lodash/countBy"
 import intersection from "lodash/intersection"
+import {SITE, SITE_SUBJECT} from "./siteConstants";
 
 enum boardCompletions {
     "any" = "Any",
@@ -68,6 +69,9 @@ export const generateGameboardSubjectHexagons = (boardSubjects: string[]) => {
 };
 
 export const determineGameboardSubjects = (board: GameboardDTO) => {
+    if (SITE_SUBJECT === SITE.CS) {
+        return ["compsci"];
+    }
     let allSubjects: string[] = [];
     board.questions?.map((item) => {
         let tags = intersection(["physics","maths","chemistry"], item.tags || []);
--- a/src/app/services/gameboards.tsx
+++ b/src/app/services/gameboards.tsx
@@ -80,10 +80,10 @@ export const determineGameboardSubjects = (board: GameboardDTO) => {
     }
     );
     // If none of the questions have a subject tag, default to physics
-    if (allSubjects.length == 0) {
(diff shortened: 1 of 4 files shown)


src/app/services/gameboards.tsx
```diff
@@ -63,4 +66,22 @@ export const generateGameboardSubjectHexagons = (boardSubjects: string[]) => {
...
-}
+};
+
+export const determineGameboardSubjects = (board: GameboardDTO) => {
+    if (SITE_SUBJECT === SITE.CS) {
+        return ["compsci"];
+    }
+    let allSubjects: string[] = [];
+    board.questions?.map((item) => {
+        let tags = intersection(["physics","maths","chemistry"], item.tags || []);
```
jsharkey13, line 77: Could you format this list slightly nicer with spaces between the items? And the same for the function call arguments below? I'm used to Python where a linter would get grumpy at these things

src/app/services/gameboards.tsx
```diff
@@ -63,4 +66,22 @@ export const generateGameboardSubjectHexagons = (boardSubjects: string[]) => {
...
+        tags.forEach(tag => allSubjects.push(tag));
+    }
+    );
+    // If none of the questions have a subject tag, default to physics
+    if (allSubjects.length == 0) {
+        allSubjects.push("physics");
+    }
+    let enumeratedSubjects = countBy(allSubjects);
+    return Object.keys(enumeratedSubjects).sort(function (a,b) {return enumeratedSubjects[b]-enumeratedSubjects[a]});
```
jsharkey13, line 86: This does random (well, determined by the internals of lodash) things in the case of ties; i.e. the same number of questions per subject. To get closest to the old behaviour, could we sort first by subject? That way physics can be used in preference.
I.e. have `const subjects = ["physics", "maths", "chemistry"];` right at the start, use that array in the intersection, and then sort by `function(a, b) {return subjects.indexOf(a) - subjects.indexOf(b)}` before the existing sort? There will only over be three items max, so one more sort won't be too costly.

src/app/services/gameboards.tsx
```diff
@@ -63,4 +66,22 @@ export const generateGameboardSubjectHexagons = (boardSubjects: string[]) => {
...
+        return ["compsci"];
+    }
+    let allSubjects: string[] = [];
+    board.questions?.map((item) => {
+        let tags = intersection(["physics","maths","chemistry"], item.tags || []);
+        tags.forEach(tag => allSubjects.push(tag));
+    }
+    );
+    // If none of the questions have a subject tag, default to physics
+    if (allSubjects.length == 0) {
```
jsharkey13, line 82: I think this should be `===` triple equals, just to be safe?

PR #232, opened 2020-05-06
Add a button to allow previewing of content version changes

jsharkey13:
This uses the editor as a redirect proxy to GitHub; but the editor is already configured with the correct GitHub repo to use in all cases, and so this prevents duplicated configuration.


jsharkey13:
You may need to clear your cache and/or reload the editor with cache disabled to pick up the new version of the editor code, since it's heavily cached.

PR #235, opened 2020-05-15
Fix subject preference rendering

bh412:
Previously rendered before subject preferences returned. Now has useEffect to update and useState to manage controlled checkbox

### DIFF
--- a/src/app/components/elements/inputs/subjectInterestInput.tsx
+++ b/src/app/components/elements/inputs/subjectInterestInput.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useReducer, useState} from "react";
+import React, {useEffect, useState} from "react";
 import {CustomInput} from "reactstrap";
 
 interface SubjectInterestInputProps<T> {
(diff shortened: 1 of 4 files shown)


src/app/components/elements/inputs/subjectInterestInput.tsx
```diff
@@ -1,17 +1,28 @@
...
 interface SubjectInterestInputProps<T> {
     stateObject: T;
     propertyName: string;
     setStateFunction: (stateObject: T) => void;
 }
+
 export const SubjectInterestInput = (props: SubjectInterestInputProps<any>) => {
     const {stateObject, propertyName, setStateFunction} = props;
-    return <CustomInput id={`${propertyName}-checkbox`} type="checkbox" className="isaac-checkbox" defaultChecked={stateObject[propertyName] === true}
+    const [thisPref, setThisPref] = useState(stateObject[propertyName]);
```
mlt47, line 12: This is sort of a hack because it just duplicates the data already stored in the stateObject prop. I'm guessing we're having trouble with getting the component to (re-)render at the correct time? It would be nice if we could do it properly. I'll have a little look at it but feel free to  have a look yourself if you feel you want to, @bh412.

PR #236, opened 2020-05-15
Feature/questionnaire

bh412:
Gets notifications from the API, requires relevant API change to select only role relevant notifications


src/scss/common/modals.scss
```diff
@@ -4,6 +4,12 @@
   max-width: none !important;
 }
 
+.selection-buttons {
+  height: 90%;
+  margin-bottom: 1rem !important;
+
+}
```
mlt47, line 11: I see what you're trying to do here but it feels a little hacky. This same effect can be achieved using bootstrap's predefined classes (particularly the flexbox stuff: https://getbootstrap.com/docs/4.4/utilities/flex/). Have a go at using that, they are super useful.

src/app/components/elements/modals/NotificationModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+                window.open(currentNotification.externalReference.url, "_blank");
+            }
+        }
+    }
+
+    return <React.Fragment>
+        <Col>
+            <Row className="justify-content-md-center mb-3">
+                <Col>
+                    {currentNotification ? currentNotification.value : "no content"}
```
mlt47, line 36: Let's hope we don't have cases where there is no content!
Also is the content going to be string only or is it an Isaac Content block?

src/app/components/navigation/IsaacApp.tsx
```diff
@@ -67,14 +77,24 @@ export const IsaacApp = () => {
...
     useEffect(() => {
         dispatch(requestCurrentUser());
         dispatch(requestConstantsSegueEnvironment());
         dispatch(fetchGlossaryTerms());
+        dispatch(requestNotifications());
     }, []);
 
+    useEffect(() => {
+        if (showNotification(user) && notifications && notifications.length > 0) {
+            dispatch(openActiveModal(notificationModal(notifications[0])));
```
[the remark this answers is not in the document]
bh412: On Physics we only show the first returned, but we do check for new ones every 5 seconds, so it would find a second, third etc quicker. We would but once per reload/ return to site, which I think would be quick enough for our purposes, and stops people being bombarded with pop ups?

src/app/components/elements/modals/NotificationModal.tsx
bh412: I'm going to have a think about this, in theory there's nothing stopping it being a content block, but it's one of content types in the editor that requires manual JSON editing, so whoever makes the notification would have to put in some effort to have anything more that value: "some string".

bh412: I've changed it to render IsaacContent, which should work as expected with a value that isn't just a string

PR #238, opened 2020-05-19
Feature/single assignment progress

bh412:
Allows teachers+ to open single assignment progress in a new tab. This can be favourited to return to if wanted. Allows refreshing for updates (a live event or something) without having to re-find the assignment

Spent far too much time essentially rewriting the assignment progress page. Stripped it all out and used exported code from it, tweaking allowed types to work. Moved shared types and functions out to AppTypes/ assignments service. 

Updates react-router-dom and it's types to 5.1.0, which seems to work and compile fine.


src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -410,14 +412,53 @@ const AssignmentDetails = (props: AssignmentDetailsProps) => {
                 <Button color="link">{isExpanded ? "Hide " : "View "} <span className="d-none d-md-inline">mark sheet</span></Button>
                 <span className="d-none d-md-inline">or</span>
                 <Button className="d-none d-md-inline" color="link" tag="a" href={getCSVDownloadLink(assignment._id)} onClick={openAssignmentDownloadLink}>Download CSV</Button>
+                <span className="d-none d-md-inline">or</span>
+                <Button className="d-none d-md-inline" color="link" tag="a" href={"/assignment_progress/" + assignment._id} onClick={openSingleAssignment}>View individual assignment</Button>
```
jsharkey13, line 416: Is there a reason this is not just a 
```
<Link to={`/assignment_progress/${assignment._id}`} target='_blank'>
```
and then styled correctly? The previous one needs to be funky because of the messy API URL, this one does not.

jsharkey13: It really shouldn't need an `onClick` either!

src/app/services/assignments.ts
```diff
@@ -1,10 +1,18 @@
...
 
 function notMissing<T>(item: T | undefined): T {
     if (item === undefined) throw new Error("Missing item");
     return item;
 }
+export function hasGameboard(assignment: AssignmentDTO): assignment is SingleEnhancedAssignment {
+    return assignment.gameboard != undefined;
+}
+export function getCSVDownloadLink(assignmentId: number) {
+    return API_PATH + "/assignments/assign/" + assignmentId + "/progress/download";
```
jsharkey13, line 14: Could this be the nice backtick syntax now?

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -401,6 +397,12 @@ const AssignmentDetails = (props: AssignmentDetailsProps) => {
         openActiveModal(downloadLinkModal(event.currentTarget.href));
     }
 
+    function openSingleAssignment(event: React.MouseEvent<HTMLAnchorElement>) {
```
jsharkey13, line 400: (See below, definitely shouldn't need this!)

src/app/components/pages/SingleAssignmentProgress.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    useEffect(() => {
+        setAssignment(myOwnedAssignments?.find(x => x._id == assignmentId));
+    }, [myOwnedAssignments, assignmentId]);
+
+    return <Container id={`single-assignment-${assignmentId}`} className="mb-5">
+        <ShowLoading until={assignment && assignmentProgress}>
+
+            <TitleAndBreadcrumb intermediateCrumbs={[ASSIGNMENT_PROGRESS_CRUMB]}
+                currentPageTitle={`${assignment?.gameboard?.title || "Single Assignment Progress" }`}
```
jsharkey13, line 66: I think "Assignment Progress" is a good default title, without the "Single", i.e, 
```
currentPageTitle={`Assignment Progress: ${assignment?.gameboard?.title}`}
```
does that look ok for real gameboards? What about the group name? :/

src/app/components/pages/SingleAssignmentProgress.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </div>;
+};
+
+export const SingleAssignmentProgress = () => {
+    const dispatch = useDispatch();
+    const params = useParams();
+    const assignmentId = params.assignmentId;
+
+    dispatch(loadProgress({_id: assignmentId}));
+    dispatch(loadAssignmentsOwnedByMe());
```
jsharkey13, line 40: Surely doing this defeats the point of this page? Since this loads everything and means this page is no more lightweight nor faster than the other page?

jsharkey13: Doesn't look like there's an alternative in the API, to be fair. And the endpoint this touches is a mess! A fix later job I think!

src/app/components/pages/AssignmentProgress.tsx
bh412: I found this seemed to be the only way to handle using preventDefault etc to stop it opening/closing the tab by clicking open single assignment

src/app/components/pages/SingleAssignmentProgress.tsx
bh412: Yep, as discussed in the call, I couldn't find a nicer way without changing how bits of the api work

bh412:
Corrected the site specificity and some other points, not sure where the duplicated code is, it reuses components from the assignment progress page?

mlt47:
Also worth spending a little time on the style of the new page.
I'm not a huge fan of the grey background when it is not in an accordion and we'll probably want to ignore the page gutter for this page too, as we do on the Assignment Progress page, so that it is easier to see all of the information in the data heavy table.

PR #243, opened 2020-05-26
Cs related questions

mlt47:


src/IsaacApiTypes.tsx
```diff
@@ -246,6 +246,7 @@ export interface ContentSummaryDTO {
     summary?: string;
     type?: string;
     level?: string;
+    difficulty?: string;
```
jsharkey13, line 249: I'm not sure I understand why this is a `string`. I get that it is the same as `level`, but it's not a Long or some other Java class that can't be represented as a JSON integer and so that's how they are sent to the frontend; why are they not both then `number`?

mlt47: In the API code `ContentSummaryDTO` defines `level` and `number` as strings so the type generation here is correct, but yes we should change it in the API if it is always going to be of type Long

jsharkey13: Ah, yes. That Java file has the incorrect types (it maps it from `Integer` in the real object's base class `SeguePageDTO`), but we can't change it without breaking Orika and ElasticSearch, so we'll live with it I guess.

PR #245, opened 2020-06-05
Update account information from required fields modal

mlt47:


mlt47:
> I'm not sure I understand why, but I can see that you've moved from `Object.assign(...)` to using the spread thing in a new object?

Yes they are functionally the same in these cases I just think it is a little easier to read the new syntax. They both create a new object (and object reference) which is what we care about here for the component to re-render.

Here's a little console program to confirm:
```
a = {key: 'value'};
b = {...a};
a == b; // returns false
```

PR #246, opened 2020-06-08
Enable 2FA for admin users

sacummins:
Must be merged after API changes https://github.com/isaacphysics/isaac-api/pull/271


src/app/components/elements/panels/UserMFA.tsx
```diff
@@ -0,0 +1,160 @@
...
+        }
+        authenticatorURL = `otpauth://totp/${myUser.email}?secret=${totpSharedSecret}&issuer=${issuer}`
+        QRCode.toString(authenticatorURL, {type:'svg'}, function (err, val) {
+            if (err) {
+                console.error(err)
+                return;
+            }
+            qrCodeStringBase64SVG = new Buffer(val).toString('base64');
+        });
+    }
```
Morpheu5, line 49: This whole segment looks like it would be better inside a `useEffect` for some sort, or otherwise it regenerates the QRCode every time the component is re-rendered.

src/app/components/pages/MyAccount.tsx
```diff
@@ -206,8 +213,8 @@ const AccountPageComponent = ({user, updateCurrentUser, getChosenUserAuthSetting
                                 className={classnames({"mx-2": true, active: activeTab === ACCOUNT_TAB.passwordreset})} tabIndex={0}
                                 onClick={() => setActiveTab(ACCOUNT_TAB.passwordreset)} onKeyDown={ifKeyIsEnter(() => setActiveTab(ACCOUNT_TAB.passwordreset))}
                             >
-                                <span className="d-none d-lg-block">Change password</span>
-                                <span className="d-block d-lg-none">Password</span>
+                                <span className="d-none d-lg-block">Security Settings</span>
```
jsharkey13, line 216: This needs to be `Security settings` to appease the CS team, but I think `Account security` might be nicer?

src/IsaacApiTypes.tsx
```diff
@@ -489,6 +490,12 @@ export interface AnsweredQuestionsByDate {
     [date: string]: number;
 }
 
+export interface TOTPSharedSecretDTO {
+    userId: number;
+    sharedSecret: string;
+    created: Date;
```
jsharkey13, line 496: There probably wants to be `lastUpdated` too, if we're sending that information. I think both `created` and `lastUpdated` are sent as Unix timestamps (`number`) though, so `Date` isn't right?

mlt47: It is the type generation which incorrectly marks these as Dates. There might be a config option for it.

src/app/components/elements/panels/UserMFA.tsx
```diff
@@ -0,0 +1,161 @@
...
+                                        <Input
+                                            id="setup-verification-code" type="text" name="setup-verification-code"
+                                            value={mfaVerificationCode || ""}
+                                            onChange={e => setMFAVerificationCode(e.target.value.replace(/ /g, ""))}
+                                        />
+                                    </FormGroup>
+                                    <FormGroup>
+                                        <Button
+                                            className="btn-secondary"
+                                            disabled={SITE_SUBJECT === SITE.PHY || !mfaVerificationCode}
```
jsharkey13, line 101: Is there a reason this is disabled for physics? I know it would lock you out of the old site, so I guess that's it?

jsharkey13: That is in fact the reason. Since this is only admin-facing, for now it doesn't matter.

PR #251, opened 2020-06-22
Pwned Password Checking Support

jsharkey13:
Using the Pwned Passwords API, implement functionality to allow online checking of password strength using k-anonymity to preserve privacy. Refactor the password feedback sections to simplify and make more extensible.

We do not actually use this functionality in this commit, merely add the code. Their API hasn't changed format since it was created, so that part won't go stale; and I'd like to use this at some point in the future.
To test, either just uncomment [line 107 of `passwordStrength.ts`]() to check locally and then update the result when the response returns, or also comment out line 106 to only use Pwned Passwords. A good test password is `guns and robots` which ZXCVBN rates as `Strong` but has been breached before and so will be rated "Weak" if Pwned Passwords is working.
We'd need to modify the CSP header to allow the requests off to `https://api.pwnedpasswords.com` if this was to be used for real.


jsharkey13 (commented):
I made some minor updates, but I have tested them and it works. This does not yet enable the Pwned Passwords API, and so is not really a functional change.

PR #252, opened 2020-06-24
Parsons problems indentation fixes

Morpheu5:
Adds show-on-hover buttons to change indentation via drag-and-drop, and improves look-and-feel of the drop action by removing the animation.


Morpheu5:
> * There is still one LGTM error when React state is directly modified rather than mutated using setState(...).

Was it fixed? Was it the draggedElement? I found sometimes problematic to `Object.assign` HTMLElements with event handlers and stuff attached, but this seems to work OK.

> * The new buttons seem useful but we should disable them if they are not applicable (i.e. can't go left as you are at 0 indentation or can't go right because you're already at the previous indentation + 1).

Done. Would it be useful to have an indication when there is too large a positive gap between one row and the next? Currently one can do something like

```A
    B
        C```

then de-indent `B` and so you end up with `C` being indented by more than one level over `B`. The code currently prevents form indenting `C` more than one over the previous item directly, but it might be hard to deal with indirect changes like this.

> * The other problem that people have been reporting with the question type is when you add an element from the options to either the top or the bottom of an existing answer list. Is there a way of increasing the drop area for those two regions when an element is being dragged?

I think the library uses the size of the element to determine the size of the gap.

File: src/scss/common/questions.scss [hunk 1]
```diff
@@ -174,15 +174,60 @@ button {
...
+
+          span {
+            display: inline-block;
+            width: 20px;
+            height: 20px;
+            cursor: pointer;
+
+            &.show:hover {
+              // Assuming this is only ever going to be used on CS
+              background-color: $cs-mustard !important;
```
mlt47, line 204: This breaks the Physics build. You can create a questions.scss in cs if you want it to be cs specific, or use $primary or $secondary which both sites have defined.

PR #256, opened 2020-06-30
Change search to work with multiple document types

mlt47:


src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
```diff
@@ -28,15 +34,35 @@ export const ContentSummaryListGroupItem = ({item, search, displayTopicTitle}: {
...
             iconLabel = "Concept page icon";
+            break;
+        case (DOCUMENT_TYPE.EVENT):
+            linkDestination = `/${documentTypePathPrefix[DOCUMENT_TYPE.EVENT]}/${item.id}`;
+            icon = "📆";
+            iconLabel = "Event page icon";
+            break;
+        case (DOCUMENT_TYPE.TOPIC_SUMMARY):
+            linkDestination = `/${documentTypePathPrefix[DOCUMENT_TYPE.TOPIC_SUMMARY]}/${item.id?.slice("topic_summary_".length)}`;
+            icon = "🗃️";
```
jsharkey13, line 53: Does this icon make sense? It looks really different on different devices, as does the event one. It'll probably do for now, but it would be nice to replace all these with something more sensible before the end of the summer!

src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
```diff
@@ -28,15 +34,35 @@ export const ContentSummaryListGroupItem = ({item, search, displayTopicTitle}: {
...
+            iconLabel = "Topic summary page icon";
+            topicTitle = "Topic"
+            break;
+        case (DOCUMENT_TYPE.GENERIC):
+            linkDestination = `/${documentTypePathPrefix[DOCUMENT_TYPE.GENERIC]}/${item.id}`;
+            icon = "ℹ️";
+            iconLabel = "Topic summary page icon";
+            break;
+        default:
+            linkDestination = `/unknown/${item.id}`;
```
jsharkey13, line 63: Is this sensible? Would it not be better not to show these?

src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
mlt47: Yeah they are temporary (although the other temporary ones have been around for over a year now!)

src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
mlt47: In theory it should never happen, but you're right maybe a better solution would be to log the error to the console and not render an item for that element in the list. I'll change it now.

PR #261, opened 2020-07-03
Feature/logo pods

bh412:
Centres pods and increases logo size for physics to match old site


src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -12,7 +12,7 @@ export const HeaderPhy = () => {
         <Container className="container-fluid px-0">
             <Row>
                 <Col>
-                    <div className="header-bar mx-3 mx-md-0 d-md-flex">
+                    <div className="header-bar mb-3 mx-3 mx-md-0 d-md-flex">
```
mlt47, line 15: On mobile, this adds a bit more margin under the logo? Do we want this? I think I prefer the old amount of space on mobile.

bh412: Good point, I had it in when the nav bar was the wrong side but didn't remove it. I've removed it and I think it looks ok still

bh412:
> Because you now set the background image to always be visible, you don't need the &:focus / &hover section to add the rule for background-image in those states. (Same online 55 new)

Good point!

PR #263, opened 2020-07-16
Feature/new svg icons

bh412:
- Nicki approves for now for phy also, should look to 'hexagonising' the Question icons (in future or as part of Ben cleaning up outstanding issues)


bh412:
I'm not sure we want alt text on icons, I could see it being useful perhaps on the search page, so you know if it's a question/concept etc, or if that would be too much noise? @jsharkey13 might know

PR #265, opened 2020-07-22
Pause playing video when changing tabs or accordion sections

bh412:


src/app/components/content/IsaacVideo.tsx
```diff
@@ -46,6 +46,11 @@ function onPlayerStateChange(event: any, wrappedLogAction: (eventDetails: object
     wrappedLogAction(logEventDetails);
 }
 
+export function pauseVideo() {
+    var iframe = document.getElementsByTagName("iframe")[0].contentWindow;
```
mlt47, line 50: This only works for the first iframe. It might also be worth checking that something is returned from getElementsByTagName("iframe") or it'll throw an exception.

mlt47: Best to use `const` or `let` over `var` in modern js.

src/app/components/content/IsaacVideo.tsx
```diff
@@ -46,6 +46,11 @@ function onPlayerStateChange(event: any, wrappedLogAction: (eventDetails: object
...
+    iframe && iframe.postMessage('{"event":"command","func":"pauseVideo","args":""}', '*');
```
mlt47, line 51: I know that in JSX we are limited to only using expressions so we use short-circuiting conditionals a lot, but in code like this where we don't have that restriction using statements is preferable as it is easier to read and think about.
```
if (iframe) {
    iframe.postMessage(....);
}
```

jsharkey13: (I strongly agree on this point, since the example isn't really ideomatic TypeScript either since the function is not a boolean value).

bh412:
I've had a look and the amended way I implemented seems to be the most painless - we're already using a useCallback in the iframe ref to attach to and use the youtube window.  I think it's going to be fairly difficult to start using another ref and storing it globally, unless I have misunderstood the purpose of this useCallback as a ref

src/app/components/content/IsaacVideo.tsx
```diff
@@ -47,6 +47,15 @@ function onPlayerStateChange(event: any, wrappedLogAction: (eventDetails: object
     wrappedLogAction(logEventDetails);
 }
 
+export function pauseVideo() {
```
mlt47, line 50: I would suggest altering this to pauseAllVideos(...) to be more descriptive of its functionality.

src/app/components/content/IsaacVideo.tsx
```diff
@@ -47,6 +47,15 @@ function onPlayerStateChange(event: any, wrappedLogAction: (eventDetails: object
...
+    const iframes = document.querySelectorAll('iframe');
+    Array.prototype.forEach.call(iframes, iframe => {
+        iframe.contentWindow.postMessage(JSON.stringify({ event: 'command',
+            func: 'pauseVideo' }), '*');
+    });
```
mlt47, line 55: A more standard variation on this would be to use something like the following:
```
iframes.forEach(iframe => {
        iframe?.contentWindow?.postMessage(JSON.stringify({event: 'command', func: 'pauseVideo'}), '*');
 });
```
Becasue iframes is a js Array it inherits the forEach method through the prototype chain.
Prototypical inheritance like they have in JS is different to how most languages achieve inheritance. We don't use it much in our programs but you might find it interesting to read about.

mlt47 (approved):
Some small suggestions, but feel free to merge this after that.
It is not a big feature and the implementation is short and sweet so I agree not to bother with the refs and context solution here.

PR #270, opened 2020-08-04
Drastically reduce the amount of re-renders while using Inequality

Morpheu5:
State was being set even though no actual change was present. Odd that React wouldn't detect that, but hey.


jsharkey13:
I think React does shallow equivalence checking, and so a new object won't match even though it's equal in value. I thought LGTM might object to the direct state access, but it didn't, so I merged it!

PR #273, opened 2020-08-11
Fix particle issues with chemistry equation editor

Morpheu5:


jsharkey13 (commented):
I don't think we need the extra menu items? The outputting of `e` as an available symbol for `electron` seems to already be fixed on master, and we don't actually want to have two sets of everything in the default menu.
What were they added for?

Morpheu5:
> I don't think we need the extra menu items? The outputting of `e` as an available symbol for `electron` seems to already be fixed on master, and we don't actually want to have two sets of everything in the default menu.
> What were they added for?

Are they actually getting generated? Those are simply to catch cases where `e` or `electron` are specified as available symbols. I don't filter duplicates because if a content editor enters a symbol twice with the two names, they deserve to have the duplicated symbol :)

mlt47 (approved):
Seems like small changes to the Equality page and adding some extra types.

PR #275, opened 2020-08-13
Add admin tool to merge two existing users

Scratchcat1:


jsharkey13:
It would also be helpful if (a) it asked for confirmation in the same way clicking the "Delete" button on a user does (just JavaScript), and (b) it said "User to Keep", and "User to Delete", rather than source and target, so that it cannot be mistaken :)

jsharkey13 (approved):

PR #276, opened 2020-08-13
Feature/specific doughnut colours

Scratchcat1:


src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -21,6 +21,21 @@ const OPTIONS = {
...
 };
 
+const colourPicker = (names: string[]): { [key: string]: string } => {
+    const selected = {} as { [key: string]: string };
+    let currentIndex = 0;
+
+    for (let i = 0; i < names.length; i++) {
+        if (names[i] in specificDoughnutColours) {
+            selected[names[i]] = specificDoughnutColours[names[i]];
+        } else if (currentIndex < doughnutColours.length) {
```
bh412, line 31: What happens if currentIndex exceeds doughnutColours.length? Do you know if this never would happen, or if it would default to a random colour?

Scratchcat1: This is expected to happen and can be seen on the physics page for the levels doughnut. For any data which doesn't have a colour specified the library randomly generates one.

PR #283, opened 2020-08-19
Upgrade Inequality to support XOR and add XOR to the menu

Morpheu5:


src/app/components/elements/modals/InequalityModal.tsx
```diff
@@ -502,13 +502,14 @@ export class InequalityModal extends React.Component<InequalityModalProps> {
     }
 
     private generateLogicFunctionsItems(syntax = 'logic'): MenuItem[] {
-        const labels: any = {
-            logic: { and: "\\land", or: "\\lor", not: "\\lnot", equiv: "=", True: "\\mathsf{T}", False: "\\mathsf{F}" },
-            binary: { and: "\\cdot", or: "+", not: "\\overline{x}", equiv: "=", True: "1", False: "0" }
+        let labels: any = {
```
Scratchcat1, line 505: `labels` doesn't appear to be reassigned but is assigned with `let` instead of `const`.

Morpheu5: Interesting, I wonder what prompted me to make that change… well spotted!

PR #285, opened 2020-08-20
Boolean logic grammar integration

Morpheu5:


jsharkey13 (changes requested):
So far:

 - this seems to submit a `formula` but the question type needs a `logicFormula`.
 - the click-to-edit box says "Enter your expression here" but the text entry box says "or toye your formula here". Both using "expression" would be nice?
 - I had expected `~A | ~B` and `not A or not B` to both parse, but the latter doesn't like the `not`s. Is that not supported?
 - the `badCharacters` probably doesn't want to allow anything beyond `A-Za-z&|01()~¬∧∨⊻` I think. I don't know if you support `+` for AND and `.` for OR, but if not we need to ensure they are not allowed.

jsharkey13:
This looks much better, but I found that if I typed an answer, then modified it using the editor, the text input box did not update and showed something inconsistent with the box above. Also when you reload the page, the text entry box does not get filled in with the previous answer text, only the equation editor box.

bh412 (approved):
Updates box correctly when drag/drop used. Multiple equations don't crash it, boolean logic still works. +/- works.

bh412:
James' changes have been addressed

PR #286, opened 2020-08-21
Add the isaac card content type in the design of the menu card.

Scratchcat1:
Adds a menu card replacement which supports card decks and groups.

Test page is at `/pages/_test_card_deck`.


Morpheu5 (approved):

mlt47 (changes requested):
This looks good so far.
We need to add a card deck object too for these content defined cards.

src/app/components/content/IsaacCard.tsx
```diff
@@ -0,0 +1,63 @@
+import React from "react";
+import {Card, CardBody, CardTitle, Col, Row} from "reactstrap";
+import * as ApiTypes from "../../../IsaacApiTypes";
```
Morpheu5, line 3: It's not particularly bad, but let's try to only import what we need. It help keeping things legible.

src/app/components/content/IsaacCardDeck.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import * as ApiTypes from "../../../IsaacApiTypes";
```
Morpheu5, line 2: Same as above.

Morpheu5 (approved):
Looks OK to me.

src/app/components/content/IsaacCardDeck.tsx
```diff
@@ -0,0 +1,14 @@
...
+import {IsaacCard} from "./IsaacCard";
+import {IsaacCardDeckDTO} from "../../../IsaacApiTypes";
+
+interface IsaacCardDeckProps {
+    doc: IsaacCardDeckDTO,
+    className?: string
+}
+
+export const IsaacCardDeck = ({doc, className}: IsaacCardDeckProps) => {
+    return <div className={"card-deck isaac-cards-body " + className}>
```
mlt47, line 11: In the case that the class name is left empty (as when a card deck is defined in the content) this div will receive an extra class "undefined".

Have you thought about how you would want content defined pages to show multiple rows of card decks? At the moment there would not be any gaps between each row which is probably not what we would want.
There are various ways around this -  you could add a css rule for .card-deck, add a css rule for .card-deck + .card-deck, give this component's className a default value of "mb-3", add a className propery to the card deck content object, or maybe it should be the card that has the margin?
What would be your suggestion?

mlt47, line 11: I've just noticed that bootstrap also has a .card-columns option for a Pinterest style card layout.
We don't need to do it now as we don't have a use case for it, but how would you suggest we accommodate that option if the content team were to request it?

mlt47 (changes requested):
Looks good. Have a think about how multiple card-deck rows should be presented to the user and the content editor.

src/app/components/content/IsaacCardDeck.tsx
Scratchcat1: `.card-deck + .card-deck` seems the best solution. We don't need to add any logic to detect if there are two consecutive card decks and the styling is only applied when required.

Scratchcat1: Since they both operate on a list of cards it would be a waste to define `IsaacCardColumns`. Adding a flag (or enum/string to also support card groups) would be more flexible.

mlt47: Seems a fine solution to me for now.

mlt47 (approved):

PR #289, opened 2020-08-24
Always show physics hints

bh412:


src/app/components/content/IsaacQuestion.tsx
```diff
@@ -257,8 +257,10 @@ export const IsaacQuestion = withRouter(({doc, location}: {doc: ApiTypes.IsaacQu
                 </RS.Col>
             </RS.Row>}
 
-            {SITE_SUBJECT === SITE.PHY && !validationResponse?.correct &&
-                <IsaacTabbedHints questionPartId={doc.id as string} hints={doc.hints} />
+            {SITE_SUBJECT === SITE.PHY &&
+                <div className={validationResponse && validationResponse.correct ? "mt-5" : ""}>
```
Morpheu5, line 261: What is wrong with `validationResponse?.correct ?`?

Morpheu5 (approved):
Otherwise good.

src/app/components/content/IsaacQuestion.tsx
bh412: You could end up with an undefined className, which the type system won't allow - It would be nice if it defaulted undefined to the equivalent of empty string

bh412: Good point, I'd done something else and forgotten to take it out - done now

PR #290, opened 2020-08-25
Correct text spacing for triple across menu cards

bh412:


Morpheu5 (approved):

jsharkey13 (approved):
I don't think this is the _best_ way to do it, perhaps one for @mlt47 to suggest a refactor for on his return, but it does work and so will do for now.

PR #294, opened 2020-09-02
Log clone board on new board creation

bh412:
If the board is successfully created, log the previous Id and the new Id


mlt47 (approved):
Nice and simple 👍

PR #295, opened 2020-09-03
Clean up available symbols so we don't end up with stuff like [''] which breaks menu generation

Morpheu5:


src/app/components/elements/modals/InequalityModal.tsx
```diff
@@ -281,7 +281,7 @@ export class InequalityModal extends React.Component<InequalityModalProps> {
                 ...defaultMenuItems
             }
         }));
-
+debugger;
```
jsharkey13, line 284: Whilst I personally think millions of debugger statements through the code are a good way to annoy anyone trying to use DevTools - we don't want this here!

jsharkey13 (changes requested):

src/app/components/elements/modals/InequalityModal.tsx
Morpheu5: 🤦

jsharkey13 (approved):

PR #299, opened 2020-09-14
Feature/my progress plus plus

Scratchcat1:
Add the following changes to the my progress page:
- Add github style heatmap instead of the activity graph.
- Add the most recently answered and easiest unsolved questions as links.
- Add currently set assignments to page.


jsharkey13:
We have no merged the replacement for this, and so no longer need this. But it was a helpful stepping stone to the final thing!

PR #302, opened 2020-09-17
Add header design with progress bar

bh412:
- Adds WebSocket support for header progress bar
- Uses react-circular-progress bar for neat header bar, billboard doesn't quite have the right text customisation
- Adds 'Hello name' for up to 20 characters to header
- Should be easy to add weekly streak updating via WebSocket as already generated in API but I think should be another small card once this is out/live and had feedback on


src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -59,7 +66,22 @@ export const HeaderPhy = () => {
                             <MainSearch />
                         </div>
                     </div>
-
+                </Col>
+                {user && user.loggedIn &&
```
Scratchcat1, line 70:
```suggestion
                {user?.loggedIn &&
```

src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -1,18 +1,22 @@
...
     const user = useSelector(selectors.user.orNull);
+    const streakRecord = useSelector((state: AppState) => state?.streakRecord ||
+        state?.userProgress?.userSnapshot?.dailyStreakRecord);
     const mainContentId = useSelector(selectors.mainContentId.orDefault);
     return <header className="light">
         <Container className="container-fluid px-0">
-            <Row>
-                <Col>
+            <Row className="align-items-center">
+                <Col md={user && user.loggedIn ? 11 : 12}>
```
Scratchcat1, line 19:
```suggestion
                <Col md={user?.loggedIn ? 11 : 12}>
```

src/app/components/navigation/IsaacApp.tsx
```diff
@@ -95,7 +97,8 @@ export const IsaacApp = () => {
     useEffect(() => {
         if (isLoggedIn(user)) {
             dispatch(requestNotifications());
-            checkForWebSocket();
+            checkForWebSocket(user, userSnapshot);
+            dispatch(getProgress());
```
mlt47, line 101: Is the call to getProgress() necessary? I'm guessing it is to populate userSnapshot but as userSnapshot is not a dependency of this effect it won't get rerun when the snapshot is there.

src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -1,18 +1,22 @@
...
 import {selectors} from "../../../state/selectors";
+import {AppState} from "../../../state/reducers";
+import {HeaderDailyStreakGauge} from "../../elements/views/DailyStreakGauge";
 
 export const HeaderPhy = () => {
     const user = useSelector(selectors.user.orNull);
+    const streakRecord = useSelector((state: AppState) => state?.streakRecord ||
+        state?.userProgress?.userSnapshot?.dailyStreakRecord);
     const mainContentId = useSelector(selectors.mainContentId.orDefault);
     return <header className="light">
```
mlt47, line 16: As Connor mentioned, on mobile the Menu button is now out-of-place.

src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -42,7 +46,10 @@ export const HeaderPhy = () => {
                                     <React.Fragment>
                                         <div className="my-account mx-5 mx-sm-2">
                                             <Link to="/account">
-                                                <span>MY ACCOUNT</span>
+                                                {user.givenName && user.givenName.length <= 20 &&
+                                                    <strong className="d-none d-md-block">Hello {user.givenName}</strong>
+                                                }
+                                                <span>YOUR ACCOUNT</span>
```
mlt47, line 52: Nomensa's designs also favoured this to the person shaped icon on mobile. It would improve Physics' header because the person icon is currently misaligned.

src/app/components/navigation/IsaacApp.tsx
```diff
@@ -95,7 +97,8 @@ export const IsaacApp = () => {
     useEffect(() => {
         if (isLoggedIn(user)) {
             dispatch(requestNotifications());
-            checkForWebSocket();
+            checkForWebSocket(user, userSnapshot);
```
mlt47, line 100: Do the websockets need to be cleaned up when the user changes?

src/app/state/reducers.tsx
```diff
@@ -478,6 +481,26 @@ export const currentGameboard = (currentGameboard: CurrentGameboardState = null,
...
+export type StreakRecordState = StreakRecord | null;
+export const streakRecord = (streakRecord: StreakRecordState = null, action: Action) => {
+    switch (action.type) {
+        case ACTION_TYPE.STREAK_RECORD_UPDATE:
+            return action.streakRecord;
+        default:
+            return streakRecord;
+    }
+};
+
```
mlt47, line 503: Why do we have separate user snapshot state here if it also exists in userProgress?

mlt47 (changes requested):
A few suggestions with the main ones being how it looks on devices of different sizes, checking that the useEffect is doing what you think it is doing and having a think about where the data should live in redux.

src/app/state/reducers.tsx
bh412: We never post to update userProgress, it's generated API side and returned on request - I think having this as a separate thing that we update on the action and store in redux makes more sense in this case as it is given to us by the websocket as it's own object whenever the snapshot is updated, as opposed to altering the API to have it update the userProgress then return that by the websocket, changing the websocket? I might be misunderstanding

src/app/components/navigation/IsaacApp.tsx
bh412: On user logout and user_consistency_error we close the websocket, I think that is enough?

mlt47: Ah yes, I see that in the consistency checker now.

I would prefer it if websocket closure was handled by this useEffect's clean up callback.
In that case, the user consistency checker will change the user which will, in turn, cause this useEffect's clean up callback.
This is nicer because the websocket connections are handled here, the user consistency checking by the user consistency middleware. There's a better separation of concerns.
If we were to hypothetically remove the user consistency checker it is likely that we would forget that that middleware was also used to manage websocket closures on logout, its not really the user consistency checker's "one job" to handle websockets, it's just another job that it has been lumped with because it was convenient.

bh412: I see, that makes sense, done that, checking it works as expected

src/app/state/reducers.tsx
bh412: I've altered the way it works to modify the userProgress in the Redux state

src/app/components/navigation/IsaacApp.tsx
bh412: I've found it needs to be there for the initial population of userSnapshot, otherwise you get no progress until you visit the myProgress page/attempt a question. It doesn't need to be recalled as the webSocket for the userSnapshot then updates the progress

mlt47:
Hi Ben here are the changes I would suggest, as we talked through on Google meet. I implemented the first three suggestions in https://github.com/isaacphysics/isaac-react-app/tree/feature/hello-ben-header to make sure the changes were possible/sensible, up to you if you want to keep them or not.

- We should remove `streakRecord` and `userSnapshot` redux state as it is duplicating information already stored in `userProgress.userSnapshot` - redux's top-level `userSnapshot` wasn't actually being used for anything in the application.
- Cange the new action type name to `USER_SNAPSHOT_PARTIAL_UPDATE` (and remove the duplicated action, type, etc. as they are both doing the same thing).
- Let's not gispatch `userProgress()` in IsaacApp. WebSockets will update the progress bar and we're quite sure almost all user browsers will support websockets. If we receive complaints we should create a more lightweight endpoint asking only for the streak data.
- In the API PR `getDetailedUserStatistics()` returns `"streakRecord"` and `"weeklyStreakRecord"` - we won't handle the first properly because we now expect `"dailyStreakRecord"`.
- Merge conflict.
- Move tool-tip from HeaderPhy and CS to the HeaderStreakGague (or fix its contents in physics - still refers to the day streak).
- We should capitalize "daily streak" on the My Progress page.
- Progress bar style improvements (Exo font? fatter progress bar).
- I'd be tempted to rollback the CS streaks counter as it needs more time to get the style looking nice and we haven't let them know it is going to be a new feature!
- Rename the `progressCoulours` variable to `progressColour`.
- Worth double checking the CSS changes don't break existing CSS as there is quite a lot of it.

bh412:
I think I've implemented the suggestions - @mlt47 , locally it freezes the tab with dev tools open and hovering over the header progress bar to open the tooltip, I can't find an Id clash and it doesn't throw errors, maybe a rendering bug?

mlt47 (approved):

PR #303, opened 2020-09-21
Allow event managers to send event type admin emails

Scratchcat1:


bh412 (approved):
Nice and simple, easy to alter if we expand to staff

PR #305, opened 2020-09-21
Automatically log out and display a warning when receiving a 412 precondition not met

Scratchcat1:


jsharkey13:
We won't be using the 412 Precondition Failed response any time soon, so I am closing this. But it will be useful to refer to this for the `X-Session-Expires` header stuff, since the `handleSessionInvalidation(...)` method could be useful!

PR #307, opened 2020-09-23
Remove invisible encoding character from cookie banner which converts…

Scratchcat1:
… it into an emoji

Invisible variation selector 16 was causing the i symbol to be rendered as an emoji.


jsharkey13 (approved):
I would say this looks good to me, but I think the point is that I can't see the difference :P

(This did not need to be a PR, by the way, since it's small, clear and agreed in advance).

PR #308, opened 2020-09-24
 Fix layout issues on set assignments and add groups modal 

Scratchcat1:
- Fix touching buttons on narrow screens on the add groups modal.
- Fix unaligned cards on set assignments page


Morpheu5 (approved):

src/scss/common/assignments.scss
```diff
@@ -0,0 +1,4 @@
+.assignments-grid {
```
bh412, line 1: Keeps the cards the same height but when 'assign/ unassign' clicked all row expands to match

src/app/components/pages/SetAssignments.tsx
```diff
@@ -144,7 +144,7 @@ const Board = (props: BoardProps) => {
     const boardSubjects = determineGameboardSubjects(board);
     const boardLevels = determineGameboardLevels(board);
 
-    return <Card className="board-card">
+    return <Card className="board-card h-100">
```
bh412, line 147: Not sure we want this to be the set height here

src/app/components/elements/modals/GroupsModalCreators.tsx
```diff
@@ -59,23 +59,23 @@ export const groupInvitationModal = (firstTime: boolean) => {
         body: <CurrentGroupInviteModal firstTime={firstTime} />,
         buttons: [
             <RS.Row key={0}>
-                <RS.Col>
+                <RS.Col className="my-2">
```
bh412, line 62: I think this modal is actually fine, and it's the 'GameboardCreatedModal' that needed work, but I've done this change to it

bh412 (commented):
Mostly works but set assignments has the expanding card quirk so going to move this to won't fix too minor - A nice implementation though and I'll see if there are any other card bits it should be applied to

bh412:
We've since accomplished this in another way so this is redundant.

PR #310, opened 2020-09-25
Try to stop full URLs as gameboard IDs

bh412:
Simple check for "http" being manually entered into the gameboard ID if provided by a staff member. Won't affect http being present in random generated ID from backend if no ID provided by staff.

### DIFF
--- a/src/app/services/validation.ts
+++ b/src/app/services/validation.ts
@@ -20,7 +20,7 @@ export const validateEmail = (email?: string) => {
 };
 
 export const validateUrl = (url?: string) => {
-    return !url || !url.includes("http");
+    return !url || /^[a-z0-9_-]+$/.test(url);
 };
 
 export const isDobOverThirteen = (dateOfBirth?: Date) => {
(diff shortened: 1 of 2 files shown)


jsharkey13 (changes requested):
This still allows invalid IDs, and blocks valid ones. Just blocking anything not matching something like `^[a-z0-9_-]+$` would be better, I think. Not sure if we should allow uppercase letters, there seems no reason to even if they won't break anything.

(Please test that regex, and don't just trust something I made up as I went!)

src/app/services/validation.ts
```diff
@@ -19,6 +19,10 @@ export const validateEmail = (email?: string) => {
     return email && email.length > 0 && email.includes("@");
 };
 
+export const validateUrl = (url?: string) => {
```
jsharkey13, line 22: Sorry to be a further pain, but could this be `isValidGameboardId`?

jsharkey13: And take `gameboardId` as the argument?

jsharkey13: I think that's the correct way round, it returns true if valid, which is undefined or matching?

jsharkey13 (changes requested):

src/app/services/validation.ts
bh412: Done!

jsharkey13 (approved):

PR #311, opened 2020-09-25
Add a banner to events card images with a description and icons to th…

Scratchcat1:
…e events cards

- Adds a banner to the event card image which describes if it is a teacher student or virtual event.
- Adds icons to the bottom of the card if the card is tagged teacher, student or virtual.


bh412 (approved):
Fixed an issue with the icons of safari/chrome but looks nice otherwise

PR #313, opened 2020-10-02
Group progress page

Morpheu5:
Adds a `/group_progress` page where group owners can see the progress overview of their students on all the assignments all in one go. This is different from `/assignment_progress` in that it doesn't show details of the assignments but rather an aggregated summary.


jsharkey13:
- Are we not implementing the "total" column, like we have for the `/assignment_progress` page?
 - How do we decide whether to show question parts or question pages in the column counts?
 - Can we make sure that we don't make the request for a group's progress until the user clicks the accordion section? With HTTP/2 that will send many many hugely expensive simultaneous requests on page load, and that will cause the server to fall over. I also tried it as Lisa's account locally (78 groups) and the whole page was unresponsive for many minutes waiting for all the requests to finish before I could do anything. Loading spinners (use breakpoints to artificially delay requests for testing) would be nice if not present, too!

bh412:
3 things from the content team @Morpheu5:
1) For the percentage completion row, in between assignment name and the first person's result, can it be made grey unless that column is focussed on, like the Assignment Progress page.
2) Can we modify the assignment ordering - make it due date order ascending from the left, then if it doesn't have a due date set, use the default ordering (creation date) ascending.
3) Make sure we only display a sensible number of assignments before they have to become visible by scrolling using the arrows in the header bar - teachers may have 20, 50, 100 assignments set, so only display an amount that we have width for at once, and click the arrows to rotate through. Content would like the first (name) and last 2 (assignments and q completed) cols to be frozen and always visible, so it's the middle chunk of assignments which are to scroll through
4) Choose a fixed width for the assignment columns and make the name column wider if needs be for few assignments

bh412 (approved):
Works as expected by content team after back and forth and meetings

PR #314, opened 2020-10-05
Add hint logging for physics hint tabs with hint index

bh412:
- Resumes logging which hints are viewed now they are hints
- Doesn't conflict with the CS hint modal logging

### DIFF
--- a/src/app/components/content/IsaacHints.tsx
+++ b/src/app/components/content/IsaacHints.tsx
@@ -1,13 +1,12 @@
 import {ListGroup, ListGroupItem} from "reactstrap";
 import {IsaacHintModal} from "./IsaacHintModal";
-import React, {useEffect, useState} from "react";
+import React from "react";
 import {ContentDTO} from "../../../IsaacApiTypes";
 import {IsaacContent} from "./IsaacContent";
 import {AppState} from "../../state/reducers";
 import {useDispatch, useSelector} from "react-redux";
 import {Tabs} from "../elements/Tabs";
 import {logAction} from "../../state/actions";
-import {isDefined} from "../../services/miscUtils";
 
 const PrintOnlyHints = ({hints}: {hints?: ContentDTO[]}) => {
     const printHints = useSelector((state: AppState) => state?.printingSettings?.hintsEnabled);
--- a/src/app/components/content/IsaacHints.tsx
+++ b/src/app/components/content/IsaacHints.tsx
@@ -1,11 +1,13 @@
 import {ListGroup, ListGroupItem} from "reactstrap";
 import {IsaacHintModal} from "./IsaacHintModal";
-import React from "react";
+import React, {useEffect, useState} from "react";
 import {ContentDTO} from "../../../IsaacApiTypes";
 import {IsaacContent} from "./IsaacContent";
 import {AppState} from "../../state/reducers";
-import {useSelector} from "react-redux";
+import {useDispatch, useSelector} from "react-redux";
 import {Tabs} from "../elements/Tabs";
+import {logAction} from "../../state/actions";
+import {isDefined} from "../../services/miscUtils";
 
 const PrintOnlyHints = ({hints}: {hints?: ContentDTO[]}) => {
     const printHints = useSelector((state: AppState) => state?.printingSettings?.hintsEnabled);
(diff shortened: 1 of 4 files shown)


src/app/components/elements/Tabs.tsx
```diff
@@ -22,7 +23,7 @@ function callOrString(stringOrTabFunction: StringOrTabFunction, tabTitle: string
 }
 
 export const Tabs = (props: TabsProps) => {
-    const {className="", tabTitleClass="", tabContentClass="", children, activeTabOverride, activeTabChanged, deselectable=false} = props;
+    const {className="", tabTitleClass="", tabContentClass="", children, activeTabOverride, activeTabChanged, deselectable=false, setViewedHint} = props;
```
jsharkey13, line 26: Several issues here;
 - `setViewedHint` is a poor name. This is a callback function that gets called when a tab is clicked. Let's call it `onTabClick`, since `onClick` is revered by the HTML spec for something slightly different.
 - `activeTabChanged` is already a callback for a function that gets called when a tab is clicked. We don't need a second one to implement this! But it also has a terrible name, so refactoring `activeTabChanged` to either `onTabClick` or `onActiveTabChange` would be great. Then you can use it in `IsaacHints`.

src/app/components/elements/Tabs.tsx
```diff
@@ -32,6 +33,7 @@ export const Tabs = (props: TabsProps) => {
             nextTabIndex = -1;
         }
         setActiveTab(nextTabIndex);
+        setViewedHint && nextTabIndex !== -1 && setViewedHint(nextTabIndex);
```
jsharkey13, line 36: If you do refactor to fix this duplication, note that you'll need to do the filtering for the `-1` special case in the `IsaacHints` callback function, since some places do need the `activeTabChanged` callback to be called on the `-1` case!

src/app/components/content/IsaacHints.tsx
```diff
@@ -34,9 +36,19 @@ export const IsaacLinkHints = ({hints, questionPartId}: HintsProps) => {
     </div>;
 };
 
-export const IsaacTabbedHints = ({hints}: HintsProps) => {
+export const IsaacTabbedHints = ({hints, questionPartId}: HintsProps) => {
+    const dispatch = useDispatch();
+    const [viewedHint, setViewedHint] = useState();
+
+    useEffect(() => {
```
jsharkey13, line 43: I don't think this is sensible? I don't see why React is needed here at all, you just want a function that takes a hint index and does the logging:
```typescript
    function onHintView(viewedHintIndex: number) {
        if (viewedHintIndex > -1) {
            const eventDetails = {type: "VIEW_HINT", questionId: questionPartId, hintIndex: viewedHintIndex};
            dispatch(logAction(eventDetails));
        }
    }
```

jsharkey13 (changes requested):
You've re-implemented something that already exists!

@mlt47 might like to weigh in on the React-ness, but I think it doesn't need the useEffect weirdness either.

src/app/components/elements/Tabs.tsx
jsharkey13: I meant "reserved" here, not "revered", but actually it kinda works either way :P

jsharkey13: I wonder if `onTabClick` might actually not be clear enough (it's the button being clicked, not anywhere in the tab), but `onActiveTabChange` implies that the tab actually changed, and I think it might get called even if you don't change the tab but just click the button again?
Maybe the function should take `(tabIndex: number, previousTabIndex?: number)` which is safe and backwards compatible with the existing uses. Then you can log only when the two are not the same. Or, and this might need testing, you could suppress the callback if the active tab didn't change and then call it `onActiveTabChange` and be done with it!

src/app/components/elements/Tabs.tsx
```diff
@@ -33,9 +32,8 @@ export const Tabs = (props: TabsProps) => {
             nextTabIndex = -1;
         }
         setActiveTab(nextTabIndex);
-        setViewedHint && nextTabIndex !== -1 && setViewedHint(nextTabIndex);
-        if (activeTabChanged) {
-            activeTabChanged(nextTabIndex);
+        if (onActiveTabChange && nextTabIndex !== -1) {
```
jsharkey13, line 35: I think we need this callback even when the tab is `-1`, just in case, i.e. we should call the callback if `activeTab !== nextTabIndex`, instead of `nextTabIndex !== -1`.

jsharkey13 (approved):

PR #315, opened 2020-10-05
Use all the tags in /glossary topic filter

Morpheu5:
Previously, the topic filter only showed the topics associated to the terms currently showing, and that was wrong.

This also fixes the case issue by using the actual tag names instead of startcasing their IDs.

Certain topic tags can have parentheses in their names, so this avoids stuff like

> (Used in: GUIs (OCR))


src/app/components/pages/Glossary.tsx
```diff
@@ -143,6 +137,15 @@ export const Glossary = withRouter(({ location: { hash } }: GlossaryProps) => {
...
 
+    const getTags = (docTags?: string[]) => {
+        if (SITE_SUBJECT !== SITE.CS) {
+            return [];
+        }
+        if (!docTags) return [];
+
+        return (docTags as TAG_ID[]).map(id => tags.getById(id)); //tagHierarchy.map(tag => ({title: tag.title}));
+    }
+
```
mlt47, line 148: I would be tempted to move this into the IsaacGlossaryTerm component then you can remove its `tags` prop.

src/app/components/pages/Glossary.tsx
```diff
@@ -169,11 +172,11 @@ export const Glossary = withRouter(({ location: { hash } }: GlossaryProps) => {
...
-                                    { filterTopic === "" ? "Filter by topic" : _startCase(filterTopic) }
+                                    { !isDefined(filterTopic) ? "Filter by topic" : filterTopic.title }
                                 </DropdownToggle>
                                 <DropdownMenu>
-                                    <DropdownItem onClick={() => setFilterTopic("")}>All topics</DropdownItem>
-                                    {topics.map(e => <DropdownItem key={e} onClick={() => setFilterTopic(e)}>{_startCase(e.replace(/[^a-zA-Z0-9]/, ' '))}</DropdownItem>)}
+                                    <DropdownItem onClick={() => setFilterTopic(undefined)}>All topics</DropdownItem>
+                                    {topics.map(e => <DropdownItem key={e.id} onClick={() => setFilterTopic(e)}>{e.title}</DropdownItem>)}
                                 </DropdownMenu>
                             </Dropdown>}
```
mlt47, line 181: I would express a slight preference for react-select over reactstrap's Dropdown. Dropdown looks a little clunky, particularly on mobile. React-select does not need to be multi-value (to keep the current filtering logic) and you then get the optional autocomplete from typing in the field.
I'm not sure if we would then prefer to add the fields groupings like we have on the teacher gameboard builder, if they are scanning the long list to find a topic it would probably help.

mlt47 (changes requested):
This is sensible and seems to be simplifying the code which is always good!
I have made two small suggestions, would be happy to hear your view.

src/app/components/pages/Glossary.tsx
```diff
@@ -10,55 +10,51 @@ import {PrintButton} from "../elements/PrintButton";
...
-    const [topics, setTopics] = useState<string[]>([]);
-    const [filterTopic, setFilterTopic] = useState("");
-    const [topicsDropdownOpen, setTopicsDropdownOpen] = useState(false);
-
+    const topics = tags.allTags.sort((a,b) => a.id === b.id ? 0 : (a.id > b.id ? 1 : -1));
+    const [filterTopic, setFilterTopic] = useState<Tag>();
     const rawGlossaryTerms = useSelector((state: AppState) => state && state.glossaryTerms);
 
     const glossaryTerms = useMemo(() => {
         if (searchText === '') {
```
mlt47, line 36: Did you consider doing the exam board filtering here as well as in the Glossary Term component? I guess it always needs to be done in the glossary term component as that is not just used on the Glossary page. The only advantage of also doing it here would be to make the alphabet selector reflect the exam board filter. Won't be noticeable for most cases but would be if there was a Z term with only an entry for AQA, for example.

mlt47 (changes requested):
A few more small suggestions/questions to consider before merging.

src/app/components/pages/Glossary.tsx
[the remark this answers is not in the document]
Morpheu5: Been meaning to do that and then forgot. Thanks.

src/app/components/pages/Glossary.tsx
```diff
@@ -167,15 +163,14 @@ export const Glossary = withRouter(({ location: { hash } }: GlossaryProps) => {
                         </Col>
                         <Col className="mt-3 mt-md-0">
```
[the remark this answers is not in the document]
Morpheu5: We could probably do away with the alphabet but I can see how the search and exam board filters might be desirable. Trouble is, I don't know why they aren't showing up.

mlt47 (approved):
Looks good.

I would consider moving examBoard from being a property of Glossary Terms to being recorded as a tag, as that is what we do with all other content which is exam board specific. It would also make things easier in the case that we were to support a third exam board. The logic would not need to change if userExamboard in ["aqa", "wjec"] but not "ocr" for example, where as currently you are using userExamboard == "" or undefined to mean available to all.

PR #316, opened 2020-10-06
Add a warning for users typing unclear function calls in text entry

jsharkey13:
If a user types "Tsin(theta)" this will be parsed as "T\*s\*i\*n\*theta", which is not what the user intended. Whilst this is visible, in the fact that the result has italic and not upright letters for the sin in the preview, this fact is not obvious to most users. Instead add a warning underneath.
We have to allow "a" and "c" as characters before to allow "asin(x)" or "arcsin(x)", and this seems the simplest regex to achieve this.

### DIFF
--- a/src/app/components/content/IsaacSymbolicQuestion.tsx
+++ b/src/app/components/content/IsaacSymbolicQuestion.tsx
@@ -180,7 +180,12 @@ const IsaacSymbolicQuestionComponent = (props: IsaacSymbolicQuestionProps) => {
                 }
                 setErrors(errors);
             } else {
-                setErrors(undefined);
+                if (/[A-Zbd-z](sin|cos|tan|log|ln|sqrt)\(/.test(pycode)) {
+                    // A warning about a common mistake naive users may make (no warning for asin or arcsin though):
+                    setErrors(["Make sure to use spaces or * signs before function names like 'sin' or 'sqrt'!"])
+                } else {
+                    setErrors(undefined);
+                }
                 if (pycode === '') {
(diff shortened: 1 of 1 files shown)


src/app/components/content/IsaacSymbolicQuestion.tsx
```diff
@@ -180,7 +180,12 @@ const IsaacSymbolicQuestionComponent = (props: IsaacSymbolicQuestionProps) => {
                 }
                 setErrors(errors);
             } else {
-                setErrors(undefined);
+                if (/[A-Zbd-z](sin|cos|tan|log|ln|sqrt)\(/.test(pycode)) {
```
Morpheu5, line 183: According to this, `csqrt` is a valid function.

bh412: The regex could maybe be altered to only test the trig functions this way, maybe like `([A-Zbd-z](sin|cos|tan)|[A-Za-z](log|ln|sqrt))`?

jsharkey13: Yes, I was going for simplicity of regex, rather than completeness or correctness of values. Really, so long as the warning happens enough users are aware of the issue, we're good. I have no objections to Ben's regex, although if we go down that route perhaps the regex should actually be:
```javascript
/([A-Za-z](log|ln|sqrt)|[A-Zbd-z](sin|cos|tan|cosec|sec|cot))\(/.test(...)
```

bh412 (approved):
Going with the quick regex to catch all cases, works as expected

PR #318, opened 2020-10-07
Report client errors to the server as a Log Event

mlt47:
The motivation for these changes were for improving monitoring and alerting on client errors.

### DIFF
--- a/src/app/components/pages/ClientError.tsx
+++ b/src/app/components/pages/ClientError.tsx
@@ -1,28 +1,41 @@
-import React from "react";
+import React, {useEffect} from "react";
 import {Link} from "react-router-dom";
 import {Col, Container, Row} from "reactstrap";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
 import ReactGA from "react-ga";
 import {WEBMASTER_EMAIL} from "../../services/siteConstants";
 import {FallbackProps} from "react-error-boundary";
-import {useSelector} from "react-redux";
+import {useDispatch, useSelector} from "react-redux";
 import {selectors} from "../../state/selectors";
+import {logAction} from "../../state/actions";
 
-export const ClientError = ({resetErrorBoundary, error, componentStack}: FallbackProps) => {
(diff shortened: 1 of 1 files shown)


src/app/components/pages/ClientError.tsx
```diff
@@ -1,28 +1,41 @@
...
     }
+
+    useEffect(() => {
+        const {userId, ...informationNotIncludingUserId} = usefulInformation;
+        dispatch(logAction({...informationNotIncludingUserId, type: "CLIENT_SIDE_ERROR"}));
+    }, []);
+
     const plainTextUsefulInformation = "\n\n---- Useful Error Information ----\n\n" + Object.entries(usefulInformation)
-        .map(([label, value]) => `${label}: ${value}`).join("\n\n")
+        .map(([key, value]) => `${usefulInformationLabels[key as keyof typeof usefulInformation]}: ${value}`).join("\n\n")
```
Morpheu5, line 38: `keyof` is cool, but I'm not sure I understand it side by side with `typeof`. What is the semantic here?

Morpheu5: Ah, no, I think I get it: it's because we don't know in advance the "type" (i.e., the keys) of `usefulInformation`?

Morpheu5 (approved):

src/app/components/pages/ClientError.tsx
mlt47: Yes, you're correct, usefulInformation is a value not a type and I think keyof only operates on types.

It is a little annoying that this is necessary because in line 25 I'm sort of telling the compiler that they have the same fields but it only knows that the keys are strings by this point. There might be a solution using a string enum to avoid the `as` cast, but I'm not fully convinced it would work, without trying it.

Morpheu5: Have you checked that this isn't a known bug by any chance? Much of the unnecessary work I did early on on Inequality came down to bugs and features not implemented yet in Typescript.

PR #319, opened 2020-10-12
Type guard user service

jsharkey13:
The LoggedInUser type was not actually a logged in user, and so you could not make a type assertion in services/user.ts about the return value of `isLoggedIn(user)`. Move to PotentialUser as the type for a user that might be logged in, and use LoggedInUser as a wrapper to a RegisteredUserDTO for an actual user object.

May be best to review as two separate commits, since git can't do the diff correctly for the renamed `.tsx` -> `.ts` file.


mlt47 (approved):
This looks sensible

PR #323, opened 2020-10-20
Adds COVID-19 Lessons cards to gcse and alevel pages

bh412:
Add covid lessons cards to gcse and alevel pages

### DIFF
--- a/src/app/components/pages/Gcse.tsx
+++ b/src/app/components/pages/Gcse.tsx
@@ -26,20 +26,26 @@ export const Gcse = () => {
             }}/>
         </Row>
         <Row className="card-deck isaac-cards-body my-3">
-            <IsaacCard doc={{title: "Mentoring Scheme", subtitle: "Weekly gameboards guided by an experienced teacher.",
-                clickUrl: "/pages/isaac_mentor", image: {src: "/assets/phy/key_stage_sprite.svg#teacher-hat"}
+            <IsaacCard doc={{title: "COVID-19 Lessons", subtitle: "Virtual lessons following the Mastering Essential GCSE Physics book.",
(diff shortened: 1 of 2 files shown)


src/app/components/pages/Gcse.tsx
```diff
@@ -26,20 +26,26 @@ export const Gcse = () => {
...
+            }}/>
         </Row>
         <Row className="card-deck isaac-cards-body mb-5 mt-3">
-            <IsaacCard doc={{title: "Quick Quizzes", subtitle: "Revise the equations needed at GCSE with our quick quizzes.",
-                clickUrl: "/pages/gcse_quizzes", image: {src: "/assets/phy/key_stage_sprite.svg#lightning"}
-            }}/>
             <IsaacCard doc={{title: "Events", subtitle: "Find one of our face-to-face or virtual workshops.",
                 clickUrl: "/events?types=student", image: {src: "/assets/phy/teacher_features_sprite.svg#calendar"}
             }}/>
+            <Col md={6}/>
```
jsharkey13, line 48: What's this empty column doing?

bh412: It stops the single card being full width

bh412: Is that too quick and nasty without reworking the card type?

jsharkey13: Perhaps a `//FIXME: ...` comment explaining as much, so that it's clear to anyone else too?

jsharkey13: Ah, it'll need to be `/* FIXME: ... */}` in TSX.

PR #324, opened 2020-10-22
Add wildcard for phys and maths book boards

bh412:
- Makes in progress/perfect etc messages black on physics
- Adds wildcard
- Adds regex to check if wildcard should be shown

### DIFF
--- a/src/app/components/pages/Gameboard.tsx
+++ b/src/app/components/pages/Gameboard.tsx
@@ -13,7 +13,7 @@ import {Redirect} from "react-router";
 import {SITE, SITE_SUBJECT} from "../../services/siteConstants";
 import tags from "../../services/tags";
 import {selectors} from "../../state/selectors";
-import {detetmineWildcardSubject, showWildcard} from "../../services/gameboards";
+import {showWildcard} from "../../services/gameboards";
 
 function getTags(docTags?: string[]) {
     if (SITE_SUBJECT !== SITE.PHY) {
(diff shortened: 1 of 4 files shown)


src/app/services/gameboards.tsx
```diff
@@ -85,6 +85,30 @@ export const generateGameboardSubjectHexagons = (boardSubjects: string[]) => {
     );
 };
 
+export const showWildcard = (board: GameboardDTO) => {
+    const re = new RegExp('(phys_book_gcse_ch.*|pre_uni_maths.*)');
+    return board?.id && re.test(board.id)
+};
+
+export const detetmineWildcardSubject = (wildcard: IsaacWildcard) => {
```
jsharkey13, line 93: Why are we marking a wildcard as a colour at all? Surely it is less confusing if it is in black and thus more visually distinct? This could lead to, say, maths concept pages on physics boards being purple and not blue. At least black gives no opinions!

src/app/components/pages/Gameboard.tsx
```diff
@@ -78,11 +76,33 @@ const gameboardItem = (gameboard: GameboardDTO, question: GameboardItem) => {
     </RS.ListGroupItem>;
 };
 
+export const Wildcard = (wildcard: IsaacWildcard) => {
+    const itemClasses = "p-3 content-summary-link text-info bg-transparent";
+    const icon = <img src="/assets/wildcard.svg" alt=""/>;
```
jsharkey13, line 81: Leaving aside the issue that the questions need some alt text for their icons given that we're using colour to distingush things, this wildcard is only distinct because of this icon. So it needs altText; perhaps "Supplementary material link icon", which is pushing the altText rules but I think the most useful.

jsharkey13 (commented):
Seems to work ok, just some questions about intended behaviour

src/app/components/pages/Gameboard.tsx
bh412: Shall add

src/app/services/gameboards.tsx
bh412: Nicki and Lisa's request, I had it as black originally

jsharkey13 (approved):

jsharkey13:
Oops. Maybe I should have got someone to sort the LGTM warning first.

PR #325, opened 2020-10-27
Re-request FastTrack concept state on page change

mlt47:
Progress within the session on concept questions was not being maintained between page changes.
A refresh used to be needed to get the concept progress of other quesitons, now we reload it on page change.


bh412 (approved):
Works as expected with a test board and code changes look concise

PR #326, opened 2020-10-29
Feature/gameboard filter query params

mlt47:


src/app/components/pages/GameboardFilter.tsx
```diff
@@ -11,87 +11,130 @@ import {GameboardViewer} from './Gameboard';
...
     return items.map(item => item.value).join(",");
 }
 
-export const GameboardFilter = withRouter((props: {location: {hash?: string}}) => {
+function processQueryString(query: string): {queryLevels: Item<number>[], querySelections: Item<TAG_ID>[][]} {
+    const {levels, subjects, fields, topics} = queryString.parse(query);
+    const tagHierarchy = tags.getTagHierarchy();
+
+    let levelItems: Item<number>[] = [];
+    if (levels) {
```
bh412, line 44: I think this bit needs looking over, for the issue I raised on slack, otherwise it seems to work well!

bh412 (changes requested):
Works well aside from the issue raised on slack, when rectified it will be good to go

src/app/components/pages/GameboardFilter.tsx
mlt47: Yes, well spotted. Line 47 also needs a length comparison, clearly!

bh412 (approved):
The changes add the functionality we want and fix the levels in url bug

PR #328, opened 2020-11-02
Show full Derivative(y;x;x) in symbol menu.

Morpheu5:
We used to have incomplete `d•/dx`-style derivatives only, ignoring whatever was in the first argument. We now include `d^2 y/dx^2` if the first argument of Derivative is a single letter.


bh412 (approved):
Correctly adds the nth order derivative into the symbol menu when specified in available symbols. Doesn't interfere with first order derivatives

PR #329, opened 2020-11-17
Support disabling indentation for Parsons problems

Morpheu5:
It is sometimes desirable to prevent items from being indented in Parsons problems. This uses the `IsaacParsonsQuestionDTO::disableIndentation` flag to achieve the effect. This will require a toggle control to be added in the editor to turn the flag on and off.

### DIFF
--- a/src/app/components/content/IsaacParsonsQuestion.tsx
+++ b/src/app/components/content/IsaacParsonsQuestion.tsx
@@ -124,7 +124,7 @@ class IsaacParsonsQuestionComponent extends React.Component<IsaacParsonsQuestion
             } else if (e instanceof TouchEvent && e.touches[0]) {
                 cursorX = e.touches[0].clientX;
             }
-            if (isDefined(this.props.doc.disableIndentation) && !this.props.doc.disableIndentation && this.state.initialX && x) {
+            if ((!isDefined(this.props.doc.disableIndentation) || !this.props.doc.disableIndentation) && this.state.initialX && x) {
                 const d = Math.max(0, x - this.state.initialX);
                 const i = Math.min(Math.floor(d/PARSONS_INDENT_STEP), Math.min(this.state.currentMaxIndent, PARSONS_MAX_INDENT));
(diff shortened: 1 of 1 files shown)


src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -315,10 +317,10 @@ class IsaacParsonsQuestionComponent extends React.Component<IsaacParsonsQuestion
...
                                     {this.props.currentAttempt && this.props.currentAttempt.items && this.props.currentAttempt.items.map((item, index) => {
-                                        const canDecreaseIndentation = !_isUndefined(item?.indentation) && item.indentation > 0;
-                                        const canIncreaseIndentation = !_isUndefined(item?.indentation) && index !== 0 && item.indentation <= this.getPreviousItemIndentation(index) && item.indentation < PARSONS_MAX_INDENT;
+                                        const canDecreaseIndentation = isDefined(this.props.doc.disableIndentation) && !this.props.doc.disableIndentation && isDefined(item?.indentation) && item.indentation > 0;
+                                        const canIncreaseIndentation = isDefined(this.props.doc.disableIndentation) && !this.props.doc.disableIndentation && isDefined(item?.indentation) && index !== 0 && item.indentation <= this.getPreviousItemIndentation(index) && item.indentation < PARSONS_MAX_INDENT;
```
jsharkey13, line 323: This condition isn't right?
`canIncreaseIndentation` should be false if `disableIndentation` is defined and true, but true it it's undefined/false.

Should it not be 
`(!isDefined(disableIndentation) || !disableIndentation) && ... `
or equivalently
 `!(isDefined(disableIndentation) && disableIndentation) && ... `
no?

It'll need fixing elsewhere too, see the PR for the list of places you changed it.

jsharkey13 (changes requested):
Doesn't quite work. It _always_ disables indentation, unless you explicitly set it to be false. It's allowed to be undefined, and that should be the same as false!

jsharkey13 (approved):

PR #331, opened 2020-11-20
Remove hyperbolic functions from trig functions menu.

Morpheu5:
For some reason, some of the hyperbolic functions were in the trig functions list, and so would get included in the wrong menu tab as well as in the correct menu tab.

This fixes this mistake.


jsharkey13 (approved):
Although it'd be nice to only commit changes to `package-lock.json` if they're necessary. Since we're on Windows and Mac and Linux, many of these changes are about optional dependencies that don't seem meaningful :/

PR #333, opened 2020-11-30
Simplify and update My Progress page activity graph

jsharkey13:
This replaces #332, and simplifies the code dramatically. The only real difference is that this assumes that the API will provide zero values for months with no activity, and that it will only provide months when activity is meaningful. Currently this is all handled by Postgres, so it seems unlikely that this will go away even if the API changes dramatically.
It also now includes the months that were zero from registration as it used to, which may not actually be desired?

### DIFF
--- a/src/app/components/elements/views/ActivityGraph.tsx
+++ b/src/app/components/elements/views/ActivityGraph.tsx
@@ -49,7 +49,8 @@ export const ActivityGraph = ({answeredQuestionsByDate}: {answeredQuestionsByDat
             zoom: {enabled: true},
             legend: {show: false},
             spline: {interpolation: {type: "monotone-x"}},
-            bindto: "#activityGraph"
+            bindto: "#activityGraph",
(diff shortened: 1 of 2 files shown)


src/app/components/elements/views/ActivityGraph.tsx
```diff
@@ -47,13 +29,13 @@ export const ActivityGraph = ({answeredQuestionsByDate}: {answeredQuestionsByDat
                 colors: {activity: "#ffb53f"},
                 xFormat: "%Y-%m-%d"
             },
-            axis: {x: {type: "timeseries", tick: {fit: false, count: 8}, min: minTime, max: maxTime}},
+            axis: {x: {type: "timeseries", tick: {fit: false, format: '%b %Y', count: Math.min(8, selectedDates.length)}}},
```
bh412, line 32: I think this causes the only issue I've noticed, aside from the colour if we want to change that. If you only have 1 month with activity it is zoomed far out

bh412 (approved):
Altered changes look good, works well

PR #334, opened 2020-12-03
Improvements to group reservations

Morpheu5:
Goes hand-in-hand with [this one](https://github.com/isaacphysics/isaac-api/pull/305)

- Hide the booking form if the user is already booked in.
- Show the correct phrasing on the button if a user is reserved but not booked in.


bh412 (approved):
Correct behaviour on return for students booking

PR #335, opened 2020-12-03
Fix a bug with complete derivatives in menu

Morpheu5:
When the order of a derivative is 1 and there is a compound numerator (i.e., one with a single letter) the letter would not appear on the canvas.
This fixes the issue.


bh412 (approved):
Corrects a missed condition

PR #336, opened 2020-12-10
Improve event handling in InequalityModal

Morpheu5:
This might fix or at least alleviate issues where different symbolic parts of a multi-part question would interfere with each other on the same page.

InequalityModal is still a class-style component so maybe a re-write as a function component might help with the general maintainability of the whole thing.

**Important:** Whoever tests this, you may want to liaise with Lee while testing as he seems to be able to trigger the bug way more reliably than I can.


bh412 (approved):
Can't see any answer altering behaviour

PR #337, opened 2020-12-16
Close Inequality menu on drag-out for small screens

Morpheu5:
This only applies when the user is dragging a potential symbol from the menu onto the canvas, and nothing else. The menu can still be closed manually by touching the active tab.

This should address [this ticket](https://tickets.isaacphysics.org/scp/tickets.php?id=22238).


bh412 (approved):
Closes as expected for small screens

PR #338, opened 2021-01-14
Prevent "Select All" of invalid unreserved users.

Morpheu5:
At the moment we disable the "All" checkbox if there are invalid (unverified) users in the "unbooked" table of the reservations modal. This PR modifies the selection code to exclude users that are invalid from the "All" selection when that checkbox is ticked.


bh412 (approved):
Sensible changes

PR #341, opened 2021-01-22
Add level to Phy question search results

bh412:
Add non-zero levels to Physics question search results


jsharkey13 (changes requested):
This is better than nothing. I think if it is going to go on the line with the topic title, then you need a comma and a space after the topic title.
Maybe use
```
const displayLevel = SITE_SUBJECT === SITE.PHY && level !== undefined && level !== "0";
```
and have
```
{displayTopicTitle && displayLevel && <span>, </span>}
```
before the level span and remove the padding on it?

Otherwise I don't think this breaks the other two places where these elements are used!

jsharkey13 (approved):
🎉

PR #342, opened 2021-01-22
Stop location being displayed if tagged virtual

bh412:


src/app/components/pages/EventDetails.tsx
```diff
@@ -101,6 +101,8 @@ export const EventDetails = ({match: {params: {eventId}}, location: {pathname}}:
             "Book now" :
             event.isWithinBookingDeadline ? "Apply" : "Apply - deadline past";
```
jsharkey13, line 102: Nothing to do with your PR, @bh412, but am I right in thinking that "Deadline past" is grammatically incorrect and it should be "Deadline passed", or am I missing context?

The PR looks fine, so I will merge it either way!

jsharkey13 (approved):

PR #343, opened 2021-01-26
Update teacher view modal to show all currently reserved

bh412:
- Shows all your currently booked users in the modal before having to select a group
- Requires corresponding API changes


Morpheu5 (approved):
Code looks OK and performs as intended.

PR #344, opened 2021-01-27
Add figure numbering to page fragments

mlt47:
The main problem with supporting figure numbering is that you can have multiple page fragments on a single page and the figure numbers should not restart from 1 for each fragment.
We now create a figure numbering "scope" for each page - unfortunately implemented by piggy-backing on the already overloaded TrackedRoute component.
I believe that it would be possible for a page with two page fragments to load out of order so that Figure 2 appears before Figure 1 on the final rendered page but that did not happen in my constructed page. Even then, each reference would still at least be unique.
No examples exist in the wild so I constructed a page with two page fragments each containing a unique Figure and figure reference (details to reproduce in last paragraph).
It certainly isn't perfect but it does the right thing in more cases than it used to.

My temporary unpublished page fragment can be found at: `/master/content/general_pages/support/unpublished_temp_support_for_teachers_general.json`
I suggest adding a page fragment pointing to its ID in the support page component then visiting `/support/teacher/general` and inspecting the figure numbering for the section titled: "How do I connect to my students?" in both fragments. They should contain a fig 1 and fig 2 because I gave them unique figure numbers. Please delete the temporary editor file after the PR is merged, if we eventually add a real world case it can be added to our regression tests.

### DIFF
--- a/src/app/components/elements/PageFragment.tsx
+++ b/src/app/components/elements/PageFragment.tsx
@@ -4,6 +4,7 @@ import {ShowLoading} from "../handlers/ShowLoading";
 import {IsaacContent} from "../content/IsaacContent";
 import {useDispatch, useSelector} from "react-redux";
 import {fetchFragment} from "../../state/actions";
+import {WithFigureNumbering} from "./WithFigureNumbering";
 
 
 interface PageFragmentComponentProps {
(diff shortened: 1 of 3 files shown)


bh412 (approved):
Works well with the test page and retains numbering elsewhere. Maybe worth trying to bring things out of TrackedRoute in future, in general, not just this PR

PR #345, opened 2021-01-29
Ensure figure references are all bold

bh412:


mlt47 (approved):
This is straightforward and fixes the highlighted error case. I added one other place where we create the `Figure n` text.

PR #348, opened 2021-02-03
Improvement/concept conditional return

bh412:
- Uses persistence to store Concept page source key to return to concept list or question
- Removes it in IsaacApp if not at correct URL


src/app/components/navigation/IsaacApp.tsx
```diff
@@ -109,6 +109,12 @@ export const IsaacApp = () => {
         }
     }, [dispatch, user, notifications]);
 
+    useEffect(() => {
+        if (!(window.location.href.includes("/questions/") || window.location.href.includes("/concept"))) {
+            persistence.remove(KEY.CONCEPT_PAGE_SOURCE);
+        }
+    });
```
mlt47, line 116: I think I would rather this be a middleware acting upon ROUTER_PAGE_CHANGE actions rather than  (check notificationCheckerMiddleware service for an example). As there are no dependencies for this useEffect it is run on any re-render.

src/app/services/localStorage.ts
```diff
@@ -5,7 +5,8 @@ export enum KEY {
     FIRST_LOGIN = "firstLogin",
     REQUIRED_MODAL_SHOWN_TIME = "requiredModalShownTime",
     LAST_NOTIFICATION_TIME = "lastNotificationTime",
-    ANONYMISE_USERS = "anonymiseUsers"
+    ANONYMISE_USERS = "anonymiseUsers",
+    CONCEPT_PAGE_SOURCE = "conceptPageSource"
```
mlt47, line 9: Perhaps concept page referrer is more descriptive of this data

src/app/components/navigation/IsaacApp.tsx
mlt47: This middleware can then also do the saving to persistence

src/app/components/pages/Concept.tsx
```diff
@@ -31,6 +33,17 @@ export const Concept = withRouter(({match: {params}, conceptIdOverride}: Concept
     useEffect(() => {dispatch(fetchDoc(DOCUMENT_TYPE.CONCEPT, conceptId));}, [conceptId]);
     const doc = useSelector((state: AppState) => state?.doc || null);
     const navigation = useNavigation(doc);
+    const conceptPageSource = persistence.load(KEY.CONCEPT_PAGE_SOURCE);
+
+    const [canGoBack, setCanGoBack] = useState(conceptPageSource?.includes("/questions/") || conceptPageSource?.endsWith("/concepts"));
```
mlt47, line 38: This doesn't need to be react state because it is never re-set

src/app/components/pages/Concept.tsx
```diff
@@ -31,6 +33,17 @@ export const Concept = withRouter(({match: {params}, conceptIdOverride}: Concept
     useEffect(() => {dispatch(fetchDoc(DOCUMENT_TYPE.CONCEPT, conceptId));}, [conceptId]);
     const doc = useSelector((state: AppState) => state?.doc || null);
     const navigation = useNavigation(doc);
+    const conceptPageSource = persistence.load(KEY.CONCEPT_PAGE_SOURCE);
```
mlt47, line 36: On the old site this would get cleared after first read. If you want the value to persist across this component's re-renders you can do something like this `const conceptPageReferrer = useRef(persistence.load(KEY.CONCEPT_PAGE_SOURCE));` and then clean CONCEPT_PAGE_SOURCE

mlt47 (changes requested):
I've suggested a few changes as I think it can be simplified and pulled together (perhaps by using middleware and redux).
Do you know if there is a case that requires this to be stored in local storage rather than something less persistent, because currently I would be tempted to clear it onbeforeunload which would sort of defeat the point of having it stored there?

mlt47:
We have decided not to add this feature as it had too many edge-cases and was not adding functionality that was not already supported by the browser's back button.

PR #349, opened 2021-02-05
Improvement/user my progress

bh412:
- Separate progress actions out to getMyProgress and getUserProgress.
- Separate redux states for myProgress and userProgress.
- Stops updating userSnapshot on webSocket message for userProgress reducer, stopping viewed user's streak updating to viewer's.
- Hides daily streak if 0.
- Leaves achievements alone on progress page, would only potentially affect admins viewing teachers progress, would see their own achievements.
- Still shows the teacher/admin/progress viewer's streak in the header as a separate thing, could be confusing, hopefully it makes sense to them though.


src/app/components/pages/MyProgress.tsx
```diff
@@ -70,7 +71,9 @@ export const MyProgress = withRouter(({user, match: {params: {userIdOfInterest}}
         return <Unauthorised />
     }
 
-    const userName = `${userProgress?.userDetails?.givenName || ""}${userProgress?.userDetails?.givenName ? " " : ""}${userProgress?.userDetails?.familyName || ""}`;
+    const viewedProgress = (!viewingOwnData && isTeacher(user)) ? userProgress : myProgress;
```
mlt47, line 74: I think I prefer `progress` over `viewedProgress`

src/app/components/pages/MyProgress.tsx
```diff
@@ -50,16 +50,17 @@ export const MyProgress = withRouter(({user, match: {params: {userIdOfInterest}}
     const viewingOwnData = userIdOfInterest === undefined || (user.loggedIn && parseInt(userIdOfInterest) === user.id);
 
     const dispatch = useDispatch();
+    const myProgress = useSelector((state: AppState) => state?.myProgress);
     const userProgress = useSelector((state: AppState) => state?.userProgress);
-    const achievements = useSelector((state: AppState) => state?.userProgress?.userSnapshot?.achievementsRecord);
+    const achievements = useSelector((state: AppState) => state?.myProgress?.userSnapshot?.achievementsRecord);
     const answeredQuestionsByDate = useSelector((state: AppState) => state?.answeredQuestionsByDate);
```
mlt47, line 56: This has the same problem as myProgress/userProgress they should be separate redux states to avoid accidental confusion between the states. This is less likely to happen here but seeing as we've separated one, it would be confusing to not separate the other.

mlt47 (changes requested):
Seems to work correctly but there is a small change request for consistency

src/app/components/pages/MyProgress.tsx
mlt47: Sorry, this comment was not clear as to what line I was referring to! I meant the answeredQuestionsByDate should be split as that shows either your one or the student's one.

achievements actually don't need to be split this way because achievements, by design, are only shown if you are viewing your own progress (and we shouldn't request it if we're not showing it).

bh412: My bad! Now altered to separate out [my/user]AnsweredQuestionsByDate mirroring [my/user]Progress

mlt47 (approved):
Great, thanks for those changes and sorry for the confusion!

PR #353, opened 2021-02-15
Toggle button for question reminder in Inequality

Morpheu5:
This adds a way of hiding the question reminder in Inequality which sometimes can block out the actual editor.

Fair warning: `pointer-events` is an evil CSS property that can interfere with browser mouse events.


src/scss/common/inequality-modal.scss
```diff
@@ -16,16 +16,30 @@
 
     .question-reminder {
         position: absolute;
-        bottom: 140px;
+        bottom: 30px;
         left: 250px;
         right: 250px;
-        text-align: center;
+        text-align: justify;
```
bh412, line 22: This PR seems to move the text to be left aligned, maths stays central though

src/scss/common/inequality-modal.scss
```diff
@@ -16,16 +16,30 @@
...
         border-radius: 20px;
         background-color: rgba(0, 0, 0, 0.6);
         color: white;
-        padding: 1.25rem 1px 0;
+        padding: 1.25rem;
         pointer-events: none;
         z-index: 255 !important;
+
+        .reminder-toggle {
```
bh412, line 30: Slightly worried that this doesn't achieve what it's trying to address. The box is still visible on small screens

bh412 (changes requested):
On mobile screens the question is even bigger now with the button within it, and when it's 'hidden' the button and full width grey box is still there. Would it be possible to move the show/hide button out to a hexagon, like the 'centre' button and fully hide the grey question reminder box when 'hidden'. 
I'm also worried about the comment on the PR, you say it sometimes hides the full equation editor?

src/scss/common/inequality-modal.scss
Morpheu5: Yeah, the alternative is to have an extra hexagonal button. I might do that if I can't move the grey box enough out of the way without removing it entirely.

src/scss/common/inequality-modal.scss
Morpheu5: The question linked in the card is horrendously long and it looks much better when justified. Math is always centred when on a line of its own.

bh412: Ok - felt I should flag and make sure it's a conscious decision as it changes behaviour from current

src/scss/common/inequality-modal.scss
bh412: Not ideal but I'm not sure how to get around the grey box issue, having just the current button with no grey box around would also look very odd

Morpheu5: I fixed this by adding an extra button.

bh412 (approved):
Works as intended now and hides question box

PR #355, opened 2021-02-16
Allow the compose scripts to manage both physics and CS containers

jsharkey13:
This also improves the validation of inputs initially and before they get passed into docker-compose, which seemed sensible.

It also removes the "special-case" nature of the test database initialisation; since the Docker Postgres image will not use the init
scripts if the data directory is blank, it ought to be safe to include the test scheme for all non-prod databases. Doing this dramatically simplifies the compose script.

This also restricts the types of docker-compose commands that can be run, to prevent passing unexpected things into it. If any commands have been missed, they can easily be added and a helpful error is printed.

### DIFF
--- a/compose
+++ b/compose
@@ -1,76 +1,74 @@
 #!/bin/bash
 
-if [ $# -eq 0 ]
-  then
-    echo "Usage: compose [dev|staging|test] VERSION"
-    exit 1
-fi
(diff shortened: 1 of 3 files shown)


bh412 (approved):
Code changes themselves look sensible and all named variables are controlled for phy/cs. Runs as expected with `config` locally - not tested on server

PR #360, opened 2021-02-22
Force physics filter to specify problem solving questions

mlt47:
This PR has an accompanying backend PR.

This will allow the content team to safely tag Maths questions as both `book` and `problem_solving` questions.
In the future, the filter will surface a question category selection (for Physics at least).


bh412 (approved):
Works as intended and overrides any questionCategories a user may try to provide

PR #362, opened 2021-02-22
Soft limit search character length to 255

bh412:
Adds a 255 character limit to the search boxes, before a user hits the 1000 character backend bad request


jsharkey13 (approved):

jsharkey13:
I tested this too, good to merge if/when tests pass!

PR #364, opened 2021-02-23
Quick help button for Inequality

Morpheu5:
This adds a quick help button inside the equation editor. The button fires up a modal window that shows a page fragment depending on the current editor mode. The modes are `maths`, `chemistry`, and `logic`. These are appended to the string `eqn_editor_help_modal_` to form the page fragment's ID that will be pulled from ES. If the ID can't be found, an error message is displayed.


bh412 (approved):
Works nicely with the template help text, just waiting on the content teams to write their various help docs

PR #366, opened 2021-02-25
Quiz: set quiz

robinmessage:


src/app/components/elements/modals/QuizSettingModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+                        setFeedbackMode(item.value);
+                    }
+                    addValidated('feedbackMode');
+                }}
+                onBlur={() => addValidated('feedbackMode')}
+                options={feedbackOptionsList}
+                styles={{
+                    control: (styles) => ({...styles, ...(feedbackModeInvalid ? {borderColor: '#dc3545'} : {})}),
+                }}
+            />
```
mlt47, line 104: For selects in modals we need to add these properties to make sure it doesn't get clipped by the modal:
```
menuPortalTarget={document.body}
styles={{menuPortal: base => ({...base, zIndex: 9999})}}
```
I believe the QuestionSearchModal is the only place where we've had to do this in the past. If there are going to be a few I guess we could create a <SelectInModal ... /> for reuse which just sets those props?

src/app/state/selectors.tsx
```diff
@@ -130,9 +132,30 @@ export const selectors = {
...
+    console.log(groupCache);
+    return quizAssignments.map(assignment => {
+        console.log(assignment);
+        const groupName = groupCache[assignment.groupId as number]?.groupName;
+        return {
+            ...assignment,
+            groupName,
+        } as AppQuizAssignment;
+    });
+}
```
mlt47, line 157: Sounds like this will be useful more generally too which is great. (Also, reminder to remove the console.log(...)! :) )

mlt47 (commented):
This looks good so far. Do you have strong feelings about how you'd like to hide these features if we're going to follow a branch merging and release pattern similar to the API? The branch is still set to WIP according to GitHub, so I'll review again at a later point.

src/app/components/elements/modals/QuizSettingModal.tsx
robinmessage: I don't have any other modals planned but I'll pull that out if I do.

src/app/state/selectors.tsx
robinmessage: Good point, oops! Yes, I guess it could come in handy elsewhere although you do have to be aware of needing to fetch the active groups.

robinmessage:
I don't have a strong feeling about how to merge this all.
We could make all the menus and/or routes show up only for admins, but then it is fiddly to test. If there is any kind of feature flag or dev/prod flag, I could use that?

It could live on a long-running branch, but I don't know when you will want to release it, so that might not make sense.

So, feature flag or dev/prod flag?

mlt47:
> 
> 
> I don't have a strong feeling about how to merge this all.
> We could make all the menus and/or routes show up only for admins, but then it is fiddly to test. If there is any kind of feature flag or dev/prod flag, I could use that?
> 
> It could live on a long-running branch, but I don't know when you will want to release it, so that might not make sense.
> 
> So, feature flag or dev/prod flag?

Cool, yeah I was thinking admin initially but a flag is less likely to go wrong.
I'm going to suggest a simple `const QUIZZES_FEATURE = false` in constants, you can have it to true for your development.
I'll add a quick automated regression test making sure "Quiz" does not appear in the teacher nav bar to make sure it doesn't make it into a release!

mlt47 (approved):
This is good. Will merge behind static constant feature flag when tests pass. Will change the style a little before releasing to public :)

robinmessage:
Thanks Meurig, I would have put the flag in but you've done it for me 👍 and the rename makes a lot of sense.

PR #370, opened 2021-03-08
Improvement/video end time

bh412:
- Requires content teams to add (?/&)end=[number in seconds] for a video to automatically stop at that point


jsharkey13 (approved):
Looks good! I wonder though if allowing `end=...` might cause someone to inadvertantly use `start=...` as well, which won't work.

Perhaps we need to change it to:
```
const possibleStartTime = /[?&](t|start)=([0-9]+)/.exec(src);
```
and to use the right Regex group later:
```
const optionalStart = possibleStartTime ? `&start=${possibleStartTime[2]}` : "";
```
just to be safe?

jsharkey13:
:tada: 🙌

PR #373, opened 2021-03-17
Allow video thumbnail scaling

bh412:


src/scss/common/elements.scss
```diff
@@ -16,6 +16,24 @@
   text-decoration: none;
 }
 
+.content-video-container {
+  position: relative;
+  width: 80%;
+  margin-left: 10%;
+  margin-right: 10%;
```
mlt47, line 23: This affects the aspect ratio, is there a reason it is not width 100%?

src/app/components/content/IsaacVideo.tsx
```diff
@@ -103,7 +103,7 @@ export function IsaacVideo(props: IsaacVideoProps) {
     return <div>
         <div className="no-print content-value text-center">
             { embedSrc ?
-                <iframe ref={videoRef} className="mw-100" title={altText} width="614" height="390" src={embedSrc} frameBorder="0" allowFullScreen/>
+                <div className="content-video-container"><iframe ref={videoRef} className="mw-100" title={altText} width="614" height="390" src={embedSrc} frameBorder="0" allowFullScreen/></div>
```
mlt47, line 106: We need to remove width and height attributes to allow the CSS to maintain the aspect ratio for us.

src/scss/common/elements.scss
bh412: I thought it looked better for side by side videos, but removing the height/width from the video iframe corrects this

mlt47: I see, that makes sense. Cool 🙂

mlt47 (approved):

PR #374, opened 2021-03-22
Move phy sig fig error feedback to front end

bh412:
Corresponding API PR: https://github.com/isaacphysics/isaac-api/pull/338


jsharkey13 (approved):

jsharkey13:
This is actually safe to merge without the API change, since the explanation is already tagged correctly. It just won't fix CS until the [API PR](https://github.com/isaacphysics/isaac-api/pull/338) is done.

PR #375, opened 2021-03-24
Restore functionality to Inequality number menu on iOS 14

Morpheu5:
Apparently now we have to use `onTouchEnd` to intercept click events on iOS.


jsharkey13 (approved):
Works fine on iOS on an iPhone and iPad, I can drag numbers as normal and build up longer numbers before dragging them too!

PR #379, opened 2021-03-26
Update the downtime warning banner for the planned Easter maintenance

jsharkey13:
Might be worth someone checking those `new Date(...)` calls as well as me, but they are supposed to be the Wednesday before and going away at 9am on the day.


bh412 (approved):
The dates are correct with BST, just checking we don't need `&nbsp;` after the `</a>` within the span?

jsharkey13:
No need for a no-break space since there's no newline to cause whitespace trimming, but good thought because that has bitten us before now!

PR #381, opened 2021-03-26
Quiz: final tweaks

robinmessage:


mlt47 (approved):
I'm happy with these, and I don't think the cleaning up of the quiz attempt is too bad. At least it is in the useEffect and therefore close to where the loading logic is.

PR #382, opened 2021-03-31
Feature/code block

bh412:
I tested using http://localhost:8003/pages/code_block_test (https://editor.isaaccomputerscience.org/#!/edit/master/content/general_pages/code_block_test.json)

Doing the registerLanguage in index-cs seems a little odd but also maybe the most sensible place, it needs to be done once when the site is opened and it's only nescessary for cs.

This adds the codeSnippet content type to the front end, currently for Python, PHP, JS and C#.

Uses the corresponding API and editor changes


src/app/components/content/IsaacCodeSnippet.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import hljs from 'highlight.js';
+
+interface IsaacCodeProps {
+    doc: CodeSnippetDTO;
+}
+
+export const IsaacCodeSnippet = ({doc}: IsaacCodeProps) => {
+    useEffect(() => {
+        hljs.highlightAll();
```
mlt47, line 13: This will highlight all code and pre elements on the page (code snippet or not) if it has the appropriate class name. Is this what we want? It is slightly inefficient and might cause problems when we have many code blocks on a page (it looks for all code and pre elements on the page x number of code snippets on every render).

Maybe https://highlightjs.readthedocs.io/en/latest/api.html#highlightblock-block is more appropriate if we pass it a ref (using useRef()) to the element to highlight?

mlt47 (approved):
This works and looks quite neat. Please see suggestions

src/app/components/content/IsaacCodeSnippet.tsx
```diff
@@ -0,0 +1,32 @@
...
+                <pre>
+                    <code className={doc.disableHighlighting ? 'plaintext' : doc.language}>
+                        {doc.code}
+                    </code>
+                </pre>
+            </Col>
+        </Row>
+        {doc.url && <Row>
+            <Col className="text-center">
+                <a href={doc.url} target="_blank" rel="noopener noreferrer">View on GitHub</a>
```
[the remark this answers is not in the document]
bh412: Yeah, I made it clear they would need to edit both github and here and they were happy with that (they prefer it being independent to github). Before I did this task James had a look at pulling in the github files (as gists(?) I believe) and found it didn't work. Bit again, the Cs content team wanted it to not use github directly

src/app/components/content/IsaacCodeSnippet.tsx
bh412: I'll have a look at this!

PR #386, opened 2021-04-09
Only import necessary languages from highlight.js to reduce bundle size (type issues?)

bh412:
Reduces highlight.js bundle size to ~72KB
Known type issue, not sure how best to proceed (ts-ignore or otherwise): https://github.com/highlightjs/highlight.js/issues/2682


jsharkey13:
There looks to be a further fix in a commit not yet in a release: https://github.com/highlightjs/highlight.js/commit/779be2b9cf1fcedeb4623cf446e18f0d481e1a4d
I suggest we hold off for that, to try and avoid the need for ts-ignore.

But:
 - we do not want the major changes to `package-lock.json` this PR introduces (it breaks a lot more than `highlight.js`, actually).
 - we do not need the separate `@types/highlight.js` dependency, since it provides nothing.

bh412:
Replaced with https://github.com/isaacphysics/isaac-react-app/pull/399 due to package-lock issues

PR #388, opened 2021-04-30
Stops pre-sorting progress when sorting by each question

bh412:


Morpheu5 (commented):
LGTM. The only way to make this better would be to extract the generation of sorting attributes to a function.

PR #396, opened 2021-05-14
Only show topics on glossary and board builder searches

bh412:


jsharkey13 (approved):
Still needs actually testing but the changes look sensible!

jsharkey13:
Looks good on the site too 🎉

PR #398, opened 2021-05-27
Implement UI for additional managers to remove themselves from groups

bh412:
- When trying to delete a group as one of it's additional managers, instead opens the 'additional manager self removal options'.
- When a group you are an additional manager of is selected, a 'remove self as additional manager' button is shown instead of the 'archive group'.
- When a group you are an additional manager of is selected, instead of showing the 'edit additional managers' button, it is altered to 'view additional managers'.
- When a group you are an additional manager of is selected, the group name is shown in a disabled input box, without the update button, to make it more obvious you can't edit the name of a group you don't own.

### DIFF
--- a/src/app/components/elements/modals/GroupsModalCreators.tsx
+++ b/src/app/components/elements/modals/GroupsModalCreators.tsx
@@ -25,6 +25,48 @@ interface CurrentGroupInviteModalProps {
     firstTime: boolean;
 }
 
+interface AdditionalManagerRemovalModalProps {
+    groupToModify: AppGroup;
+    user: RegisteredUserDTO;
+    showArchived: boolean;
+}
+
+const AdditionalManagerRemovalModalBody = ({groupToModify}: AdditionalManagerRemovalModalProps) => {
+    return <React.Fragment>
+        <p>You are about to remove yourself as a manager from &apos;{groupToModify.groupName}&apos;.  The group will not appear any more on your
+            Assignment Progress page or on the Manage Groups page.  You will still have student connections with the
+            students who agreed to share data with you.  The group owner will <strong>not</strong> be notified.</p>
+    </React.Fragment>
+};
+
+export const additionalManagerRemovalModal = ({groupToModify, user, showArchived}: AdditionalManagerRemovalModalProps) => {
+    return {
+        closeAction: () => {store.dispatch(closeActiveModal())},
+        title: "Additional manager self removal",
+        body: <AdditionalManagerRemovalModalBody groupToModify={groupToModify} user={user} showArchived={showArchived}/>,
+        buttons: [
+            <RS.Row key={0}>
+                <RS.Col>
+                    <RS.Button block key={2} color="secondary" onClick={() => {
+                        store.dispatch(closeActiveModal());
+                    }}>
+                        Cancel
+                    </RS.Button>
+                </RS.Col>
+                <RS.Col>
+                    <RS.Button block key={1} color="secondary" onClick={() => {
+                        store.dispatch(deleteGroupManager(groupToModify, user, showArchived));
+                        store.dispatch(closeActiveModal());
+                        store.dispatch(selectGroup(null));
+                    }}>
+                        Confirm
+                    </RS.Button>
+                </RS.Col>
+            </RS.Row>
+        ]
+    }
+};
+
 const CurrentGroupInviteModal = ({firstTime}: CurrentGroupInviteModalProps) => {
     const group = useSelector(selectors.groups.current);
     return group && <React.Fragment>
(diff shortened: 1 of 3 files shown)


src/app/components/elements/modals/GroupsModalCreators.tsx
```diff
@@ -25,6 +25,48 @@ interface CurrentGroupInviteModalProps {
...
 
+interface AdditionalManagerRemovalModalProps {
+    groupToModify: AppGroup;
+    user: RegisteredUserDTO;
+    showArchived: boolean;
+}
+
+const AdditionalManagerRemovalModalBody = ({groupToModify}: AdditionalManagerRemovalModalProps) => {
+    return <React.Fragment>
+        <p>You are about to remove yourself as a manager from &apos;{groupToModify.groupName}&apos;.  The group will not appear any more on your
```
jsharkey13, line 36: "... will no longer appear ..." parses better?

jsharkey13: Has any of this been past CS copy-edit? I assume not?

src/app/components/elements/modals/GroupsModalCreators.tsx
```diff
@@ -25,6 +25,48 @@ interface CurrentGroupInviteModalProps {
...
+            Assignment Progress page or on the Manage Groups page.  You will still have student connections with the
+            students who agreed to share data with you.  The group owner will <strong>not</strong> be notified.</p>
+    </React.Fragment>
+};
+
+export const additionalManagerRemovalModal = ({groupToModify, user, showArchived}: AdditionalManagerRemovalModalProps) => {
+    return {
+        closeAction: () => {store.dispatch(closeActiveModal())},
+        title: "Additional manager self removal",
```
jsharkey13, line 45: This title could do with review too, when it happens.

src/app/components/pages/Groups.tsx
```diff
@@ -271,9 +276,12 @@ const GroupEditor = ({group, selectGroup, updateGroup, createNewGroup, groupName
             {group && <React.Fragment>
                 <Row>
                     <Col>
-                        <Button block color="tertiary" onClick={toggleArchived}>
+                        {isUserGroupOwner ? <Button block color="tertiary" onClick={toggleArchived}>
                             {group.archived ? "Unarchive this group" : "Archive this group"}
-                        </Button>
+                        </Button> :
+                        <Button block color="tertiary" onClick={() => user?.loggedIn && showAdditionalManagerSelfRemovalModal(group, user, showArchived)}>
+                            Remove self as additional manager
```
jsharkey13, line 283: This text is particularly unclear and clunky for a button, too.

jsharkey13: "Leave group" or "Stop managing group" may be alternatives, but I suggest CS copy-edit needs the final say.

jsharkey13 (commented):
I think the wording needs some work still, with the copy-edit team.

I also don't like the button placement location; it's enormous and duplicates the functionality of the "View group managers" button above, which doesn't actually allow you to do the thing that should logically be on that modal already. Having two very similar sounding buttons so close (and one much larger than the other) isn't nice!

jsharkey13:
(I recognise this is exactly what physics asked you to do. I'm just saying that since it affects CS too, it needs to be cleared by them, and would benefit from some thought about button placement!)

src/app/components/elements/modals/GroupsModalCreators.tsx
bh412: This was physics provided, I'll send it to Diane - unless someone else would be better?

bh412:
Good point - yep, I'll send it to the CS team, this was all just provided by the Physics content team

src/app/state/actions.tsx
```diff
@@ -1424,11 +1428,14 @@ export const addGroupManager = (group: AppGroup, managerEmail: string) => async
...
 
-export const deleteGroupManager = (group: AppGroup, manager: UserSummaryWithEmailAddressDTO) => async (dispatch: Dispatch<Action>) => {
+export const deleteGroupManager = (group: AppGroup, manager: UserSummaryWithEmailAddressDTO, showArchived?: boolean) => async (dispatch: Dispatch<Action>) => {
     dispatch({type: ACTION_TYPE.GROUPS_MANAGER_DELETE_REQUEST, group, manager});
     try {
         await api.groups.deleteManager(group, manager);
         dispatch({type: ACTION_TYPE.GROUPS_MANAGER_DELETE_RESPONSE_SUCCESS, group, manager});
+        if (typeof showArchived === "boolean") {
+            dispatch(loadGroups(showArchived) as any);
+        }
```
mlt47, line 1438: We have an isDefined(...) which is a type guard that will assure that showArchived is not undefined and so must be boolean. Better to use that than compare the string value of typeof

mlt47 (approved):
The "react-ness" of this is fine but I have made one minor recommendation

PR #399, opened 2021-06-02
Upgrade highlightjs to 11.0.0 and reduce bundle size

bh412:
highlightBlock has been deprecated and replaced with highlightElement


jsharkey13 (approved):
This looks great, and I have verified the bundle size is reduced from 7.63MB uncompressed to 6.41MB (or 1.17MB to 946kb compressed). I have not checked an actual page, but Ben says it still works as expected 🎉

PR #403, opened 2021-06-22
Feature/add latex component

mlt47:


jsharkey13 (approved):
This is so much cleaner, and I like it! I did a quick test, and it seems titles still work okay so I think we should use this until someone complains and then fix that!

PR #404, opened 2021-06-28
Add temporary front end quiz error modification

bh412:
As this functionality is intended to change in future this adds a temporary front end error modification if a quiz a user is trying to attempt has already been assigned to them by one or more teachers.

The modification provides a link to the `/quizzes` page so they can find their quiz assignment(s)


mlt47 (approved):
This is fine as a temporary fix. Be sure to create a Trello card for a more robust solution to be implemented for Quizzes 2.0

PR #409, opened 2021-07-01
Updates reactstrap to v8.9.0 and react-popper as a dependency. Some a…

bh412:
…lterations needed


bh412:
Not suitable to merge quite yet

PR #411, opened 2021-07-06
Feature/page help modal

bh412:
Enables modals consisting of page fragments to be included instead of a help tooltip. Limited to Physics for now via `showModal` in `PageTitle.tsx`.


src/app/components/elements/PageTitle.tsx
```diff
@@ -28,12 +32,32 @@ export const PageTitle = ({currentPageTitle, subTitle, help, className, level}:
...
 
+    interface HelpModalProps {
+        modalId: string;
+    }
+
+    const HelpModal = (props: HelpModalProps) => {
+        return <>
+            <PageFragment fragmentId={props.modalId}/>
+        </>
+    };
```
chrisjpurdy, line 43: What do the empty tags signify here? Is it some kind of configurable default tag?

bh412: It's the short syntax for a react fragment: https://reactjs.org/docs/fragments.html#short-syntax

mlt47: As the React.Fragment has only one child we do not gain anything from wrapping this PageFragment.

mlt47: Have you tried to see  what happens when there is not a page fragment with that ID? It would be good to be fallback on showing the data in props.help maybe?

Morpheu5 (approved):

src/app/components/elements/PageTitle.tsx
Morpheu5: If props.help is meant to be shown as a tooltip, then maybe we should fall back to a tooltip (though I can see how that might be a bit complex).

src/app/components/elements/PageFragment.tsx
```diff
@@ -22,13 +23,15 @@ export const PageFragment = ({fragmentId, renderFragmentNotFound}: PageFragmentC
     }, [dispatch, fragmentId]);
 
     const notFoundComponent = <div>
+        {isDefined(renderFragmentNotFound) && typeof renderFragmentNotFound !== "boolean" ?
+            <div className="my-4">{renderFragmentNotFound}</div> : <>
```
mlt47, line 27: This feels a little icky to reason about. How about receiving a notFound prop which can be a component or undefined but defaults to either <React.Fragment /> or the "content not found" block depending on which we choose most frequently when we we actually use PageFragments?

(If I remember correctly a default value can be specified in the function args like in Python which is probably neatest)

src/app/components/pages/books/QuantumMechanicsPrimer.tsx
```diff
@@ -5,14 +5,10 @@ import {PageFragment} from "../../elements/PageFragment";
...
-    const pageHelp = <span>
-        The Isaac Physics Quantum Mechanics Primer
-    </span>;
-
     return <Container className="physics">
         <Col>
             <div className="book-intro">
-                <TitleAndBreadcrumb className="mb-5" currentPageTitle="Quantum Mechanics Primer" help={pageHelp} />
+                <TitleAndBreadcrumb className="mb-5" currentPageTitle="Quantum Mechanics Primer" />
                 <PageFragment fragmentId="quantum_mechanics_primer_intro"/>
```
mlt47, line 12: Are we removing the pageHelp because it is not very useful?

src/app/components/elements/PageFragment.tsx
mlt47: Say we default to <React.Fragment /> we can create a constant in this file defining the content not found component so that the calling components look like `<PageFragment fragmentId="something" />` or `<PageFragment fragmentId="otherthing" ifNotFound={contentNotFoundComponent} />`

src/app/components/pages/books/QuantumMechanicsPrimer.tsx
bh412: Yep! The removals are Phy only pages that Phy requested

src/app/components/elements/PageFragment.tsx
```diff
@@ -9,7 +9,7 @@ import {WithFigureNumbering} from "./WithFigureNumbering";
 
 interface PageFragmentComponentProps {
     fragmentId: string;
-    renderFragmentNotFound?: boolean;
+    renderFragmentNotFound?: ReactElement;
```
mlt47, line 12: This change makes (the rather icky) line 18 always set showFragment to true. I don't actually think this will cause a problem because wherever you had renderFragmentNotFound={false} you've swapped false for a "no op" of sorts. Worth double-checking and then removing the showFragment.

src/app/components/pages/AllTopics.tsx
```diff
@@ -84,7 +86,7 @@ export const AllTopics = () => {
                             All: <>
                                 <Row>
                                     <Col lg={{size: 8, offset: 2}} className="bg-light-grey pt-md-4">
-                                        <PageFragment fragmentId={`${stage}_all_topics`} renderFragmentNotFound={false} />
+                                        <PageFragment fragmentId={`${stage}_all_topics`} renderFragmentNotFound={allTopicsFallback} />
```
mlt47, line 89: I'd slightly prefer `renderFragmentNotFound={<React.Fragment />}`.
I'd also consider renaming `renderFragmentNotFound` to `ifNotFound`.

PR #414, opened 2021-07-09
Add CSS-based line numbering to code blocks

jsharkey13:
This might be fragile, but so far seems to work surprisingly well! As noted it is based heavily on the PrismJS line numbering, and only works for our pre elements that are not line-wrapped!


src/index-cs.tsx
```diff
@@ -7,7 +7,7 @@ import {Provider} from "react-redux";
 import {store} from "./app/state/store";
 import {IsaacApp} from './app/components/navigation/IsaacApp';
 import {printAsciiArtLogoToConsole} from "./app/services/easterEggs/csLogoAsciiArt";
-import * as highlightJsConfig from "./app/services/highlightJsConfig";
+import * as highlightJsConfig from "./app/services/highlightJs";
```
jsharkey13, line 10: I think this probably needs renaming, now that I renamed the file :/

src/scss/common/code.scss
```diff
@@ -0,0 +1,48 @@
+// Adapt the PrismJS code for line numbering:
+// https://github.com/PrismJS/prism/blob/v1.24.1/plugins/line-numbers/prism-line-numbers.css
+
+$code-indent: 3.8em;
```
jsharkey13, line 4: We should have a note here that it's only good for up to 999 lines

mlt47 (approved):
This looks good and appropriate to me

jsharkey13:
I do agree with @mlt47's comment in a meeting that this is not a particularly React-y way of doing it, and the more I think about it the more I think it would be very easy to accomplish in a very React-y way now that I understand how it works and have simplified it down.
Perhaps I can talk to @chrisjpurdy about a refactoring-into-React task he could try at some point!

bh412 (approved):
Will leave merging until looking at 'React-i-fying' has been done (or looked at), otherwise works nicely!

PR #417, opened 2021-07-14
Add basic syntax highlighting for the Isaac CS pseudocode syntax

jsharkey13:
This was fun, although I do wonder: if it's well-defined enough for syntax highlighting, is it not at that point just a new language to learn?

This PR extends partly from #414, so ideally review and merge that first (at which point this may have conflicts to resolve).

Still some type issues with highlight.js to be resolved: highlightjs/highlight.js#3272 and highlightjs/highlight.js#3273, but I am using `@ts-ignore` on the definition method until they are resolved.


Morpheu5 (approved):
Looks sensible enough to me.

PR #419, opened 2021-07-22
Add link to events safeguarding page in navbar

chrisjpurdy:
I don't think anything else needs changing here, and it works seemingly as intended locally


jsharkey13 (approved):
I knew this would be easy 🎉

PR #420, opened 2021-07-26
Add displayUnit support for numeric questions

chrisjpurdy:
In line with James' API pull request, the existence of a non-null/non-empty `displayUnit` overrides `requireUnits`

If the question has a `displayUnit`, the unit dropdown is *essentially* disabled - it's default value is set to the `displayUnit`,
the colour is changed so the dropdown doesn't stand out as much, the caret is removed, it is permanently closed etc.

When there is a `displayUnit` the dropdown is still registered as clickable. This isn't ideal but the same can be said for the
numeric modal help button thingy (nothing happens when it gets clicked, but changes the mouse pointer etc.), so unless 
there's an easy fix I don't think it matters that much.

### DIFF
--- a/src/IsaacApiTypes.tsx
+++ b/src/IsaacApiTypes.tsx
@@ -127,6 +127,7 @@ export interface IsaacNumericQuestionDTO extends IsaacQuestionBaseDTO {
     requireUnits?: boolean;
     availableUnits?: string[];
     knownUnits?: string[];
+    displayUnit?: string;
 }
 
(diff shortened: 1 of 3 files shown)


bh412:
@chrisjpurdy I've pushed a commit with an alteration to prevent the dropdown being clickable, with some associated style changes. I think it slightly neatens the logic but if you disagree please do revert the changes!

chrisjpurdy (commented):
The CSS change is neat and doesn't change the dropdown visually while keeping it un-selectable - @bh412 cheers for fixing that!

PR #421, opened 2021-07-28
Set assignments table view

chrisjpurdy:
I have moved a lot of the code and logic from MyGameboards across to this, editing to
remove filtering by completion as I guessed that it made no sense here. I feel like sorting
by number of assigned groups would be nice to take the place of sorting by completion
(but this requires an API change).

Improvements might be to perhaps refactor out the grid and card view patterns from 
SetAssignments and MyGameboards to be more maintainable, and add a regression test
or two for the modal behaviour.


chrisjpurdy:
This is done apart from refactoring it to use `ActiveModal` instead of one inline. I might not do this, because the `ActiveModal` doesn't seem to re-render when one of the components inside changes (this might just be in my specific case).

Morpheu5 (approved):
Works as expected and the code looks sensible.

bh412 (changes requested):
is the changed `package-lock.json` needed, it's a lot of lines! I think it would be better to default to card view rather than table view for desktop as well, so no one is confused by the sudden change! Also, the card layout appears to have broken (on CS at least) with the cards now touching, hopefully a simple style fix!

chrisjpurdy:
I've fixed the CSS bug, and the layout looks all good now - with the change to `package-lock.json` I think that's all the culminated changes to `package-lock.json` from all of the other commits merged into this branch from master.

bh412:
Works well!

PR #422, opened 2021-07-29
Feature/quiz rubric addition

bh412:
If a rubric is present on an isaacQuiz, render it on the first summary page as a content object.

### DIFF
--- a/src/app/components/elements/quiz/QuizAttemptComponent.tsx
+++ b/src/app/components/elements/quiz/QuizAttemptComponent.tsx
@@ -125,8 +125,7 @@ function QuizRubric({attempt}: {attempt: QuizAttemptDTO}) {
 }
 
 function QuizSection({attempt, page}: { attempt: QuizAttemptDTO, page: number }) {
-    const allSections = attempt.quiz?.children;
-    const sections = allSections?.filter(section => section["type"] === "isaacQuizSection");
+    const sections = attempt.quiz?.children;
     const section = sections && sections[page - 1];
     return section ?
         <Row className="question-content-container">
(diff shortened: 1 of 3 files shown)


src/app/components/elements/quiz/QuizAttemptComponent.tsx
```diff
@@ -105,6 +111,19 @@ function QuizHeader({attempt, preview}: QuizAttemptProps) {
     }
 }
 
+function QuizRubric({attempt}: {attempt: QuizAttemptDTO}) {
+    const rubric = attempt.quiz?.rubric;
+    if (rubric) {
+        return <div>
```
jsharkey13, line 117: It doesn't matter either way (and I like explicit ifs more than inline ones on the whole), but elsewhere we do a lot of conditional adding elements, to avoid repeating bits.
```
return <div>
    {rubric && <IsaacContentValueOrChildren ...>
        ...
    </IsaacContentValueOrChildren>}
</div>
```
which I think would work like that here, and save the need for returning an empty div separately. But it ought to be exactly equilvalent, so I point it out only as a "did you know this was an alternative"? :)

jsharkey13 (approved):

src/app/components/elements/quiz/QuizAttemptComponent.tsx
jsharkey13: Oh, and if you _did_ know and chose not to for a good reason, please tell me (so that I can learn something too)! 🙌

jsharkey13: Also, a more important point, after looking at the editor; what happens if the rubric is an empty content object (no value or children, but not undefined or null). Will ValueOrChildren just return something sensible and empty?

bh412: That's just me, I didn't think to have the div outside the conditional! As far as I could bodge it testing, doing it this way doesn't let you have no value / children when removing it with the cross, it's just empty, so you'd have to go into the json to actually break it - though if you managed to / think that would be the case let me know!

PR #423, opened 2021-07-29
Add simple filter to content errors

chrisjpurdy:
A really small change to add a filter to the admin content errors page. The filter checks whether the search string is included in any of the error messages in the list inside each `ContentErrorItem`

### DIFF
--- a/src/app/components/pages/AdminContentErrors.tsx
+++ b/src/app/components/pages/AdminContentErrors.tsx
@@ -1,11 +1,11 @@
-import React, {useEffect} from "react";
+import React, {useEffect, useState} from "react";
 import {AppState} from "../../state/reducers";
 import {getAdminContentErrors} from "../../state/actions";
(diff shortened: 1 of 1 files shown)


src/app/components/pages/AdminContentErrors.tsx
```diff
@@ -35,6 +35,9 @@ export const AdminContentErrors = () => {
     useEffect(() => {dispatch(getAdminContentErrors());}, [dispatch]);
     const errors = useSelector((state: AppState) => state?.adminContentErrors || null);
 
+    const [errorFilter, setErrorFilter] = useState<string>("");
```
mlt47, line 38: No need to specify string as the type parameter here as TS can infer it from the initial value

mlt47 (approved):
Nice and neat

Morpheu5 (approved):

PR #425, opened 2021-08-09
Add Boolean logic notation preference to My account page 

chrisjpurdy:
Backend change [here](https://github.com/isaacphysics/isaac-api/pull/375).

This adds an option on the "My account" page to change how Boolean logic is displayed - either using engineering, maths, or exam board specific notation.

The choice is stored as a map from each option to a Boolean value. In the case of multiple `true` values, the option occurring last (in the order they are listed above) is selected. I also added a `userPreferences` selector to neaten things a bit (removed in the latest commit for name clash and relative uselessness compared to other ways of accessing preferences).

The user needs to have the `BOOLEAN_NOTATION` beta feature enabled to see the "My account" option, and for their selection to affect which katex macros are used.

I'm wondering if line 162 in `UserDetails.tsx` (`value = ["ENG", ...`) gets run more times than needed, and if so whether it matters/could be changed?

### DIFF
--- a/src/app/components/elements/LaTeX.tsx
+++ b/src/app/components/elements/LaTeX.tsx
@@ -267,7 +267,7 @@ export function katexify(html: string, user: PotentialUser | null, examBoard: EX
                 const latexMunged = munge(latexUnEntitied);
                 let macrosToUse;
                 if (SITE_SUBJECT == SITE.CS) {
-                    if (user?.loggedIn && booleanNotation && !booleanNotation.BOARD_SPECIFIC) {
+                    if (user?.loggedIn && booleanNotation) {
                         macrosToUse = booleanNotation?.ENG ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
                     } else {
                         macrosToUse = examBoard == EXAM_BOARD.AQA ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
@@ -330,7 +330,7 @@ export function katexify(html: string, user: PotentialUser | null, examBoard: EX
 
 export function LaTeX({markup}: {markup: string}) {
     const user = useSelector(selectors.user.orNull);
-    const booleanNotation = useSelector((state: AppState) => (state?.userPreferences?.BETA_FEATURE?.BOOLEAN_NOTATION &&
+    const booleanNotation = useSelector((state: AppState) => (state?.userPreferences?.BETA_FEATURE?.AUDIENCE_CONTEXT &&
         state?.userPreferences?.BOOLEAN_NOTATION) || null);
     const screenReaderHoverText = useSelector((state: AppState) => state && state.userPreferences &&
         state.userPreferences.BETA_FEATURE && state.userPreferences.BETA_FEATURE.SCREENREADER_HOVERTEXT || false);
--- a/src/app/components/elements/LaTeX.tsx
+++ b/src/app/components/elements/LaTeX.tsx
@@ -3,7 +3,7 @@ import {useSelector} from "react-redux";
 import {selectors} from "../../state/selectors";
 import {AppState} from "../../state/reducers";
 import {useUserContext} from "../../services/userContext";
-import {FigureNumberingContext, FigureNumbersById, PotentialUser} from "../../../IsaacAppTypes";
+import {FigureNumberingContext, FigureNumbersById, PotentialUser, UserPreferencesDTO} from "../../../IsaacAppTypes";
 import {EXAM_BOARD} from "../../services/constants";
 import he from "he";
 import {SITE, SITE_SUBJECT} from "../../services/siteConstants";
@@ -225,7 +225,7 @@ const ENDREF = "==ENDREF==";
 const REF_REGEXP = new RegExp(REF + "(.*?)" + ENDREF, "g");
 const SR_REF_REGEXP = new RegExp("start text, " + REF_REGEXP.source + ", end text,", "g");
 
-export function katexify(html: string, user: PotentialUser | null, examBoard: EXAM_BOARD | null, screenReaderHoverText: boolean, figureNumbers: FigureNumbersById) {
+export function katexify(html: string, user: PotentialUser | null, userPreferences : UserPreferencesDTO | null, examBoard: EXAM_BOARD | null, screenReaderHoverText: boolean, figureNumbers: FigureNumbersById) {
     start.lastIndex = 0;
     let match: RegExpExecArray | null;
     let output = "";
@@ -262,7 +262,11 @@ export function katexify(html: string, user: PotentialUser | null, examBoard: EX
                 const latexMunged = munge(latexUnEntitied);
                 let macrosToUse;
                 if (SITE_SUBJECT == SITE.CS) {
-                    macrosToUse = examBoard == EXAM_BOARD.AQA ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
+                    if (user?.loggedIn && !userPreferences?.BOOLEAN_NOTATION?.BOARD_SPECIFIC) {
+                        macrosToUse = userPreferences?.BOOLEAN_NOTATION?.ENG ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
+                    } else {
+                        macrosToUse = examBoard == EXAM_BOARD.AQA ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
+                    }
                 } else {
                     macrosToUse = KatexBaseMacros;
                 }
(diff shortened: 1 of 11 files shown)


chrisjpurdy:
Still need to add beta feature flag

src/app/components/pages/MyAccount.tsx
```diff
@@ -161,6 +164,17 @@ const AccountPageComponent = ({user, updateCurrentUser, getChosenUserAuthSetting
         setMyUserPreferences({...myUserPreferences, SUBJECT_INTEREST: newSubjectInterests});
     }
 
+    function setBooleanNotation(newBooleanNotation: BooleanNotation) {
+        // Makes a new object, with all the boolean notation flags being false apart
+        // from those that are set as true in the newBooleanNotation parameter
+        const fullNewBooleanNotation = ["ENG", "MATH", "BOARD_SPECIFIC"]
+            .reduce((acc, key) => acc[key as keyof BooleanNotation] ? acc : {...acc, [key]: false}, newBooleanNotation);
```
mlt47, line 171: I suppose a constant could be re-used with everything set to false (as long as you're careful not to mutate it :) )

src/app/components/elements/panels/UserDetails.tsx
```diff
@@ -140,6 +149,31 @@ export const UserDetails = (props: UserDetailsProps) => {
...
+                        onChange={(event: ChangeEvent<HTMLInputElement>) => {
+                                setBooleanNotation({[event.target.value]: true})
+                            }
+                        }
+                        invalid={submissionAttempted && !validateBooleanNotation(booleanNotation)}
+                    >
+                        <option value={undefined}></option>
+                        <option value={"BOARD_SPECIFIC"}>Exam board specific</option>
+                        <option value={"MATH"}>And (&and;) Or (&or;) Not (&not;)</option>
+                        <option value={"ENG"}>And (&middot;) Or (+) Not (bar)</option>
```
mlt47, line 173: You could use a string enum for the possible options, this will allow you to iterate over it and make sure that we update all places in the future if we add an option or change a value - particularly handy with common strings like "math" which I imagine pops up all over ourcodebase!

src/app/components/elements/LaTeX.tsx
```diff
@@ -225,7 +230,7 @@ const ENDREF = "==ENDREF==";
 const REF_REGEXP = new RegExp(REF + "(.*?)" + ENDREF, "g");
 const SR_REF_REGEXP = new RegExp("start text, " + REF_REGEXP.source + ", end text,", "g");
 
-export function katexify(html: string, user: PotentialUser | null, examBoard: EXAM_BOARD | null, screenReaderHoverText: boolean, figureNumbers: FigureNumbersById) {
+export function katexify(html: string, user: PotentialUser | null, examBoard: EXAM_BOARD | null, booleanNotation : BooleanNotation | null,  screenReaderHoverText: boolean, figureNumbers: FigureNumbersById) {
```
mlt47, line 233: As this function now has a lot of arguments it might make sense to start to think about alternatives. A custom hook that looks a little like `useKatex(markdown: string): string` and then calls the other hooks inside it is an option. This does require more testing boilerplate though and, as this is one of the few functions which is reasonably well tested in the front-end, it's probably not worth making the change now unless you're particularly interested in figuring out how best to test custom hooks :)

src/app/services/validation.ts
```diff
@@ -92,7 +99,8 @@ export function allRequiredInformationIsPresent(user?: ValidationUser | null, us
     return user && userPreferences &&
         (SITE_SUBJECT !== SITE.CS || (validateUserSchool(user) && validateUserGender(user) && validateExamBoard(user))) &&
         (userPreferences.EMAIL_PREFERENCE === null || validateEmailPreferences(userPreferences.EMAIL_PREFERENCE)) &&
-        (SITE_SUBJECT !== SITE.CS || validateSubjectInterests(userPreferences.SUBJECT_INTEREST));
+        (SITE_SUBJECT !== SITE.CS || validateSubjectInterests(userPreferences.SUBJECT_INTEREST)) &&
+        (SITE_SUBJECT !== SITE.CS || (userPreferences.BETA_FEATURE?.BOOLEAN_NOTATION === null || validateBooleanNotation(userPreferences.BOOLEAN_NOTATION)));
```
mlt47, line 103: The method allRequiredInformationIsPresent(...) is also used for deciding whether or not to display a modal to the user on sign in. This was necesary as we allow for registration via Google but we have other required fields that they need to be set. As we have already created the account at that point we just show them a modal to ask them to set it quite regularly until they do. The trouble is there is no way to select the boolean notation option in that modal yet so you'll need to add that :)
Going to dev tools > application > local storage will allow you to delete requiredModalShownTime to see it again without needing to wait the required time!
Also I see that modal even when I don't have BOOLEAN NOTATION as a beta feature on my account so worth double checking that works as you'd expect.

mlt47 (changes requested):
Some small suggestions, but good work

src/app/components/pages/MyAccount.tsx
chrisjpurdy: Good point!

src/app/components/elements/LaTeX.tsx
chrisjpurdy: I think I'll keep it as is for now 👍 I'd be interested in figuring that out at some point if this function comes up again.

src/app/services/validation.ts
chrisjpurdy: image

chrisjpurdy: Added in the upcoming commit - just wondering whether a choice of Boolean logic notation should even be a required account setting, since it is quite specific?

mlt47: That looks neat. But yes, after thinking about the GCSE student case in particular (where the boolean logic notation option is not actually relevant to them) you might be right and we should not "require" the setting of that option. It can be there on the account page if they want to change it but they should not be forced to consider it. It is worth confirming this with Allen of the RPF team.
If we are not making it required, I would suggest also having a DB migration script which gives existing users the notation that they are used to seeing depending on their exam board (be careful as there is both an exam board field in the users table and as a user_preference, I think we use the users one - we'll remove and replace both for a new field with my changes soon which will clear things up a little). This will make sure that the site doesn't change behaviour suddenly for them without them triggering that change.

Ah well, good to have refactored out the component anyway!

mlt47 (commented):
I've made a few further comments after reviewing Ben's similar PR more thoroughly

src/app/components/elements/inputs/BooleanNotationInput.tsx
```diff
@@ -0,0 +1,35 @@
...
+                // is mapped to true in the current value of booleanNotation
+                Object.keys(BOOLEAN_NOTATION).reduce((val: string | undefined, key) => booleanNotation[key as keyof BooleanNotation] === true ? key : val, undefined)
+            }
+            onChange={(event: ChangeEvent<HTMLInputElement>) => {
+                setBooleanNotation({[event.target.value]: true})
+            }
+            }
+            invalid={submissionAttempted && !validateBooleanNotation(booleanNotation)}
+        >
+            <option value={undefined}></option>
```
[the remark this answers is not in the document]
chrisjpurdy: I've removed the `undefined` option in the latest commits - the 'default' is an option that sets both flags in `booleanNotation` to `false`

src/app/components/elements/inputs/BooleanNotationInput.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return <FormGroup>
+        <Label className="d-inline-block pr-2 form-required" htmlFor="boolean-notation-preference">
+            Boolean logic notation
+        </Label>
+        <Input
+            type="select" name="select" id="boolean-notation-preference"
+            value={
+                // This chooses the last string in this list that (when used as a key)
+                // is mapped to true in the current value of booleanNotation
+                Object.keys(BOOLEAN_NOTATION).reduce((val: string | undefined, key) => booleanNotation[key as keyof BooleanNotation] === true ? key : val, undefined)
```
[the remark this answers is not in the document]
chrisjpurdy: I've removed that `=== true` (my bad) and have changed the default value to a new `BOOLEAN_NOTATION.NONE`, which doesn't have a corresponding flag in the `BooleanNotation` type but corresponds to both flags set to `false`

src/app/components/elements/LaTeX.tsx
```diff
@@ -262,7 +267,11 @@ export function katexify(html: string, user: PotentialUser | null, examBoard: EX
                 const latexMunged = munge(latexUnEntitied);
                 let macrosToUse;
                 if (SITE_SUBJECT == SITE.CS) {
-                    macrosToUse = examBoard == EXAM_BOARD.AQA ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
+                    if (user?.loggedIn && booleanNotation) {
+                        macrosToUse = booleanNotation?.ENG ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
+                    } else {
+                        macrosToUse = examBoard == EXAM_BOARD.AQA ? KatexMacrosWithEngineeringBool : KatexMacrosWithMathsBool;
+                    }
```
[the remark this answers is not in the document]
chrisjpurdy: Sorted!

src/app/components/elements/inputs/BooleanNotationInput.tsx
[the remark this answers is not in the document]
chrisjpurdy: `<option value={BOOLEAN_NOTATION.NONE}>No preference</option>` - line 39 in `BooleanNotationInput.tsx`

mlt47 (approved):
Merged into `september` and altered slightly. Will now close this PR to try and avoid confusion.

mlt47:
Merged into `september` will get merged into `master` in time

PR #426, opened 2021-08-10
Add physics sig fig feedback changes

bh412:


src/app/components/content/IsaacQuestion.tsx
```diff
@@ -30,13 +30,19 @@ export const IsaacQuestion = withRouter(({doc, location}: {doc: ApiTypes.IsaacQu
...
     const fastTrackInfo = useFastTrackInformation(doc, location, canSubmit, correct);
 
-    const sigFigsFeedback = <p>
+    const tooManySigFigsFeedback = <p>
         Whether your answer is correct or not, it has the wrong number of&nbsp;
         <strong><a target='_blank' href='/solving_problems#acc_solving_problems_sig_figs'> significant figures</a></strong>.
     </p>;
 
+    const tooFewSigFigsFeedback = <p>
+        We can&apos;t mark this until you provide more sig figs
```
jsharkey13, line 43: I don't think they meant this message to be used verbatim 😅 
I'd check with the content team the preferred phrasing (and it mustn't contain abbreviations like "sig figs" and probably still needs to link to the sig figs page since it will be the commonest case).

jsharkey13 (approved):
I don't think there is anything technical wrong with this (although does the `&nbsp;` and the `...> significant figures` not lead to double spacing, or does the browser remove it?).

I still don't think the new sig fig message is particularly nice or clear to the user in what it means, though.

PR #427, opened 2021-08-10
CS GCSE Navigation changes

chrisjpurdy:
This should stay as a draft until I've shown it to the CS content team
---

Details to requested changes are [here](https://docs.google.com/presentation/d/1sI4WHCkkWhaFA0JrKqyvRW1y7Forj5ZiAqnstWgP5mQ/edit#slide=id.ge496edc7ba_0_38).

I've kept to the specification apart from keeping the featured question, and have instead hidden it more (requiring the user to explicitly click the tab on the homepage - they are always shown the CS journeys tab first).

The NCCE logo is pulled from it's respective site, and the TC logo is a huge inline SVG - I'm not sure the best way to store these or where things like this are normally stored, but this approach will work if released as is. I've allowed the TC and NCCE links to allow referrer information, which is probably a good thing (they can know that we've referred users to them via our homepage).


chrisjpurdy:
The conflict was just that a link to the safeguarding page was added to the Events menu dropdown in `master` - this branch didn't have it, so I just added it in.

src/app/components/pages/AllTopics.tsx
```diff
@@ -1,13 +1,30 @@
...
+    const stage = (() => {
+        switch (stageString) {
+            case STAGE.GCSE:
+                return STAGE.GCSE;
+            case STAGE.A_LEVEL:
+                return STAGE.A_LEVEL;
+            default:
+                return STAGE.A_LEVEL;
+        }
+    })();
```
mlt47, line 27: `const stage = stageString === STAGE.GCSE ? STAGE.GCSE : STAGE.A_LEVEL;

src/app/components/elements/FeaturedContentTabs.tsx
```diff
@@ -15,11 +15,25 @@ export function FeaturedContentTabs() {
...
+                            <IsaacContent doc={cserOfTheMonth} />
+                        </div>
+                    }}
+                    ifNotFound={<div className="computer-scientist-of-the-month mt-4 mb-5 text-center">
+                        Unfortunately, we don't currently have a Computer Science Journey to display.<br />
+                        Please check back later!
+                    </div>}
+                />,
+
                 "Featured question": <Row className="feattab-row">
```
mlt47, line 37: Ask in the meeting but I think we'll end up removing the Featured Question for its maintenance cost

src/app/components/pages/AllTopics.tsx
```diff
@@ -61,16 +86,20 @@ export const AllTopics = () => {
...
-            </Row>
+            <Tabs className="pt-3" tabContentClass="pt-3">{{
+                Isaac: <Row>
+                    <Col lg={{size: 8, offset: 2}} className="bg-light-grey py-md-4 d-md-flex">
+                        {topicColumn(firstColTags)}
+                        {topicColumn(secondColTags)}
+                    </Col>
+                </Row>,
+                AQA: <div className="bg-light-grey py-md-3"><PageFragment fragmentId="a_level_specification_aqa" /></div>,
+                OCR: <div className="bg-light-grey py-md-3"><PageFragment fragmentId="a_level_specification_ocr" /></div>
```
mlt47, line 101: These will also need to be conditional on stage. We can ask at the meeting which exam board specifications GCSE would like to produce for the initial release.

src/app/components/pages/AllTopics.tsx
```diff
@@ -61,16 +86,20 @@ export const AllTopics = () => {
...
-                </Col>
-            </Row>
+            <Tabs className="pt-3" tabContentClass="pt-3">{{
+                Isaac: <Row>
+                    <Col lg={{size: 8, offset: 2}} className="bg-light-grey py-md-4 d-md-flex">
+                        {topicColumn(firstColTags)}
+                        {topicColumn(secondColTags)}
+                    </Col>
+                </Row>,
+                AQA: <div className="bg-light-grey py-md-3"><PageFragment fragmentId="a_level_specification_aqa" /></div>,
```
mlt47, line 100: Let's create the `a_level_specification_aqa` page fragment for them from the existing pages. We might need to use a `<Row>` and `<Col lg={{size: 8, offset: 2}}>` rather than a div for it to look correct.

src/app/components/site/cs/HomepageCS.tsx
```diff
@@ -16,64 +16,99 @@ export const HomepageCS = () => {
...
+                                                    </svg>
+                                                </a>
+                                            </Col>
+                                            <Col>
+                                                <a className="w-100 w-md-75 w-lg-100" href="https://teachcomputing.org"
+                                                   target="_blank" rel="noopener">
+                                                    <img className="w-100" alt="NCCE Logo"
+                                                         src="https://www.stem.org.uk/sites/default/files/fieldable-panel-panes/rich-text/NCCE%20logo.png"/>
+                                                </a>
+                                            </Col>
```
mlt47, line 65: These logos take up a bit too much screen real-estate at the moment, try shrinking them to match the mock up or a little less.
We want to promote our sister projects but would still prefer it if they were to use the resources on our site first! :)

mlt47: As for the storage of logos, they should probably be in /public/assets (NCCE already exists at bottom of the homepage, for example). James is really the expert on deciding the best image formats and what is an acceptable download size for these though

src/app/components/site/cs/HomepageCS.tsx
```diff
@@ -16,64 +16,99 @@ export const HomepageCS = () => {
...
-                            title="Isaac Computer Science introduction video" width="640" height="345" className="mw-100 pt-lg-4"
-                            src="https://www.youtube-nocookie.com/embed/nW4J-NVDziw?enablejsapi=1&rel=0&fs=1&modestbranding=1&origin=home"
-                            frameBorder="0" allowFullScreen
-                        />
-                    </Col>
-                </Row>
+                {user && user.loggedIn ?
+                    <Container className="pb-2 pt-4">
+                        <h1>Welcome {user.givenName}</h1>
+                    </Container>
```
mlt47, line 22: For the logged in user, the homepage is looking a bit bland now.
On a separate branch let's try an alternative view: stacking the three buttons on the left and adding /public/assets/ics_hero.svg on the right so that we can see what that looks like in the meeting. The ics_hero.svg was on the homepage when we first launched isaac cs, you could look back through git-history time to copy the layout.

src/app/components/site/cs/HomepageCS.tsx
```diff
@@ -16,64 +16,99 @@ export const HomepageCS = () => {
...
+                            <Button size="lg" tag={Link} to={"/"} color="secondary" block>
+                                A Level resources
+                            </Button>
+                        </div>
+                        <div className="flex-fill w-100 w-md-auto p-2">
+                            <Button size="lg" tag={Link} to={"/events"} color="primary" outline block>
+                                Events
+                            </Button>
+                        </div>
+                    </Container>
```
mlt47, line 98: It is important that these buttons are emphasised enough. On the mocked page they actually had white text but you'll need to increase the font-size for the contrast to be accessible on the magenta-ish background. Maybe the shrinking of the other logos will make these buttons more prominent again and so no change will be necessary. Worth talking about briefly at the meeting with the content team, though.

src/app/components/pages/AllTopics.tsx
mlt47: Just a suggestion for now until we expand the stages further in the future!

src/app/components/site/cs/HomepageCS.tsx
chrisjpurdy: After the meeting the TC logo is now in the footer (NCCE logo was already there), and the TC logo has been added to assets.

src/app/components/site/cs/HomepageCS.tsx
chrisjpurdy: I've made this change in the most recent commit - it looks a lot better now, although I need opinions on the responsiveness and exact design choices.

src/app/components/site/cs/HomepageCS.tsx
chrisjpurdy: I've changed these buttons to all be the primary colour (an orangey-yellow). The text stands out much more and the design is more in keeping with the homepage in general. 
In the meeting Allen suggested an alternative to have all the buttons as transparent with thick black outlines, this would be an easy change if they don't like my choice here.

src/app/components/pages/AllTopics.tsx
chrisjpurdy: I think I already did this? Did it look right in the demo - if not I'll make this change!

src/app/components/elements/FeaturedContentTabs.tsx
chrisjpurdy: Sorry! I didn't bring this up - I'll keep it there for now and see if they get back to us on it?

src/app/components/pages/AllTopics.tsx
mlt47: The exam boards for GCSE are in the constant EXAM_BOARDS_CS_GCSE, and unfortunately they are not the same as A Level but yep quick conditional will do the trick. Construct a fragment ID using the exam board and the stage, if you make him the fragments to edit (maybe copy an existing a level specification there to encourage consistency)  that would be even better.

EXAM_BOARDS_CS_A_LEVEL holds 5 exam boards, we are now only sticking to AQA and OCR for September. Possibly introduce the others by the new year. Feel free to comment out the other three: [aqa, ocr, /* other exam boards */]

src/app/components/pages/AllTopics.tsx
mlt47: It didn't have the offset but I think it looked OK, maybe as the page used to be full width it'll look better like that. You can try the other style to see what looks best to you.

src/app/components/site/cs/HomepageCS.tsx
mlt47: Cool 👍

mlt47:
Merged into `september` branch. Any additional changes should start from there.

PR #428, opened 2021-08-11
Add an Isaac CS logo themed spinner instead of the normal one

chrisjpurdy:
When I saw the Isaac CS logo, this popped into my head for some reason so I went ahead and did it - I don't mind if it gets scrapped but I thought it would be fun 🎆 

The commit overrides the default Reactstrap spinner with my new one if the site is CS. The colour could be changed easily if needed. One thing I didn't know and might need to add is anything the Reactstrap spinner had for accessibility etc.


src/scss/cs/isaac.scss
```diff
@@ -288,3 +288,8 @@ body {
         background-repeat: no-repeat;
     }
 }
+
+#cs-spinner {
```
mlt47, line 292: You might want to make this a class rather than an ID as it is plausible, although a little ugly, to have multiple spinners on a page!

mlt47 (commented):
That's fun!
Not sure about accessibility, bootstrap docs might have something about it. There are sometimes really useful blogs about accessibility. You could also try to use the site with a screen reader and compare - NVDA is a common one for Windows.
Maybe you show it to to the content team / at the tech meeting.

src/scss/cs/isaac.scss
chrisjpurdy: Just sorted that out in the last commit - weirdly when I was testing it didn't seem to complain, and essentially treated the id as a class (multiple spinners with the same id happy inherited the CSS)

mlt47: Browsers tend to be quite lenient (see [tag soup](https://en.wikipedia.org/wiki/Tag_soup)), until they are not!  It is usually libraries that I've found which sometimes break catastrophically if the one ID per page contract is broken.

chrisjpurdy: Noted! 😄

PR #432, opened 2021-08-20
Override default modal handing of escape key and backdrop click

chrisjpurdy:
This is to make sure the scrollbar is unlocked on close.

This feels like a patch on top of an issue with how 'toggle' works, since that is what gets called in the Reactstrap `Modal` class component. It works now but there should be a shorter and neater (but probably more complicated) solution?

### DIFF
--- a/src/app/components/elements/modals/ActiveModal.tsx
+++ b/src/app/components/elements/modals/ActiveModal.tsx
@@ -34,17 +34,26 @@ export const ActiveModal = ({activeModal}: ActiveModalProps) => {
             // @ts-ignore
             modalRef.current.handleEscape = (e) => {
                 // @ts-ignore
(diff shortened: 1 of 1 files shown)


mlt47 (commented):
Well done for getting this to work!
However, it's probably best not to depend on the _private variables of libraries. They could change those values in a [patch or a minor release](https://semver.org/) which our package manager regularly updates automatically. In this case we would be no worse off than we currently are, but still best not to do it.

Your intuition was correct that we were doing something silly with toggle - probably me!
I've managed to find an alternative solution which seems to work:
https://github.com/isaacphysics/isaac-react-app/commit/234d8b94742b71d682479c3bebd5a9f4edfbf978

mlt47:
Alternative solution on September branch

PR #438, opened 2021-09-03
Hotfix for share link button

chrisjpurdy:
This fixes the problem with the share text being de-attached from the share button, and with the click-away to close not working smoothly


Morpheu5 (approved):

bh412 (changes requested):
Sorry to be a pain but when you go down to mobile view with the cards on Set Assignments / My Gameboards, the share link button now hangs over the right of the card, probably a quick fix. Once that's sorted I'm happy with it! (I stopped the titles being quite so width restricted on cards, but that didn't cause this overhang)

PR #441, opened 2021-09-07
Hotfix/safari infinite loop fix

bh412:


Morpheu5 (approved):
Tested on Safari 13.1.2 which is the oldest I have access to.

mlt47:
I must admit I had seen @bh412 make a merge between this and master and assumed it was the other way around so thought this PR was merged into master ages ago...
Analytics hasn't reported seeing this client side error much at all since the emergency fix in early September, but I prefer the code in this PR and it might get the number down to zero so I would like to get it merged.
If you are happy to merge it Ben, I'm happy to take responsibility of fixing and re-releasing if it does happen to cause errors for a different browser version this time. It shouldn't, as we're now explicitly giving the browser a chance to do its own thing before we check query param values, but there is always space for browser implementation details to cause trouble with this sort of task.

PR #442, opened 2021-09-08
Checks for canvas before trying to remove it

mlt47:
We have been seeing front-end errors reporting:

TypeError: Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.
  at.componentWillUnmount (https://isaacphysics.org/...

This should fix that but I have not determined the true cause to reproduce.

### DIFF
--- a/src/app/components/elements/modals/InequalityModal.tsx
+++ b/src/app/components/elements/modals/InequalityModal.tsx
@@ -464,8 +464,9 @@ class InequalityModalComponent extends React.Component<InequalityModalProps> {
             }));
             this.setState({ sketch: null });
         }
-        if (inequalityElement) {
-            inequalityElement.removeChild(inequalityElement.getElementsByTagName('canvas')[0]);
+        const canvas = inequalityElement?.getElementsByTagName('canvas')[0];
+        if (canvas) {
+            inequalityElement.removeChild(canvas);
         }
 
         document.documentElement.style.width = '';


mlt47:
Hi @Morpheu5, I've requested your review on this to sanity check the change. It can wait until you're next in.
Do you know what could cause the canvas not to be a child of the referenced DOM element so that we can reproduce the behaviour and make sure this fixes the error? It is possible that it is browser specific behaviour - that could be checked from the logged_event records' userAgent strings if you suspect this.

Morpheu5:
> Do you know what could cause the canvas not to be a child of the referenced DOM element

Last time I reviewed this mechanism, I think I went over all the attach/detach code and was satisfied that this should not happen. p5 works in mysterious ways so it's possible that the canvas element is not garbage collected correctly before or after being removed from the canvas. For all practical purposes, the overall behaviour is correct -- we used to have a host of duplicate canvasses lying around catching mouse events, and I should have solved that, but I can't tell if there are retained canvasses within p5. Depending on how occasional the errors are, there may be a deallocation race condition and we got unlucky.

Anyway, this change should get rid of the error. The error itself is probably just a symptom of the underlying condition, but I think we can put up with it as long as the users are not impacted (not disproportionally, at least).

> It is possible that it is browser specific behaviour - that could be checked from the logged_event records' userAgent strings if you suspect this.

I would be quick to point at browsers too in this case, so that may be a good thing to look for. Do we have a quick way of filtering for this specific error?

src/app/components/elements/modals/InequalityModal.tsx
```diff
@@ -464,8 +464,9 @@ class InequalityModalComponent extends React.Component<InequalityModalProps> {
             }));
             this.setState({ sketch: null });
         }
-        if (inequalityElement) {
-            inequalityElement.removeChild(inequalityElement.getElementsByTagName('canvas')[0]);
+        const canvas = inequalityElement?.getElementsByTagName('canvas')[0];
```
Morpheu5, line 467: Given that, on that page, there should only be canvasses as children of that element, this should fix (or at least mask) the issue. Maybe one way of investigating the issue is to start from `document` or `windows` and remove all the canvasses recursively and see what other errors come up -- there's a remote possibility that the hidden canvas underlying the text entry thing is the culprit.

Morpheu5 (approved):
LGTM for a quick fix. See comments for further investigation. I'll have a look at this next time I'm in.

mlt47:
> I would be quick to point at browsers too in this case, so that may be a good thing to look for. Do we have a quick way of filtering for this specific error?

```SQL
SELECT timestamp, event_details->>'userAgent', event_details->>'errorDetails'
FROM logged_events
WHERE event_type = 'CLIENT_SIDE_ERROR'
AND event_details->>'errorDetails' ILIKE '%TypeError: Failed to execute%'
ORDER BY timestamp DESC
LIMIT 300; 
```

PR #443, opened 2021-09-09
Feature/admin password reset

bh412:


src/app/components/pages/AdminUserManager.tsx
```diff
@@ -124,6 +124,12 @@ export const AdminUserManager = () => {
         dispatch(adminUserSearch(searchQuery));
     };
 
+    const attemptPasswordReset = (email: string | undefined) => {
+        if (email) {
```
Morpheu5, line 128: I know this works fine but it would be great if we could start using the `isDefined` function that I made (and I think I added in here somewhere). I came across a couple of instances in which a `number` was `0` so it read falsey despite being defined and `0` being a valid value.

Morpheu5 (changes requested):
Other than that tiny change, this looks fine.

PR #445, opened 2021-09-11
Add physics spinner 🎆 

chrisjpurdy:
A bit of fun!

The behaviour of styling SVG in an `img` tag with stylesheets is really weird and didn't want to play ball, so I added an inline style on PHY so that it behaves - this means that currently the `size` option doesn't affect the size of the spinner on the physics site.


src/app/components/handlers/IsaacSpinner.tsx
```diff
@@ -11,11 +10,8 @@ export interface IsaacSpinnerProps {
...
 export const IsaacSpinner = ({size = "md", className, color = "primary"} : IsaacSpinnerProps) => {
-    return SITE_SUBJECT === SITE.CS ?
-        <div role="status" className="pb-1">
-            <img className={classNames(`isaac-spinner-${size}`, className)} alt="" src="/assets/isaac-cs-typer-css.svg"/>
-            <span className="sr-only">Loading...</span>
-        </div>
-        :
-        <Spinner className={className} color={color} size={size}/>;
+    return <div role="status" className="pb-1">
+        <img style={SITE_SUBJECT === SITE.PHY ? {width: "auto", height: "5.5rem"} : {}} className={classNames(`isaac-spinner-${size}`, className)} alt="" src={SITE_SUBJECT === SITE.CS ? "/assets/isaac-cs-typer-css.svg" : "/assets/isaac-phy-apple-grow.svg"}/>
```
bh412, line 14: Having the size here isn't the prettiest but you say it doesn't work in the scss file so we'll live with it!

bh412 (approved):
I think it looks neat! The style changes don't negatively effect anything on CS - If the content teams decide they would like something a bit different it should be simple to change.

PR #446, opened 2021-09-17
Move quizzes to QuizSummaryDTO and alter rendering logic

bh412:


src/app/components/pages/quizzes/SetQuizzes.tsx
```diff
@@ -145,6 +145,7 @@ const SetQuizzesPageComponent = ({user, location}: SetQuizzesPageProps) => {
                             {filteredQuizzes.map(quiz =>  <RS.ListGroupItem className="p-0 bg-transparent" key={quiz.id}>
                                 <div className="d-flex flex-grow-1 flex-column flex-sm-row align-items-center p-3">
                                     <span className="mb-2 mb-sm-0">{quiz.title}</span>
+                                    {quiz.visibleToStudents && <div className="small text-muted d-none d-md-block ml-2">visible to students</div>}
```
jsharkey13, line 148: I don't think there is enough spacing/separation here? The title and this label run into one another and I don't know what the best way to deal with that is.

jsharkey13 (approved):
The code looks fine, but I am worried the label is too squished into the title?

PR #447, opened 2021-09-21
Add filters for published state of content

chrisjpurdy:
Added two checkboxes to `admin/content_errors`, to show errors for published and/or unpublished content.

### DIFF
--- a/src/app/components/pages/AdminContentErrors.tsx
+++ b/src/app/components/pages/AdminContentErrors.tsx
@@ -36,7 +36,9 @@ export const AdminContentErrors = () => {
     const errors = useSelector((state: AppState) => state?.adminContentErrors || null);
 
     const [errorFilter, setErrorFilter] = useState<string>("");
-    const errorReducer = (show: boolean, errorStr: string) => show || errorStr.toLowerCase().includes(errorFilter.toLowerCase())
+    const errorReducer = (show: boolean, errorStr: string) => show || errorStr.toLowerCase().includes(errorFilter.toLowerCase());
+
+    const [publishedFilter, setPublishedFilter] = useState<number>(2);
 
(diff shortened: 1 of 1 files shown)


src/app/components/pages/AdminContentErrors.tsx
```diff
@@ -36,7 +36,9 @@ export const AdminContentErrors = () => {
     const errors = useSelector((state: AppState) => state?.adminContentErrors || null);
 
     const [errorFilter, setErrorFilter] = useState<string>("");
-    const errorReducer = (show: boolean, errorStr: string) => show || errorStr.toLowerCase().includes(errorFilter.toLowerCase())
+    const errorReducer = (show: boolean, errorStr: string) => show || errorStr.toLowerCase().includes(errorFilter.toLowerCase());
+
+    const [publishedFilter, setPublishedFilter] = useState<number>(2);
```
mlt47, line 41: It might be more efficient to store this state as a number but the code would be more readable if you found a way for the states to be named. An Enum could offer that while maintaining the "efficiency" after compilation, but the transitions between the states would still be a little opaque if the transitions were done using arithmetic as it is done now. It is better to favour readability over efficiency until we need it and, when we do need it, it is unlikely that this optimisation will affect performance much. A select input might be a good alternative interface for this feature?

chrisjpurdy: I thought about a Select input but it seemed a bit heavy? Although it would be more compact and look better.
I definitely agree with the state, I just felt that a three/four value enum was verbose, but on refection it shouldn't take more than a second to understand how this code works and using an enum will fix that.

mlt47 (approved):
Nice and useful!

PR #449, opened 2021-09-28
Add extra information to the admin email sending page

jsharkey13:
This will help ensure emails have the right values configured if they are being customised or sent via different servers.

Depends on https://github.com/isaacphysics/isaac-api/pull/387

### DIFF
--- a/src/app/components/pages/AdminEmails.tsx
+++ b/src/app/components/pages/AdminEmails.tsx
@@ -102,9 +102,16 @@ export const AdminEmails = (props: AdminEmailsProps) => {
 
         {emailTemplateSelector && <>
             <RS.Card className="p-3 my-3">
-                <RS.CardTitle tag="h2">Subject:</RS.CardTitle>
+                <RS.CardTitle tag="h2">Details</RS.CardTitle>
                 <RS.CardBody>
-                    {emailTemplateSelector.subject}
+                    <ul>
+                        <li><b>Subject:</b> {emailTemplateSelector.subject || "no subject"}</li>
+                        {emailTemplateSelector.from && <li><b>From:</b> {emailTemplateSelector.fromName || ""} &lt;{emailTemplateSelector.from}&gt;</li>}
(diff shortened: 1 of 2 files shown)


src/app/components/pages/AdminEmails.tsx
```diff
@@ -102,9 +102,16 @@ export const AdminEmails = (props: AdminEmailsProps) => {
...
-                <RS.CardTitle tag="h2">Subject:</RS.CardTitle>
+                <RS.CardTitle tag="h2">Details</RS.CardTitle>
                 <RS.CardBody>
-                    {emailTemplateSelector.subject}
+                    <ul>
+                        <li><b>Subject:</b> {emailTemplateSelector.subject || "no subject"}</li>
+                        {emailTemplateSelector.from && <li><b>From:</b> {emailTemplateSelector.fromName || ""} &lt;{emailTemplateSelector.from}&gt;</li>}
+                        {emailTemplateSelector.replyTo && <li><b>Reply-To:</b> {emailTemplateSelector.replyToName || ""} &lt;{emailTemplateSelector.replyTo}&gt;</li>}
+                        <li><b>Sent via:</b>&nbsp;
+                            {(emailTemplateSelector.sender && emailTemplateSelector.sender.indexOf("@mail.isaac") > 0) ? "MailGun" : "University"}
```
mlt47, line 112:
```suggestion
                            {emailTemplateSelector.sender?.includes("@mail.isaac") ? "MailGun" : "University"}
```

mlt47 (approved):

src/app/components/pages/AdminEmails.tsx
jsharkey13: Do we transpile this? I deliberately avoided it because it's not Javascript but ES6 syntax, although with no IE only Opera Mini doesn't support it so it's probably fine now?

jsharkey13: Ah, we use it everywhere and it is much nicer!

mlt47: Our tsconfig file says that we target ES5, hopefully we actually use this! If not the default target for TS is ES3.
ES6 is supported by all modern browsers now too, though.

PR #454, opened 2021-10-11
Preserve topic exam board tab by recording it in URL hash

chrisjpurdy:
This was requested in [ticket #204947](https://tickets.isaaccomputerscience.org/scp/tickets.php?id=7078). When the back button is pressed after leaving the CS all topics page, the exam board tab the user was on was not recorded, so the user would have to navigate back to the tab every time. 
This is no longer the case - the active tab is taken from a URL hash (either 'all', or a lowercase exam board name) and this hash is updated when the tabs are changed.

This could be abstracted to a wrapper component around Tab that uses the URL hash to choose the active tab, but this would clash with components that also use the hash for other functions etc. so it doesn't seem worth it yet.

### DIFF
--- a/src/app/components/pages/AllTopics.tsx
+++ b/src/app/components/pages/AllTopics.tsx
@@ -1,5 +1,5 @@
-import React from "react";
-import {Link, useLocation} from "react-router-dom";
+import React, {useEffect} from "react";
+import {Link, useLocation, useHistory} from "react-router-dom";
 import {Badge, Col, Container, Row} from "reactstrap";
 import "../../services/tagsPhy";
 import tags from "../../services/tags";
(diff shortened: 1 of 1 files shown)


mlt47 (commented):
This is good and accomplishes what the user asked for.
The fragment part of the URL, after the hash, would usually be the best place to address an area inside a page. My concern is that we already have the exam board in the URL query param.
I guess the query params indicate viewing context/user settings while the URL path and fragment specifies the content that is being viewed. We have not followed this for `/topics?stage=gcse` and `/topics?stage=a_level` but I think I will have a play to see if we can get it to work that way without breaking old links.

Best to temporarily hold off from merging.

PR #461, opened 2021-10-22
CS Question finder MVP

chrisjpurdy:
Allows filtering by topic, exam board, stage, and difficulty. It looks similar to the teacher gameboard builder, but without the option to search by question name. 
A small back-end change was also needed to allow the gameboards to be generated intuitively considering the user only has the option to select topics ([see this PR](https://github.com/isaacphysics/isaac-api/pull/392)).

The reveal button on the question finder has changed colour a tiny bit (on physics) so that it has the correct-ish colour on both sites. (it might need brightening so that it stands out more perhaps)


src/app/components/pages/Gameboard.tsx
```diff
@@ -32,15 +32,16 @@ function extractFilterQueryString(gameboard: GameboardDTO): string {
...
-    return tags.getByIdsAsHierarchy(docTags as TAG_ID[]);
+    if (SITE_SUBJECT === SITE.PHY) {
+        return tags.getByIdsAsHierarchy(docTags as TAG_ID[]);
+    } else {
+        return tags.getByIdsAsHierarchy(docTags as TAG_ID[]).splice(1);
+    }
 }
 
-const GameboardItemComponent = ({gameboard, question}: {gameboard: GameboardDTO, question: GameboardItem}) => {
+const GameboardItemComponent = ({gameboard, question, showBreadcrumbs}: {gameboard: GameboardDTO, question: GameboardItem, showBreadcrumbs?: boolean}) => {
```
mlt47, line 44: Is there a reason why we don't show breadcrumbs all the time (if they're present)? As there is no GameboardItemComponent or GameboardViewer with showBreadcrumbs explicitly set to false with the logic on line 85 I think this might already be true?

mlt47 (changes requested):
This is good and also includes the concepts options by reusing an existing endpoint which is nice. If we find it is slow we can always optimise later.

There are a few suggested changes. I am also going to have a play with the CS filter layout in case there is another configuration which looks nicer/reads better.

src/app/components/pages/GameboardFilter.tsx
```diff
@@ -105,33 +105,85 @@ function generateCSBoardName(selections: Item<TAG_ID>[][]) {
...
+                        P3 covers later topics.<br />
+                        Challenge questions are solved by combining multiple concepts and creativity.<br />
+                        C1 can be attempted near the beginning of your course,<br />
+                        C3 require more creativity and could be attempted later in a course.
+                    </RS.UncontrolledTooltip>
+                </RS.Label>
+                <Select id="difficulty-selector" onChange={unwrapValue(setDifficulties)} isClearable isMulti value={difficulties} options={DIFFICULTY_ITEM_OPTIONS} />
+            </div>
+        </RS.Col>
+        {SITE_SUBJECT === SITE.PHY && <RS.Col lg={8}>
```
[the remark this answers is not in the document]
chrisjpurdy: I thought I'd removed all of these! Oops

src/app/components/pages/GameboardFilter.tsx
```diff
@@ -26,11 +33,10 @@ function itemiseTag(tag: Tag) {
...
 }
 
-function itemiseConcepts(concepts: string[] | string) {
-    const conceptsList = Array.isArray(concepts) ? concepts : [concepts]
-    return conceptsList
+function itemiseConcepts(concepts: string[]): Item<string>[] {
+    return concepts
         .filter(concept => concept !== "")
-        .map(concept => ({label: concept, value: concept}));
+        .map(concept => ({label: "<!LOADING!>", value: concept}));
```
mlt47, line 39: You can use QUESTION_FINDER_CONCEPT_LABEL_PLACEHOLDER

src/app/services/constants.ts
```diff
@@ -1026,3 +1028,5 @@ export const progressColour = {
 export const GRAY_120 = '#c9cad1';
 
 export const SEARCH_CHAR_LENGTH_LIMIT = 255;
+
+export const QUESTION_FINDER_CONCEPT_LABEL_PLACEHOLDER = "<!LOADING!>";
```
mlt47, line 1032: I personally slightly prefer "Loading..." but don't feel super strongly about it.

mlt47 (approved):

src/app/services/miscUtils.ts
```diff
@@ -27,4 +28,11 @@ export function useOutsideCallback(ref: RefObject<any>, callback : () => void, d
             document.removeEventListener("mousedown", handleClickOutside);
         };
     }, [...deps, ref]);
+}
+
+export function siteSpecific<P, C>(phy: P, cs: C) {
+    if (SITE_SUBJECT === SITE.PHY) {
+        return phy;
+    }
+    return cs;
```
mlt47, line 37: I was initially uneasy about this because of needing to remember the order of positional arguments, however, WebStorm at least tells you the argument names at wherever the method is called. That doesn't happen yet in GitHub which is a little annoying.
The other thing to consider is if/when we add a new subject. For that we can make the new argument's type `, art: NonNullable<A>` to make sure we update all the old calls to this siteSpecific function.
We already have a thing called siteSpecific which is a map of constants which are immediately evaluated to either be the Physics or CS version. We will have to alias or namespace imports if we use both in the same file which is a little annoying and it might get confusing in discussion although ultimately does the same thing. Is there perhaps a suitable alternative name? Maybe the other one becomes siteSpecificComponents or something?
Despite all this, (in WebStorm) it does look neater than short circuited &&s all over the place.
I propose we carry on with this for a little to see how it feels in practice as it is easy enough to revert the decision later.

src/app/components/pages/GameboardFilter.tsx
```diff
@@ -514,50 +527,54 @@ export const GameboardFilter = withRouter(({location}: {location: Location}) =>
...
+                // CS
+                <>
+                    <RS.Col xs={12} lg={"auto"} >
+                        {isEditingTitle
+                            ? <RS.Input defaultValue={customBoardTitle ?? gameboard?.title}
+                                        onChange={e => setPendingCustomBoardTitle(e.target.value)}
+                                        className={"mb-2 mb-lg-0"} />
+                            : <h3>{customBoardTitle ?? gameboard?.title}</h3>
+                        }
+                    </RS.Col>
```
mlt47, line 555: Did you mention to Physics that this was being re-enabled? It can be a separate card now but I think they would be interested

src/app/components/pages/GameboardFilter.tsx
```diff
@@ -504,8 +504,8 @@ export const GameboardFilter = withRouter(({location}: {location: Location}) =>
                 </RS.Row>
                 :
                 <RS.Col xs={12} className={`text-center mt-3 mb-4 m-sm-0`}>
-                    <RS.Button color={"link"} className="filter-action" onClick={() => setFilterExpanded(true)}>
-                        Edit question filters
+                    <RS.Button size="sm" color="primary" outline onClick={() => setFilterExpanded(true)}>
+                        Edit question filters ✎
```
mlt47, line 508: Might this icon also be useful (and if anything more appropriate) for the title editing button?
I would also suggest at least trying to see how that title editing button looks outlined or in a less impactful colour with less padding between it and the title so that it is more clear that they are related.

mlt47 (changes requested):

src/app/services/miscUtils.ts
chrisjpurdy: I was also unsure about whether to consider new subjects or not, but I felt like it would be such a rare occurrence and be so simple to find and replace usages that fully generifying it would be overkill. Hopefully it is easier to use than the &&s!

src/app/components/pages/GameboardFilter.tsx
chrisjpurdy: I didn't, but I thought that I made sure this was only on CS

src/app/components/pages/GameboardFilter.tsx
chrisjpurdy: Good idea - I'll give each a go and see what works best. I feel like it is fairly clear they are related just based on the button text - the reason there is a bit of padding is because I had to pick classes for responsiveness, but I'll give this a go again to see if there's a better solution.

mlt47 (approved):

PR #464, opened 2021-11-08
Feature/content email sending

bh412:


src/app/services/api.ts
```diff
@@ -164,6 +164,9 @@ export const api = {
         sendAdminEmailWithIds: (contentid: string, emailType: string, ids: number[]): AxiosPromise => {
             return endpoint.post(`/email/sendemailwithuserids/${contentid}/${emailType}`, ids);
         },
+        sendContentEmailWithIds: (sendingMethod: string | undefined, plaintextTemplate: string, htmlTemplate: string, subject: string, emailType: string, ids: number[]): AxiosPromise => {
+            return endpoint.post(`/email/sendcontentemailwithuserids/${emailType}`, {userIds: ids, overrideFromAddress: sendingMethod, plainTextContent: plaintextTemplate, htmlContent: htmlTemplate, subject: subject});
```
jsharkey13, line 168: Should this not take an `EmailTemplateDTO` as the type, alongside the list of user IDs? There is a valid one in the frontend, and so using it here and in the component seems sensible?

jsharkey13 (approved):

jsharkey13 (approved):

PR #466, opened 2021-11-09
Render LaTeX in question navigation links with `TrustedHtml`

chrisjpurdy:


bh412 (approved):
Does what it says with no issues!

PR #467, opened 2021-11-11
Float right layout fix

chrisjpurdy:


bh412 (approved):
This works in the example case. It doesn't fix the text under image issue on the book pages on Safari sadly, but wasn't intended to.

PR #470, opened 2021-11-15
Add deprecated warning banner on deprecated content

chrisjpurdy:
As per a Slack conversation with Lee, the superseded and deprecated warning banners are now merged if both need to be shown (hence the rather long component name).

This has gone through copy, so just needs a code review at this point.


Morpheu5 (approved):
The multi-way comparison in the `return` statement is horrendous but otherwise this looks OK to me.

bh412 (approved):
I agree the `return` is a bit roundabout but it works

PR #471, opened 2021-11-19
Combined Markbook

Morpheu5:
This adds a per-group summary view of quiz progress to the old Assignment Progress / My Markbook page.

There is now a significant amount of shared code between `AssignmentProgress.tsx` and `QuizTeacherFeedback.tsx` so I made a new `QuizProgressCommon.tsx` file for better reusability.


src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -24,51 +24,68 @@ import {
 } from "../../../IsaacAppTypes";
 import {selectors} from "../../state/selectors";
```
mlt47, line 25: The nav bar and and page title will need updating (for Physics, at least)

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -503,22 +496,109 @@ export const AssignmentProgressLegend = (props: AssignmentProgressLegendProps) =
...
+
+    useEffect(() => {
+        dispatch(loadQuizAssignmentFeedback(quizAssignmentId));
+    }, [quizAssignmentId]);
+
+    useEffect(() => {
+        if (resourceFound(quizAssignments)) {
+            setUserFeedback(quizAssignments.find(qa => qa.id === quizAssignmentId)?.userFeedback);
+        }
+    }, [quizAssignments])
```
mlt47, line 514: As userFeedback is a "derived" field, it doesn't need a useState(). You can do something like the following and it will always be correct because the changing of the quizAssignments redux state will trigger the re-render, i.e. re-execution of this function. 
```
let userFeedback: QuizUserFeedbackDTO[] = [];
if (resourceFound(quizAssignments)) {
    userFeedback = quizAssignments.find(qa => qa.id === quizAssignmentId)?.userFeedback);
}
    
```

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -503,22 +496,109 @@ export const AssignmentProgressLegend = (props: AssignmentProgressLegendProps) =
...
+            </div>
+        </div>
+        {(isExpanded) && <QuizProgressLoader {...props} />}
+    </div>
+};
+
 const GroupDetails = (props: GroupDetailsProps) => {
     const dispatch = useDispatch();
     const {group, pageSettings} = props;
+    const [activeTab, setActiveTab] = useState(MARKBOOK_TYPE_TAB.assignments);
```
mlt47, line 567: Physics want test to be active first, possibly conditional on whether or not there are any tests, but it is worth clarifying that with them.

mlt47 (changes requested):
Small suggestions but works well

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -24,51 +24,68 @@ import {
 } from "../../../IsaacAppTypes";
```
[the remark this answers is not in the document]
Morpheu5: Yes, the jury's out on the new name, I guess…

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -24,51 +24,68 @@ import {
...
+        }
+
+        const quizAssignments: { [id: number]: QuizAssignmentDTO[] } = {};
+        if (resourceFound(state.quizAssignments)) { // assigned by me
+            for (const qa of state.quizAssignments) {
+                if (isDefined(quizAssignments[qa.groupId || -1])) {
+                    quizAssignments[qa.groupId || -1].push(qa);
+                } else {
+                    quizAssignments[qa.groupId || -1] = [qa];
+                }
```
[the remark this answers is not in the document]
Morpheu5: I agree, I don't like the `-1` either. Having an `as number` would force the cast and I don't think TS likes that. I guess I could add a `defined`ness guard on the `qa.groupId` bit and skip over if that is not defined as in practice we are unlikely to ever use whatever gets added to the `-1` key. OTOH, we lose that information, although it's not like we do much with it.

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -87,7 +104,7 @@ type EnhancedAssignment = AssignmentDTO & {
     progress?: AppAssignmentProgress[];
 };
 
-type AppGroupWithAssignments = AppGroup & {assignments: EnhancedAssignment[]};
+type AppGroupWithAssignments = AppGroup & {assignments: EnhancedAssignment[], quizAssignments: any};
```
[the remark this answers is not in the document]
Morpheu5: Huh, well spotted!

src/app/state/actions/quizzes.tsx
```diff
@@ -238,3 +238,7 @@ export const logQuizSectionView = (quizAttemptId: number, page: number) => async
     // No actual dispatch
     api.quizzes.logQuizSectionView(quizAttemptId, page);
 };
+
+export function getQuizAssignmentCSVDownloadLink(assignmentId: number) {
+    return `${API_PATH}/quiz/assignment/${assignmentId}/download`;
+};
```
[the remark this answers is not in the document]
Morpheu5: Moved it.

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -576,6 +656,7 @@ export function AssignmentProgress(props: AssignmentProgressPageProps) {
     useEffect(() => {
```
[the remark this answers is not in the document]
Morpheu5: There's a cascade of types and subtypes that makes the whole thing very painful to pick apart. I tried for a whole minute and ran away screaming. Some of the types have also drifted apart, with the likes of the `EnhancedAssignment` to include progress which the `QuizAssignmentDTO` does not need because it already includes it, so maybe a full review/rationalization of the types on this page is in order.

src/app/components/pages/AssignmentProgress.tsx
Morpheu5: Physics currently calls it "Assignment Progress", CS calls it "My markbook". It's an easy change, just needs a decision.

src/app/components/pages/AssignmentProgress.tsx
Morpheu5: Good point. I guess I have been resorting to using state by default out of trying to avoid potential issues when not using it.

src/app/components/pages/AssignmentProgress.tsx
Morpheu5: I've asked Matthew to clarify with both content teams.

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -503,22 +496,109 @@ export const AssignmentProgressLegend = (props: AssignmentProgressLegendProps) =
...
+    let groupAssignments: JSX.Element | JSX.Element[] = <div className="p-4 text-center"><IsaacSpinner size="md" /></div>;
+    if (gameboardsLoaded) {
+        if (group.assignments.length > 0) {
+            groupAssignments = group.assignments.map(assignment => hasGameboard(assignment) && <AssignmentDetails key={assignment.gameboardId} {...props} assignment={assignment}/>).filter(e => e) as JSX.Element[];
+        } else {
+            groupAssignments = <div className="p-4 text-center">There are no assignments for this group.</div>
+        }
+    }
+    let groupTests: JSX.Element | JSX.Element[];
+    if (isDefined(group.quizAssignments) && Array.isArray(group.quizAssignments) && group.quizAssignments.length > 0) {
```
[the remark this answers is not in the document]
Morpheu5: Huh, I remember now: this was a [useless](https://lgtm.com/projects/g/isaacphysics/isaac-react-app/rev/pr-260ac0b644b577b16ce04cdf9fc6a2f2757bed9d) alert from LGTM.

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -603,8 +603,8 @@ const GroupDetails = (props: GroupDetailsProps) => {
     return <div className={"assignment-progress-details" + (pageSettings.colourBlind ? " colour-blind" : "")}>
         <AssignmentProgressLegend pageSettings={pageSettings} showQuestionKey={activeTab === MARKBOOK_TYPE_TAB.tests} />
         <Tabs className="my-4 mb-5" tabContentClass="mt-4" activeTabOverride={activeTab} onActiveTabChange={activeTabChanged}>{{
-            "Assignments": groupAssignments,
-            "Tests": groupTests
+            `Assignments (${groupAssignments.length || 0})`: groupAssignments,
+            `Tests (${groupTests.length || 0})`: groupTests
```
mlt47, line 607: I think this commit is broken because JS object keys only get evaluated if they are in `[`...`]`

src/app/components/pages/AssignmentProgress.tsx
mlt47: Physics have decided to continue with the existing names for now.

mlt47 (approved):

PR #474, opened 2021-11-23
Order doc items by id on question load

chrisjpurdy:
This gives a predictable order for items to be in (locale string order), if content want to order them.

This doesn't actually do anything with duplicate ids! I just messed up the branch names.


jsharkey13:
The shuffling of these is [explict and intentional, and easily overridable by a question property in the content object](https://github.com/isaacphysics/isaac-api/blob/v2.18.11/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java#L541-L549). This is not the right place, nor the right way, to order these.

jsharkey13 (changes requested):
This should not be done client side, and the functionality already exists API side to do this.

chrisjpurdy:
> This should not be done client side, and the functionality already exists API side to do this.

Okay!

PR #476, opened 2021-11-24
Enable typing chemical elements when there are no available symbols

Morpheu5:
I'm going to assume that chemistry questions always have available symbols specified. We could be more specific in the condition.

The UX with the input box is slightly less than idea (please see comment in code) but that's the best I could come up with. I'm open to improvement suggestions.


bh412 (approved):
Appears to work, not sure how simple it is on mobile

Morpheu5:
It should be reasonable, probably not great on smaller screens but it
should do.

Il ven 7 gen 2022, 12:14 Ben Hanson ***@***.***> ha scritto:

PR #477, opened 2021-11-26
Feature/topic summary redesign

mlt47:


src/app/services/userContext.ts
```diff
@@ -380,3 +380,35 @@ export function notRelevantMessage(userContext: UseUserContextReturnType): strin
...
+        stagesSet = new Set<Stage>([STAGE.ALL]);
+    } else {
+        stagesSet = new Set<Stage>();
+        audience.forEach(audienceRecord => audienceRecord.stage?.forEach(stage => stagesSet.add(stage)));
+    }
+    // order stages
+    const audienceStages = Array.from(stagesSet).sort(comparatorFromOrderedValues(stagesOrdered));
+    // if you are one of the options - only show that option
+    const stagesFilteredByUserContext = audienceStages.filter(s => userContext.stage === s);
+    let stagesToView = stagesFilteredByUserContext.length > 0 ? stagesFilteredByUserContext : audienceStages;
```
chrisjpurdy, line 396: Just for comprehension, isn't there only ever going to be one stage after filtering by audience context?

chrisjpurdy (approved):
Look absolutely fine, just had one question about how many user context stages you could expect.

src/app/services/userContext.ts
mlt47: Good question, yes there is only one stage in that case.
If the userContext.stage is STAGE.ALL you don't match any and so stagesFilteredByUserContext.length === 0 and we show the whole set of audience stages.
I agree it is a little involved for what it is. If you can find a way to reorganise the logic or the variable names to make it easier to read and comprehend please go for it! :)

PR #478, opened 2021-12-01
Put CS exam boards in alphabetical order everywhere

chrisjpurdy:


chrisjpurdy:
This is failing because I'm using a newer version of npm - if we are moving to npm >=8, looks like the CI bot needs updating too. I'll commit this change without package-lock.json.

Morpheu5 (approved):

PR #483, opened 2021-12-13
Adds stage and difficulty progress donut to Phy

bh412:
Relies on corresponding backend [PR](https://github.com/isaacphysics/isaac-api/pull/401) to add the necessary information to the /progress object

### DIFF
--- a/src/app/components/elements/views/QuestionProgressCharts.tsx
+++ b/src/app/components/elements/views/QuestionProgressCharts.tsx
@@ -1,11 +1,18 @@
 import React, {MutableRefObject, useEffect, useState} from 'react';
 import * as RS from "reactstrap";
-import {LevelAttempts, Levels} from "../../../../IsaacAppTypes";
+import {LevelAttempts} from "../../../../IsaacAppTypes";
 import bb, {Chart} from "billboard.js";
 import tags from "../../../services/tags";
 import Select from "react-select";
 import {ValueType} from "react-select/src/types";
-import {difficultyLabelMap, doughnutColours, specificDoughnutColours, STAGE, TAG_ID} from "../../../services/constants";
+import {
+    difficultyLabelMap,
+    doughnutColours,
+    specificDoughnutColours,
+    STAGE,
+    stageLabelMap,
+    TAG_ID
+} from "../../../services/constants";
 import {SITE, SITE_SUBJECT} from "../../../services/siteConstants";
 import {getFilteredStageOptions} from "../../../services/userContext";
 import {Difficulty} from "../../../../IsaacApiTypes";
@@ -47,12 +54,11 @@ export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
 
     const defaultSearchChoiceTag = tags.getSpecifiedTags(searchTagLevel, tags.allTagIds)[0];
     const [searchChoice, setSearchChoice] = useState(defaultSearchChoiceTag.id);
-    const [stageChoice, setStageChoice] = useState<{value: STAGE; label: string}>({value: STAGE.A_LEVEL, label: "A Level"});
+    const [stageChoice, setStageChoice] = useState<{value: STAGE; label: string}>({value: STAGE.A_LEVEL, label: stageLabelMap[STAGE.A_LEVEL]});
 
     const isAllZero = (arr: (string | number)[][]) => arr.filter((elem) => elem[1] > 0).length == 0;
     const categoryColumns = tags.getSpecifiedTags(topTagLevel, tags.allTagIds).map((tag) => [tag.title, questionsByTag[tag.id] || 0]);
     const topicColumns = tags.getDescendents(searchChoice).map((tag) => [tag.title, questionsByTag[tag.id] || 0]);
-    const levelColumns = [...Array(7)].map((_, i) => [`Level ${i}`, questionsByLevel[i as Levels] || 0]);
     const difficultyColumns = stageChoice && questionsByStageAndDifficulty[stageChoice.value] ?
         Object.keys(questionsByStageAndDifficulty[stageChoice.value])
         .sort(function(a, b){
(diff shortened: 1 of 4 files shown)


src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -37,18 +40,24 @@ const colourPicker = (names: string[]): { [key: string]: string } => {
...
 export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
-    const {subId, questionsByTag, questionsByLevel, flushRef} = props;
+    const {subId, questionsByTag, questionsByLevel, questionsByStageAndDifficulty, flushRef} = props;
 
     const topTagLevel = tags.getTagHierarchy()[0];
     const searchTagLevel = tags.getTagHierarchy()[1];
 
     const defaultSearchChoiceTag = tags.getSpecifiedTags(searchTagLevel, tags.allTagIds)[0];
     const [searchChoice, setSearchChoice] = useState(defaultSearchChoiceTag.id);
+    const [stageChoice, setStageChoice] = useState<{value: STAGE; label: string}>({value: STAGE.A_LEVEL, label: "A Level"});
```
mlt47, line 50: Quite sure we have a STAGE_OPTIONS constant somewhere which could be used rather than redefining the label here.

src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -37,18 +40,24 @@ const colourPicker = (names: string[]): { [key: string]: string } => {
...
 
     const isAllZero = (arr: (string | number)[][]) => arr.filter((elem) => elem[1] > 0).length == 0;
     const categoryColumns = tags.getSpecifiedTags(topTagLevel, tags.allTagIds).map((tag) => [tag.title, questionsByTag[tag.id] || 0]);
     const topicColumns = tags.getDescendents(searchChoice).map((tag) => [tag.title, questionsByTag[tag.id] || 0]);
     const levelColumns = [...Array(7)].map((_, i) => [`Level ${i}`, questionsByLevel[i as Levels] || 0]);
+    const difficultyColumns = stageChoice && questionsByStageAndDifficulty[stageChoice.value] ?
+        Object.keys(questionsByStageAndDifficulty[stageChoice.value])
+        .sort(function(a, b){
+            return Object.keys(difficultyLabelMap).indexOf(a) - Object.keys(difficultyLabelMap).indexOf(b);
```
mlt47, line 59: comparatorFromOrderedValues(stagesOrdered) might be neater here

mlt47 (commented):
A few small comments re. the code. Haven't tested how it looks but the commit messages sound like the colours were approved by the Physics content team.

src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -84,6 +93,23 @@ export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
             }));
         }
 
+        if (SITE_SUBJECT === SITE.PHY && !isAllZero(difficultyColumns || [])) {
```
chrisjpurdy, line 96: Doesn't matter really but `difficultyColumns` will always be an array given how it's defined, so the `|| []` check isn't needed.

src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -101,10 +127,10 @@ export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
         }
     }, [questionsByTag, questionsByLevel, categoryColumns, topicColumns, levelColumns]);
 
-    const noCharts = {[SITE.CS]: 2, [SITE.PHY]: 2}[SITE_SUBJECT];
+    const noCharts = {[SITE.CS]: 2, [SITE.PHY]: 3}[SITE_SUBJECT];
 
     return <RS.Row>
-        {SITE_SUBJECT === SITE.PHY && <RS.Col xl={12/noCharts} md={6} className="mt-4 d-flex flex-column">
+        {SITE_SUBJECT === SITE.PHY && <RS.Col xl={12/noCharts} md={4} className="mt-4 d-flex flex-column">
```
chrisjpurdy, line 133: It looks like `md` is defined as `12/noCharts` as well?

src/app/services/constants.ts
```diff
@@ -1009,7 +1009,13 @@ export const specificDoughnutColours: { [key: string]: string } = {
     [SITE.PHY]: {
         "Physics": "#944cbe",
         "Maths": "#007fa9",
-        "Chemistry": "#e22e25"
+        "Chemistry": "#e22e25",
+        "Practice\u00A0(P1)": "#509e2e",
```
chrisjpurdy, line 1013: I think the above line should be: 
`[difficultyLabelMap.practice_1]: "#509e2e",`
and similarly for each of the colours for the other difficulty labels.
This saves changing labels in more places than necessary in the future, because any changes to `difficultyLabelMap` should affect how they are displayed on these donuts.

chrisjpurdy (changes requested):
I think a few small tweaks are needed for code maintainability but the change is generally good!
Playing around with the donuts to test them made me feel that, if you have selected a topic or stage for which you have no data, the donut should be removed and show "No data". I don't know if this is intended or not, but it didn't make sense to me to select Waves in the topic dropdown and still be shown my progress on Mechanics questions (because I had no data on Waves questions). In any case, that particular change/fix should probably be a separate card.

bh412:
@chrisjpurdy On the 'disappear if nothing for selected' bit, I think I'm inclined to agree, although if doing that we should do the same for the questions by subject donut too!

src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -37,18 +40,24 @@ const colourPicker = (names: string[]): { [key: string]: string } => {
...
     const searchTagLevel = tags.getTagHierarchy()[1];
 
     const defaultSearchChoiceTag = tags.getSpecifiedTags(searchTagLevel, tags.allTagIds)[0];
     const [searchChoice, setSearchChoice] = useState(defaultSearchChoiceTag.id);
+    const [stageChoice, setStageChoice] = useState<{value: STAGE; label: string}>({value: STAGE.A_LEVEL, label: "A Level"});
 
     const isAllZero = (arr: (string | number)[][]) => arr.filter((elem) => elem[1] > 0).length == 0;
     const categoryColumns = tags.getSpecifiedTags(topTagLevel, tags.allTagIds).map((tag) => [tag.title, questionsByTag[tag.id] || 0]);
     const topicColumns = tags.getDescendents(searchChoice).map((tag) => [tag.title, questionsByTag[tag.id] || 0]);
     const levelColumns = [...Array(7)].map((_, i) => [`Level ${i}`, questionsByLevel[i as Levels] || 0]);
```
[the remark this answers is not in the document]
bh412: That's true! I can include that cleanup in this PR too

src/app/components/elements/views/QuestionProgressCharts.tsx
bh412: I can't get this to work properly without some `@ts-ignore`s - it really doesn't like being convinced that the type of `questionsByStageAndDifficulty[stageChoice.value]` is anything other than number, despite it being Difficulty

src/app/components/elements/views/QuestionProgressCharts.tsx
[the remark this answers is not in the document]
bh412: Phy asked for A-Level but maybe I could alter it

PR #484, opened 2021-12-13
Teacher account upgrade prompt consistent styling

chrisjpurdy:
This makes the "I am a teacher" prompt shown on the required account info modal the same as the prompt on the user account page. This improves consistency and makes it clearer than before what exactly it is/means.
image

The only way the user can be upgraded to a teacher account is if they have a student account, so I just check that when deciding to display the prompt. (If it's easy to pull the `RegisteredUserDTO` into the modal then I should really check using that, but otherwise this is fine unless there is any possibility of adding roles in-between student and teacher)


chrisjpurdy:
Added in the latest change:
image

src/app/components/elements/inputs/UserContextAccountInput.tsx
```diff
@@ -180,6 +178,12 @@ export function UserContextAccountInput({
                             }
                         </RS.UncontrolledTooltip>
                     </RS.Label>}
+
+                    {!teacher && <><br/><small>Teacher accounts may choose more than one exam board and stage combination. If you are a teacher, you can {" "}
+                        <Link to={TEACHER_REQUEST_ROUTE} target="_blank">
+                            upgrade your account
+                        </Link>{"."}
+                    </small></>}
```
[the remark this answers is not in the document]
chrisjpurdy: Prompt changed to: "If you are a teacher, upgrade your account to choose more than one exam board and stage." in latest commit, suggested by CS copy edit.

src/app/components/elements/inputs/UserContextAccountInput.tsx
```diff
@@ -180,6 +178,12 @@ export function UserContextAccountInput({
                             }
                         </RS.UncontrolledTooltip>
                     </RS.Label>}
+
+                    {!teacher && <><br/>
+                        <small>
+                            If you are a teacher, <Link to={TEACHER_REQUEST_ROUTE} target="_blank">upgrade your account</Link> to choose more than one exam board and stage.
```
mlt47, line 184: For Physics it should be " choose more than one stage."

chrisjpurdy: Ahhh good point

PR #486, opened 2022-01-06
Feature/assigned test counter

mlt47:
This PR adds a notification counter for assigned tests to the navigation bar.

### DIFF
--- a/src/app/components/navigation/NavigationBar.tsx
+++ b/src/app/components/navigation/NavigationBar.tsx
@@ -19,6 +19,10 @@ import {loadMyAssignments} from "../../state/actions";
 import {filterAssignmentsByStatus} from "../../services/assignments";
 import {selectors} from "../../state/selectors";
 import {SITE, SITE_SUBJECT} from "../../services/siteConstants";
+import {loadQuizAssignedToMe} from "../../state/actions/quizzes";
+import {partitionCompleteAndIncompleteQuizzes} from "../../services/quiz";
+import {isFound} from "../../services/miscUtils";
+import {RenderNothing} from "../elements/RenderNothing";
 
 
 const MenuOpenContext = React.createContext<{menuOpen: boolean; setMenuOpen: React.Dispatch<React.SetStateAction<boolean>>}>({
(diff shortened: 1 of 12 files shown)


src/app/components/navigation/NavigationBar.tsx
```diff
@@ -53,30 +57,40 @@ export const NavigationSection = ({children, title, topLevelLink, to}: Navigatio
...
-                    <span className="badge badge-pill bg-grey ml-2">{assignmentCount}</span>
-                    <span className="sr-only">Incomplete assignments</span>
-                </React.Fragment>
-            }
+            response.assignmentsCount = inProgressRecent.length;
+        }
+        if (state && isFound(state.quizAssignedToMe)) {
+            const [_completedQuizzes, incompleteQuizzes] = partitionCompleteAndIncompleteQuizzes(state.quizAssignedToMe);
+            response.quizzesCount = incompleteQuizzes.length;
         }
```
jsharkey13, line 90: I like that these two if statements are meant to do the same thing, but do them in very different ways with very different method names. Ah, _consistency_ :P

src/app/components/site/cs/NavigationBarCS.tsx
```diff
@@ -1,19 +1,25 @@
...
 
 export const NavigationBarCS = () => {
     const user = useSelector(selectors.user.orNull);
-    const assignmentBadge = useAssignmentBadge();
+    const {assignmentsCount, quizzesCount} = useAssignmentsCount();
 
     return <NavigationBar>
-        <NavigationSection title={<>My Isaac {assignmentBadge}</>}>
-            <LinkItem to="/assignments">My assignments {assignmentBadge}</LinkItem>
+        <NavigationSection title={<>My Isaac {<MenuBadge count={assignmentsCount + quizzesCount} message="Assignments or tests incomplete" />}</>}>
```
jsharkey13, line 18: Should this be "Assignments _and_ tests incomplete"? I'm sure it doesn't matter much, whatever!

Also, would "Incomplete assignments and tests" be a better ordering, so when the screen reader reads e.g. "my isaac three incomplete assignments and tests" it will make more sense?

jsharkey13 (approved):
Looks good! A brief question about phrasing, but not important.

src/app/components/site/cs/NavigationBarCS.tsx
mlt47: I prefer your phrasing. Will change.

PR #491, opened 2022-01-25
Try using react-helmet to adjust metadata on concept pages

jsharkey13:
I'm not really sure how best to test this, but a review would be a good first step.


bh412 (approved):
Implementation matches react-helmet guide. Tested on CS with concept pages, the `og:title` is successfully replaced, there aren't any concept pages with a set summary but it should do the same thing.

PR #492, opened 2022-01-25
Add "coming soon" style option to question filter hexagons

chrisjpurdy:


src/app/components/elements/svg/HierarchyFilter.tsx
```diff
@@ -98,20 +103,27 @@ export function HierarchyFilterHexagonal({tiers, choices, selections, setTierSel
...
+                            {tag.comingSoon && <div className={classNames(subject, "hexagon-coming-soon")}>
+                                {tag.comingSoon}
+                            </div>}
                         </foreignObject>
+
                         <Hexagon
-                            {...hexagon} className="hex none clickable" properties={{clickable: true}} role="button"
-                            tabIndex={0} onClick={selectValue} onKeyPress={ifKeyIsEnter(selectValue)}
+                            {...hexagon} className={classNames("hex none", {"clickable": !isComingSoon})} properties={{clickable: !isComingSoon}} role="button"
+                            tabIndex={isComingSoon ? -1 : 0} onClick={isComingSoon ? noop : selectValue} onKeyPress={isComingSoon ? noop : ifKeyIsEnter(selectValue)}
```
mlt47, line 118: Another option would have been to just set the event handler props to be `undefined` if isComingSoon. JSX gets converted into a call of React.createElement(...) which receives an object containing the props so that would work fine.

src/app/services/tagsPhy.ts
```diff
@@ -118,21 +118,21 @@ export class PhysicsTagService extends AbstractBaseTagService {
         {id: TAG_ID.kinetics, title: "Kinetics", parent: TAG_ID.physical},
         // Analytical
         {id: TAG_ID.electronicSpectroscopy, title: `Electronic Spectro${softHyphen}scopy`, parent: TAG_ID.analytical},
-        {id: TAG_ID.nmrSpectroscopy, title: `NMR Spectro${softHyphen}scopy`, parent: TAG_ID.analytical, comingSoon: "yes"},
+        {id: TAG_ID.nmrSpectroscopy, title: `NMR Spectro${softHyphen}scopy`, parent: TAG_ID.analytical, comingSoon: "Coming soon"},
```
mlt47, line 121: The pattern followed for CS was that the string could be "soon" or "January 2025" so that its value could be used in the label``Coming ${tag.comingSoon}``. The object's property could be renamed to better convey that intention, though. It probably started out life as a boolean value.

mlt47 (approved):
The code here looks fine.
My comments are minor and therefore should not hold back the release. However it would be good for the codebase health to address them after it is mereged.
I have not tested the visual part of this PR (because I'm on the big front-end update branch) but approve of the screenshots which the content team have also seen.

src/app/services/tagsPhy.ts
chrisjpurdy: Yeah I wanted to change it to a `boolean` initially until I realised how it was used by the CS tags. Another name for it would be good!

src/app/components/elements/svg/HierarchyFilter.tsx
chrisjpurdy: So `onClick={!isComingSoon && selectValue}` would work fine then?

PR #495, opened 2022-02-03
Hide deprecated content from search results

mlt47:


jsharkey13 (approved):
This is now a weird mix of chained conditions and if statements, but it looks like it will work as required and the underlying requirements are just as messy as this code!

PR #496, opened 2022-02-03
Add links to the content editor to more page types

jsharkey13:
Also modify the EditContentButton to have the not-mobile check inside itself to require less DOM components and increase consistency.

### DIFF
--- a/src/app/components/elements/quiz/QuizAttemptComponent.tsx
+++ b/src/app/components/elements/quiz/QuizAttemptComponent.tsx
@@ -103,7 +103,7 @@ function QuizHeader({attempt, preview}: QuizAttemptProps) {
     const assignment = attempt.quizAssignment;
     if (preview) {
         return <>
-            <EditContentButton doc={attempt.quiz as IsaacQuizDTO} />
+            <EditContentButton doc={attempt.quiz} />
             <div className="d-flex">
                 <span>You are previewing this test.</span>
                 <Spacer />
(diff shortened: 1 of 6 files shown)


src/app/components/elements/quiz/QuizAttemptComponent.tsx
```diff
@@ -100,11 +102,14 @@ function QuizHeader({attempt, preview}: QuizAttemptProps) {
     const dispatch = useDispatch();
     const assignment = attempt.quizAssignment;
     if (preview) {
-        return <p className="d-flex">
-            <span>You are previewing this test.</span>
-            <Spacer />
-            <RS.Button onClick={() => dispatch(showQuizSettingModal(attempt.quiz as IsaacQuizDTO))}>Set Test</RS.Button>
-        </p>;
+        return <>
+            <EditContentButton doc={attempt.quiz as IsaacQuizDTO} />
```
jsharkey13, line 106: This is the only bit I am worried about (not so much the `p` -> `div` change, but the cast `attempt.quiz as IsaacQuizDTO` which I copied from the existing lines, but seems like it shouldn't be necessary if the attempt had the correct types?

Morpheu5: That's OK. The cast is necessary because `.quiz` is an "optional" typed `IsaacQuizDTO | undefined`. If we are sure that it will always be defined, then we can just leave the forced cast here, but maybe its definedness could be asserted in advance, e.g., in the top condition `preview && isDefined(attempt.quiz)` although that would then require another branch of when it is undefined which is not necessarily a bad thing because at the moment we risk passing `undefined` to something that might not cope well with it.

PR #500, opened 2022-02-08
Hide "Coming soon" hexagon tag on Safari

chrisjpurdy:
This is because it doesn't work and is rendered weirdly. Nicki would like it fixed properly but said this was okay for now, since actually fixing it will be extremely difficult. 
I've tested on Chrome, Edge, Firefox and Safari and only on Safari is it hidden.

### DIFF
--- a/src/scss/phy/filter.scss
+++ b/src/scss/phy/filter.scss
@@ -109,3 +109,23 @@ svg {
     border: solid 1px $phy_chemistry;
   }
 }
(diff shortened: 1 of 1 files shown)


Morpheu5 (approved):

bh412 (approved):
Not sure I like how specific this needs to be to target safari, but it works and is the suggested solution!

PR #503, opened 2022-02-10
Fix shuffle bug by adding dependencies to the useCallback

mlt47:


chrisjpurdy (approved):
Tested it locally and it works as expected! 🎆

PR #504, opened 2022-02-11
Check for correct tag when displaying gameboard owner

mwtrew:


src/test/services/gameboards.test.tsx
```diff
@@ -0,0 +1,59 @@
+import {scrollVerticallyIntoView} from "../../app/services/scrollManager";
+jest.mock("../../app/services/scrollManager")
+
+import { formatBoardOwner } from "../../app/services/gameboards";
```
chrisjpurdy, line 4: This way of handling `history` being `undefined` is fine, but it's a bit arcane. Maybe a comment above the `jest.mock(...)` would be helpful, something like "this erases the definition of `history.listen` in `../../app/services/scrollManager`"?
Also I messed with it a bit and it seems like line 1 isn't needed, `jest.mock(...)` can be underneath line 4, and the tests still pass fine locally for me at least.

chrisjpurdy (approved):
Works well! I had a small suggestion around the usage of `jest.mock` in the test, but it can be merged with or without that change.

src/test/services/gameboards.test.tsx
mwtrew: Thanks, and you're right about the unnecessary import. Good to know about the mocking-everything-else approach but I prefer the simplicity of the single mock (with the new comment that explains what's going on).

PR #505, opened 2022-02-15
Log in/Sign up prompt after anonymous question attempt

chrisjpurdy:
image
image

The modal redirects back to the question that was being attempted after authentication succeeds, and is easily dismissible if the user doesn't want to log in. 
For the redirect functionality, I had to change how `updateCurrentUser` and `handleProviderCallback` worked, to make sure `AFTER_AUTH_PATH` is *always followed if specified* - I don't see this causing any issues, but I might have missed something.

For code cleanliness and to stop duplication etc. I broke up the `LogIn` component into small chunks. I also changed the "Log in with Google" button to be styled in the same way as other buttons on the site, and to fit geometrically on the modal.

Before:
image
After:
image
[Google sign in button brand guidelines for reference](https://developers.google.com/identity/branding-guidelines)


chrisjpurdy:
Currently the modal shows once every 2 hours or when local storage is cleared (the time of the modal being shown last is stored in `persistance`. It might be preferable to show it just once a session, but I'm not sure how?

Morpheu5 (approved):
Aside from me hating the modal because it's huge and it breaks the flow, this LGTM so could be merged. The timeout of two hours should help, although I would go for something closer to however long it's going to be until we log them out. OTOH, IIRC we discard any attempts done 30 minutes prior to a login, so there's going to be some balancing to do.

mwtrew:
Confirmed your test results, looks good.

mwtrew (approved):

PR #508, opened 2022-02-17
Add assembly as a possible programming language

chrisjpurdy:


bh412 (commented):
CS requested this not to be highlighted, which hasn't been defined so fine, but does this actually play nicely with the fact that an `assembly` highlight option hasn't been added to `registerLanguages()` in `services/highlightJs.ts`? Perhaps it may need ```hljs.registerLanguage('assembly', isaacPseudocodeHighlightDefinition);```

bh412 (approved):

PR #511, opened 2022-02-22
Expanding wide CS content horizontally

chrisjpurdy:
This is an idea for expanding content (specifically tables and code snippets) on large screen sizes (exclusively on the CS site).
It basically just copies the functionality of parsons questions layout over to other types of wide content. 

This also adds subtle shadows to the edge of scrollable content, implying that there is more to see.

Only two types of content are able to be marked as expandable with a property in the content model: tabs and code snippets. Tables need `class="expandable"` specified when written in content if they are to be expanded, as we can't yet associate properties in the content model with them.

There are no editor changes yet, and for content to be expandable the content teams will need to manually change the JSON properties.

Test page for all of these changes can be found on `/concepts/c0107d3b-9a96-4f4f-8949-76d1cd2771f8`.


mwtrew:
Test page for this: concepts/c0107d3b-9a96-4f4f-8949-76d1cd2771f8

src/app/components/elements/TrustedHtml.tsx
```diff
@@ -1,50 +1,118 @@
...
 }
 
-export const TrustedHtml = ({html, span, className}: {html: string; span?: boolean; className?: string}) => {
+// A portal component to manage table elements from inside the React DOM
+const Table = ({id, html, classes, rootElement}: TableData & {rootElement: RefObject<HTMLElement>}) => {
+    const parentElement = rootElement.current?.querySelector(`#table-${id}`);
+
+    const scrollRef = useRef<HTMLDivElement>(null);
+    const expandRef = useRef<HTMLDivElement>(null);
+    const {expandButton, innerClasses, outerClasses} = useExpandContent(classes.includes("expandable"), expandRef, "overflow-auto mb-4");
```
mwtrew, line 83: If we can think of a good way to do it, it would be preferable to check the `expandable` DTO property for the block rather than checking for this class in the HTML. This way, the process of enabling it is inconsistent with the other two expandable types, and ETL won't be able to check for the potentially-expandable-parent error case (unless it were to read into the HTML content or something, which I don't like the idea of). 

However, I appreciate that checking the `expandable` property from here isn't straightforward (as was probably obvious from my attempt at doing it that way).

chrisjpurdy: Unless we make tables their own content block, then this isn't really possible without introducing some redundancy (and potentially a bit of confusion?). You'd need to mark the entire content section containing the table as expandable to be consistent, but it might contain other text which wouldn't be affected by it at all - it isn't clear that only the table is being affected by the tickbox IMO.

mwtrew: I agree, the ambiguity there isn't ideal either. Both of these problems could be solved if tables were a distinct content type, perhaps that's something to investigate later (I suspect that will be difficult).

PR #513, opened 2022-02-25
Adds hash search for hardcoded support links

bh412:


jsharkey13 (approved):
Not tested (we should also check that shortcuts and links without a hash work okay as well as the new ones), but looks okay to me!

PR #516, opened 2022-03-04
Focus last focused element on modal close

chrisjpurdy:
This fixes an accessibility issue where a modal popping up resets the focused element. It adds Redux middleware that records the last focused element before a modal opens, and restores it on modal close (or resets focus to the top of the main content if no element exists).

When this is tested, the team that does might want to try navigating the teacher gameboard builder with a screen-reader to make sure it is as accessible as needed.

### DIFF
--- a/src/app/state/modalFocusMiddleware.ts
+++ b/src/app/state/modalFocusMiddleware.ts
@@ -1,24 +1,27 @@
 import {Dispatch, Middleware, MiddlewareAPI} from "redux";
 import {ACTION_TYPE} from "../services/constants";
 
-let lastFocused: Element | null = null;
+const lastFocusedStack: Element[] = [];
 
 // When a modal is closed, this tries to focus the last element that was focused before it opened. If no element exists,
 // we try to focus the page title. This is for accessibility, mostly to help when navigating the site with a screen-reader.
 export const modalFocusMiddleware: Middleware = (api: MiddlewareAPI) => (next: Dispatch) => action => {
+    const state = api.getState();
+
     switch (action.type) {
         case ACTION_TYPE.ACTIVE_MODAL_OPEN:
             // Before modal open, record the element that is currently focused
-            lastFocused = document.activeElement;
+            if (document.activeElement) lastFocusedStack.push(document.activeElement);
             break;
-        case ACTION_TYPE.ACTIVE_MODAL_CLOSE:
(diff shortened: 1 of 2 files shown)


src/app/state/modalFocusMiddleware.ts
```diff
@@ -0,0 +1,25 @@
+import {Dispatch, Middleware, MiddlewareAPI} from "redux";
+import {ACTION_TYPE} from "../services/constants";
+
+let lastFocused: Element | null = null;
```
jsharkey13, line 4: Do we ever open modals on top of modals? This single item won't cope well with that. I don't _think_ we do (maybe we can't?), but just thought I'd check!

src/app/state/modalFocusMiddleware.ts
```diff
@@ -0,0 +1,25 @@
...
+            // Before modal open, record the element that is currently focused
+            lastFocused = document.activeElement;
+            break;
+        case ACTION_TYPE.ACTIVE_MODAL_CLOSE:
+            // Before modal close, try to focus the last focused element, otherwise focus the main heading (if it exists)
+            if (lastFocused !== null && typeof (lastFocused as HTMLElement)?.focus === "function") {
+                (lastFocused as HTMLElement)?.focus();
+                lastFocused = null;
+            } else {
+                document.getElementById("main")?.focus();
```
jsharkey13, line 20: Often the default is not actually the `#main` element; we have a selector just for this:
`useSelector(selectors.mainContentId.orDefault);`

See e.g. [the "Skip to main content" link](https://github.com/isaacphysics/isaac-react-app/blob/master/src/app/components/site/cs/HeaderCS.tsx#L25) for an example.

jsharkey13 (commented):
I can't really comment on the suitability of middleware for this, but two comments on the method itself!

jsharkey13 (approved):
I should test this before I merge it.

Morpheu5:
> I should test this before I merge it.

I tested it and it looks OK to me. Merging.

mlt47:
@Morpheu5, from a discussion at a stand up there was a problem with this PR when there was a modal stacked on top of another modal as we do in the teacher gameboard builder.

Morpheu5:
> @Morpheu5, from a discussion at a stand up there was a problem with this PR when there was a modal stacked on top of another modal as we do in the teacher gameboard builder.

Hah. Reverted. I'll figure out how to reopen this one now.

PR #519, opened 2022-03-08
Feature/address inconsistent input validation error presentation during user create update

mwtrew:


src/app/components/elements/panels/UserDetails.tsx
```diff
@@ -64,22 +67,30 @@ export const UserDetails = (props: UserDetailsProps) => {
                 <FormGroup>
                     <Label htmlFor="first-name-input" className="form-required">First name</Label>
                     <Input
-                        id="first-name-input" type="text" name="givenName" maxLength={255}
```
chrisjpurdy, line 67: This shouldn't hold back merging, just a thought: I feel like it makes sense to keep the `maxLength={255}` property. Although it's tested for using `validateName`, stopping the user from inputting more than 255 characters in the input element at all seems nicer somehow?

src/app/components/pages/Registration.tsx
```diff
@@ -250,14 +266,20 @@ export const Registration = withRouter(({location}:  RouteComponentProps<{}, {},
                     {/* Form Error */}
                     <Row>
                         <Col>
+                            {attemptedSignUp && (!passwordIsValid || !emailIsValid
+                                || !validateName(registrationUser.givenName)
+                                || !validateName(registrationUser.familyName)) &&
```
chrisjpurdy, line 271: Since you use `validateName(registrationUser.givenName)` a few times, you could copy how `emailIsValid` works and define `givenNameIsValid = validateName(registrationUser.givenName)` on line 68, and do the same refactor for `familyNameIsValid`. This makes it a bit nicer to read by removing some noisy syntax.

chrisjpurdy (approved):
Looks good, just had a couple of minor comments on style!

src/app/components/elements/panels/UserDetails.tsx
mwtrew: While I don't have a strong feeling about this as it shouldn't affect any users who aren't just probing our input validation, I don't like how maxLength truncates the input without any feedback which is why I removed it.

PR #521, opened 2022-03-10
Link the maths text entry button help tooltip to the /solving_problems#symbolic_text page

Morpheu5:
From [this](https://trello.com/c/9ZIbsIK5/2394-make-the-equation-text-entry-button-clickable-for-further-help) content request.


jsharkey13:
I did not attempt to do this, and indeed the code is not live because Andrea reverted it on the branch I merged? I think they were chained together and GitHub is now very confused :/

PR #523, opened 2022-03-10
Add meta descriptions to static pages

jsharkey13:
This will need a rethink if/when physics want a customised description too.


Morpheu5 (approved):

mlt47 (approved):
All looks straightforward enough. I would almost be tempted to create a SeoDescription component and pass it a single description.

PR #524, opened 2022-03-11
Show a single accordion section even if it's irrelevant to the user

Morpheu5:
To do:

- [x] Make sure that it actually does what I think it does
- [x] Check that it doesn't mess with any regular content
- [x] Check the alert message with the content team


src/app/components/content/IsaacAccordion.tsx
```diff
@@ -36,7 +37,10 @@ export const IsaacAccordion = ({doc}: {doc: ContentDTO}) => {
     const defaultDisplay = isFound(page) && page.type === DOCUMENT_TYPE.CONCEPT ? defaultConceptDisplay : defaultQuestionDisplay;
     const accordionDisplay = mergeDisplayOptions(defaultDisplay, doc.display);
 
+    const hashAnchor = location.hash.includes('#') ? location.hash.slice(1) : 'undefined-string-will-never-match';
```
mlt47, line 40: Probably worth using the `const {hash} = useLocation();` so that it updates on following soft links (although it is less important for this particular feature, a future feature would assume it would update).

mlt47, line 40: Isn't `null` better than 'undefined-string-will-never-match' here? You later check if hashAnchor is equal to the value, if there is an accidental typo in either it will never reach that case and the compiler won't complain because the standard type is string.

src/app/components/content/IsaacAccordion.tsx
```diff
@@ -71,6 +75,15 @@ export const IsaacAccordion = ({doc}: {doc: ContentDTO}) => {
                 return section;
             })
 
+            // If we followed a direct link to a section, we want to show it regardless
+            // of any other settings. We also want to show a message somewhere on the page.
+            .map(section => {
+                if ((section.id || "").includes(hashAnchor)) {
```
mlt47, line 81: Using `includes` here rather than equality would do the incorrect thing if there were accordion sections `hash`, `hash-map`, `hash-set`, `rehash` and the user followed the link for `#hash`.

mlt47 (changes requested):

src/app/components/content/IsaacAccordion.tsx
Morpheu5: I assume you mean `useLocation()` from `react-router-dom`, but I'm uncomfortable using it because the types are so horrible – `useLocation()` returns `any`, and therefore `useLocation().hash` is also any, how am I meant to know if I can trust it to always be a string, or whether it's going to be `undefined` or `null`?

src/app/components/content/IsaacAccordion.tsx
Morpheu5: This *can't* be an equality though because a `section.id` might include a page id, IIRC. And in fact I'm testing with /concepts/prog_concepts_instructions#seq-instructions and it doesn't work if I check of equality: the `section.id` is indeed `prog_concepts_instructions|seq-instructions`.

Morpheu5: Uh, I forgot I ended up using `useLocation()` in the end, so marking this as resolved.

mlt47: Ah right that is a good point about equality. `includes` is still "wrong" though, we should `split("|")` and grab the section's ID. It is also done in `Accordion.tsx` so we can pull the shared logic to a function in a service and reuse that.

mlt47 (changes requested):

mlt47 (approved):

PR #525, opened 2022-03-11
New editor preview: renderer for CS and physics that can be used with the new editor

robinmessage:


mlt47 (approved):
That's neat, I like the solution. I tested it by just sending messages from the console rather than integrating it with the new editor but it's clear how that would connect up.

PR #526, opened 2022-03-11
Add rel=canonical meta tags to pages with many URL parameters

jsharkey13:
When a search engine indexes these pages, we only care about the full versions of the pages (I think!) and not the exam board or stage specific versions. This strips out all URL parameters and hashes and leaves just the bare URL as "canonical".
The new element is designed to work on any page so it can be reused.

I can't see why the component correctly rerenders when the page changes (it feels like it needs a prop or something so React knows to re-render it?!), but it seems to work for the places it is currently in!

---
The aim is to reduce the number of duplicate search results caused by the explosion of possible combinations of URL parameters on each page, and we can always roll it back if it doesn't work.

### DIFF
--- a/src/app/components/navigation/CanonicalHrefElement.tsx
+++ b/src/app/components/navigation/CanonicalHrefElement.tsx
@@ -1,0 +1,14 @@
+import {history} from "../../services/history";
+import {Helmet} from "react-helmet";
+import React from "react";
+
+export const CanonicalHrefElement = () => {
+    let canonicalPath = "";
+    if (history.location.pathname !== "/") {
+        canonicalPath = history.location.pathname;
+    }
+    const canonicalHref = `${window.location.origin}${canonicalPath}`;
+    return <Helmet>
+        <link rel="canonical" href={canonicalHref}/>
+    </Helmet>
+};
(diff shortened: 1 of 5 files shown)


Morpheu5 (approved):

src/app/components/navigation/CanonicalHrefElement.tsx
```diff
@@ -0,0 +1,14 @@
+import {history} from "../../services/history";
+import {Helmet} from "react-helmet";
+import React from "react";
+
+export const CanonicalHrefElement = () => {
+    let canonicalPath = "";
+    if (history.location.pathname !== "/") {
+        canonicalPath = history.location.pathname;
+    }
```
mlt47, line 9: You're right that this won't trigger a re-render in itself. The place that it might fail would be if you went from one question page to another through gameboard/topic navigation. The easiest way to get around this would be to use the [useLocation()](https://v5.reactrouter.com/web/api/Hooks/uselocation) hook to get the pathname and so it will handle the updating on path change.

mlt47 (changes requested):

mlt47 (approved):

PR #529, opened 2022-03-16
Manage Tests / Assignment Progress (tests): student performance overflows window, doesn't scroll

chrisjpurdy:
It does now! Plus the overall mark column is now sticky to the right of the table (which seemed to be what the old CSS was trying to do anyway)

A lot of the diff is cleaning up old code to make the linter happy (turning `let` into `const`, that kind of thing)


Morpheu5 (approved):

src/app/components/elements/quiz/QuizProgressCommon.tsx
```diff
@@ -160,34 +160,35 @@ export function ResultRow({pageSettings, row, assignment}: ResultRowProps) {
                     </td>
                 }).flat()
             })}
-            <td className="total-column">
+            <th className="total-column">
```
jsharkey13, line 163: These shouldn't be header cells; it makes no sense!

PR #533, opened 2022-03-21
Focus last focused element on modal close – restored after wrong merge

Morpheu5:
Sorry Chris!


chrisjpurdy:
Well, looks like bootstrap [actually does this itself](https://github.com/twbs/bootstrap/pull/13627), and it works with multiple modals. Oops!

PR #534, opened 2022-03-22
Fix bug showing the waiting list message inappropriately

Morpheu5:
Account for other states in which a user's booking can be when deciding if they should see the waiting list message.


chrisjpurdy (approved):
Change works, and makes complete sense statically 👍

PR #536, opened 2022-03-25
katex-a11y.js chokes on `\vphantom{X}` when dealing with chemistry

Morpheu5:
This could be a 'solution' but will need review.


jsharkey13:
I think this is at least better than nothing; the `hphantom` and `vphantom` are indeed about visual layout and so should add nothing to the screenreader text. But testing on some real chemistry equations (the source of this issue) found that arrows were also an issue too, so I copied in the common arrow names from the KaTeX source too.
I think this is good to go.

PR #537, opened 2022-03-25
Test question numbering

chrisjpurdy:
image

Styling might need a change, but the code is done IMO.

This PR also contains two other commits with some refactoring and missing additions from other features. The commit comments explain these.

### DIFF
--- a/src/app/components/elements/TrustedHtml.tsx
+++ b/src/app/components/elements/TrustedHtml.tsx
@@ -12,7 +12,7 @@ function manipulateHtml(html: string) {
     // This can't be quick but it is more robust than writing regular expressions...
     htmlDom.innerHTML = html;
 
-    // Table manipulation#
+    // Table manipulation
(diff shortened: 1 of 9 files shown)


src/app/components/elements/TrustedHtml.tsx
```diff
@@ -12,7 +12,7 @@ function manipulateHtml(html: string) {
     // This can't be quick but it is more robust than writing regular expressions...
     htmlDom.innerHTML = html;
 
-    // Table manipulation
+    // Table manipulation#
```
bh412, line 15: I think this just slipped in?

chrisjpurdy, line 15:
```suggestion
    // Table manipulation
```

chrisjpurdy: Yep - just removed it in the latest commit

PR #539, opened 2022-03-31
Fallback to `doc.id` if `questionPart.id` doesn't exist on cloze question quiz previews

chrisjpurdy:
Prior to the fix (on `/test/preview/quiz_alevel_boolean_01/page/1?examBoard=all&stage=all`):
image

This fix works by falling back to the `doc.id` if `questionPart.id` is `undefined`. Whether this is the right move or whether the question part id should be defined is something I'm not sure about.

### DIFF
--- a/src/app/components/content/IsaacClozeQuestion.tsx
+++ b/src/app/components/content/IsaacClozeQuestion.tsx
@@ -99,7 +99,7 @@ export function IsaacClozeQuestion({doc, questionId, readonly}: {doc: IsaacCloze
     const pageQuestions = useSelector(selectors.questions.getQuestions);
     const questionPart = selectQuestionPart(pageQuestions, questionId);
     const currentAttempt = questionPart?.currentAttempt as (ItemChoiceDTO | undefined);
-    const cssFriendlyQuestionPartId = questionPart?.id?.replace(/\|/g, '-') ?? ""; // Maybe we should clean up IDs more?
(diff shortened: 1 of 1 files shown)


chrisjpurdy:
@mlt47 this seems to happen because the following `useSelector(selectors.questions.getQuestions);` returns `undefined` if a quiz is being previewed. Would you know why this happens?

src/app/components/content/IsaacClozeQuestion.tsx
```diff
@@ -99,7 +99,7 @@ export function IsaacClozeQuestion({doc, questionId, readonly}: {doc: IsaacCloze
     const pageQuestions = useSelector(selectors.questions.getQuestions);
     const questionPart = selectQuestionPart(pageQuestions, questionId);
     const currentAttempt = questionPart?.currentAttempt as (ItemChoiceDTO | undefined);
-    const cssFriendlyQuestionPartId = questionPart?.id?.replace(/\|/g, '-') ?? ""; // Maybe we should clean up IDs more?
+    const cssFriendlyQuestionPartId = (questionPart?.id || doc.id)?.replace(/\|/g, '-') ?? ""; // Maybe we should clean up IDs more?
```
mlt47, line 102: `questoinId` is what we use to filter out the `questionPart` from the registered questions so it's safe to just use `questionId` directly. The reason this did not work on Tests/Quizzes was that Tests/Quizzes do not register their question parts. 
```suggestion
    const cssFriendlyQuestionPartId = questionId?.replace(/\|/g, '-') ?? ""; // Maybe we should clean up IDs more?
```

PR #541, opened 2022-04-05
Make all question types use new `useCurrentQuestionAttempt` hook

chrisjpurdy:
This removes a fair amount of repeated code (as well as any mention of `connect` in question components), and the refactoring of types etc. needed to implement this meant that TS can finally infer the type of `QUESTION_TYPES` (woo!).

As a required side-effect, this change also turns `IsaacParsonsQuestion` from a class component into a functional component.
I tested the refactored Parson's question component on [this question](https://isaaccomputerscience/questions/prog_string_07?examBoard=all&stage=a_level) and it seems to work identically.


Morpheu5 (approved):
Thanks for this one. Class-style components are probably my worst sin here.

mlt47 (changes requested):
Nice 👍. The types improvement is neat and it's great to have moved away from the remaining class based components.

The only bit I don't really like is that the hook wraps the action creator in the dispatch implicitly. I appreciate that the hook then has a return signature which is more like a `useState()` hook, but implicitly dispatching in this case would be different to what we do for all other actions creators in the codebase. I think it's OK for things stored in redux to look slightly different to things which are stored in React state, seeing as that's the way the react-redux library developers chose to go.

A slightly contrived example of it causing a problem would be, if a dev was to copy and paste the submission function of an existing question type into a new question type without (or before) using the `useCurrentQuestionAttempt` hook. `setCurrentAttempt(..., ...)` would get imported automatically by the IDE from actions.tsx but calling it would not change redux state, which would be surprising seeing as the exact same method worked in the c&p source.

mlt47 (approved):

mlt47 (changes requested):
If you compare the style of the empty tray in http://localhost:8003/questions/_regression_test_#acc_parsons_q with https://staging.isaaccomputerscience.org/questions/_regression_test_#acc_parsons_q something seems to have gone missing during the conversion from class component to the functional component.

chrisjpurdy:
> If you compare the style of the empty tray in http://localhost:8003/questions/_regression_test_#acc_parsons_q with https://staging.isaaccomputerscience.org/questions/_regression_test_#acc_parsons_q something seems to have gone missing during the conversion from class component to the functional component.

Fixed! It was just some faulty CSS class logic, that I've now copied properly from the old version

PR #542, opened 2022-04-07
Allow gameboard filter to populate exam board from URL properly

chrisjpurdy:
This doesn't necessarily address the possibility of the gameboard not showing up when the URL is loaded, but I haven't been able to reproduce that and can't see a reason for it happening if it does - I just need to ask CS content for examples I think.


mlt47 (approved):
Looks good and works for me.

PR #546, opened 2022-04-21
Don't express subject and field in CS question finder URL

chrisjpurdy:
After this, a URL such as `/gameboards/new?difficulties=practice_1&topics=the_internet,communication` will correctly populate the question finder. Examboard and stage information was optional anyway.


jsharkey13 (approved):
This seems to work as-expected for me locally!

PR #551, opened 2022-04-29
Fix KaTeX fonts and Webpack bundling

jsharkey13:
We were not bundling the KaTeX fonts correctly, because the `css-loader` being used to import them was ignoring all URL rewriting so sometimes they would not load correctly in a production build.

The problem is that we use the `create-react-app` folder `public` for all of our assets, both those Webpack wants to know about and those that really are static and not part of the bundle. So instead of using Webpack as it intended, we just bluntly copy the `public/assets` folder to the `/assets` path and we turned off URL rewriting so this would work. But the turning off affected things beyond our special assets folder. So this fix, which is a bit hacky, is just to treat the `/assets` URL as a special case and rewrite anything else.

(The only other changes are to give bundled images and fonts sensible names, noting that `[ext]` contains the leading dot character and so we didn't need it in our templated names).

### DIFF
--- a/config/webpack.config.common.js
+++ b/config/webpack.config.common.js
@@ -17,7 +17,7 @@ module.exports = (isProd) => {
         stats: {
             errorDetails: true
         },
-        
+
         mode: isProd ? "production" : "development",
 
         devServer: {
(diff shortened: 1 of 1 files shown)


chrisjpurdy (approved):
I've got it building and serving locally and can confirm that chemical eqns, matrices and the rest render perfectly from what I've seen. Looks fixed! :thumbsup:

PR #552, opened 2022-04-29
Update dependencies, April 2022

Morpheu5:
Updates:

- critical
  - minimist: ^1.2.6
- high
  - async: ^2.6.4
- moderate
  - highlight.js: ^11.5.1
    - This one is a bit odd as @types/remarkable depended on an old, vulnerable version, so the risk was minimal and entirely local / build-time.


chrisjpurdy (approved):
LGTM! The forced resolutions only push `minimist` and `async` up a couple of minor versions in each dependant package, and the `highlight.js` upgrade doesn't affect anything at runtime. Site builds and runs absolutely fine as far as I can tell 👍

bh412 (approved):
I've tested the areas that the updated packages touch as much as I can and it functions as expected. Running `yarn audit` on the branch now indicates 0 vulnerabilities.

PR #554, opened 2022-05-04
Add a particles tab back to Inequality's menu

Morpheu5:
This also hides the particles tab when available symbols are specified and changes the 'Elements' tab label according to whether it only contains elements or also contains particles. Elements are kept in the label even if the menu only contains particles to avoid breaking consistency

### DIFF
--- a/src/app/components/elements/modals/InequalityModal.tsx
+++ b/src/app/components/elements/modals/InequalityModal.tsx
@@ -73,6 +73,7 @@ interface InequalityModalState {
         mathsLogFunctions: MenuItem[];
         mathsDerivatives: MenuItem[];
         chemicalElements: MenuItem[];
+        chemicalParticles: MenuItem[];
         // The following is for the pseudo-text-entry menu on /equality
         parsedChemicalElements: MenuItem[];
         chemicalStates: MenuItem[];
@@ -212,6 +213,7 @@ class InequalityModalComponent extends React.Component<InequalityModalProps> {
                 letters: [],
                 otherFunctions: [],
                 chemicalElements: [],
+                chemicalParticles: [],
                 parsedChemicalElements: [],
                 chemicalStates: [],
                 chemicalOperations: [],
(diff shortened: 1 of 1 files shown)


src/app/components/elements/modals/InequalityModal.tsx
```diff
@@ -1244,10 +1252,16 @@ class InequalityModalComponent extends React.Component<InequalityModalProps> {
                     {/* Chemistry below */}
                     {this.props.editorMode === 'chemistry' &&
                     <li className={this.state.activeMenu === 'elements' ? 'active' : 'inactive'}
-                        dangerouslySetInnerHTML={{ __html: this._tabTriangle + katex.renderToString('\\text{H He Li}') }}
+                        dangerouslySetInnerHTML={{ __html: this._tabTriangle + katex.renderToString(isDefined(this.props.availableSymbols) && this.props.availableSymbols.length > 0 && this.state.menuItems.chemicalElements.map(i => i.type).includes('Particle') ? '\\text{He Li}\\ \\alpha' : '\\text{H He Li}') }}
                         onClick={() => this.onMenuTabClick('elements')}
                         onKeyUp={() => this.onMenuTabClick('elements')}
                     />}
+                    {this.props.editorMode === 'chemistry' && this.state.menuItems.chemicalParticles.length > 0 &&
```
chrisjpurdy, line 1259: You mentioned in the PR description that this menu tab is hidden if any available symbols are defined, but I can't tell if there is any logic that does that?

chrisjpurdy (approved):
I tested it out on a few chemistry questions and it looks like it works, and nothing seems to have broken. Just one comment because I was confused that the logic in the code didn't seem to line up with what I saw when testing

src/app/components/elements/modals/InequalityModal.tsx
chrisjpurdy: Is it that if available symbols is populated, content aren't supposed to define any `chemicalParticles`?

Morpheu5: Yes, more or less. If content defines available symbols, they can define elements AND particles at the same time, so `chemicalParticles` should not get added to the menu tabs.

PR #558, opened 2022-05-06
Set assignments to multiple groups at once

chrisjpurdy:
This changes the UI element that lets you select groups to assign gameboards to from a single-input `input` element to a multi-input React Select component. The input only allows a user to select multiple groups if they are staff (as per the original feature request).

Some thought had to go into how errors from the API are shown, considering that some of the assignments can fail while others succeed - this part probably needs the most attention during review.

To test this locally, you'll need to be on the back-end branch too: https://github.com/isaacphysics/isaac-api/pull/443


src/test/state/reducers.test.tsx
```diff
@@ -475,4 +475,5 @@ describe("boards reducer", () => {
         });
     });
 
+    // TODO extend to multiple assignees
```
Morpheu5, line 478: Is this still to be done?

src/app/services/api.ts
```diff
@@ -437,8 +439,8 @@ export const api = {
         unassign: (board: ApiTypes.GameboardDTO, group: ApiTypes.UserGroupDTO) => {
             return endpoint.delete(`/assignments/assign/${board.id}/${group.id}`);
         },
-        assign: (board: ApiTypes.GameboardDTO, groupId: number, dueDate?: number, assignmentNotes?: string) => {
-            return endpoint.post(`/assignments/assign`, {dueDate, gameboardId: board.id, groupId, notes: assignmentNotes})
+        assign: (assignments: AssignmentDTO[]): AxiosPromise<AssignmentSettingResponseDTO> => {
+            return endpoint.post(`/assignments/assign_bulk`, assignments);
```
Morpheu5, line 443: See deprecation comment on the other PR.

chrisjpurdy: I'm not sure what you mean? Should I add the single assignment endpoint back to the code and mark it as deprecated, because currently it isn't used anywhere.

src/test/state/reducers.test.tsx
chrisjpurdy: I've added a test for this in the last commit

src/app/services/api.ts
Morpheu5: I only meant that this change was related to that and that the `/assign` endpoint should be marked as deprecated (which you have IIRC). Also, didn't @jsharkey13 suggest we did `/assign/bulk` or something similar instead of `/assign_bulk`?

PR #562, opened 2022-05-09
Don't render superseded message if `supersededBy` is an empty string (or any falsy value)

chrisjpurdy:


jsharkey13 (approved):
This looks okay; what specific case is this intended to fix?

jsharkey13:
I found the card about it! 🎉

PR #566, opened 2022-05-10
Don't render `value` content blocks inside bootstrap rows and cols

chrisjpurdy:
There is no reason to do this as far as I can tell, and this fixes the floating figure messing with paragraph alignment issue. Minimising any collisions of negative margins + flexboxes + floating elements seems like a good idea.


Morpheu5 (approved):
This makes a lot of sense. I'm afraid we've been doing unnecessary nestings like this in other parts of the app.

mlt47 (approved):
You're right that the Row and the Col cancel out.
It is difficult to judge the blast radius of a css change like this but, from inspecting the regression test pages and concept pages with horizontal and floated images, everything seems to be consistent with the old site, so happy for this to go in.

chrisjpurdy:
> Also, not caused by this PR as it is currently on staging but I can see a `Loading glossary...` message in the hint tabs even when the glossary terms are visible!
This is caused by that glossary terms causing crashes bug fix, I can just mark `Loading glossary...` as no print - I wonder how this looked before?

PR #569, opened 2022-05-18
Hotfix/clear after auth path consistently

mlt47:


src/app/state/middleware/userConsistencyChecker.tsx
```diff
@@ -61,14 +61,17 @@ export const userConsistencyCheckerMiddleware: Middleware = (api: MiddlewareAPI)
...
         case ACTION_TYPE.USER_LOG_IN_RESPONSE_SUCCESS:
         case ACTION_TYPE.USER_UPDATE_RESPONSE_SUCCESS:
             setCurrentUser(action.user, api);
             break;
         case ACTION_TYPE.USER_CONSISTENCY_ERROR:
-            clearCurrentUser();
+            clearCurrentUserAndRedirectHome();
+            break;
+        case ACTION_TYPE.USER_UPDATE_RESPONSE_FAILURE:
+            setUserId(undefined);
```
jsharkey13, line 74: Why only set the ID to undefined here? Is it because this is the "expected" path for anonymous users? `FAILURE` makes it seem like we got a 5xx and not a 401 Unauthorized; would we use the same action for both?

At any rate, a brief comment here might help in future?

jsharkey13 (approved):
Seems plausible to me!

jsharkey13 (approved):
This makes more sense now! 🥳

PR #573, opened 2022-05-20
New "reorder" question type

chrisjpurdy:
Essentially parsons questions but without indentation.

This PR also includes a couple of small fixes to parsons questions.


jsharkey13:
How is this distinct from a Parsons's question with `disableIndentation=true`?

chrisjpurdy:
> How is this distinct from a Parsons's question with `disableIndentation=true`?

@jsharkey13 It's not really - its just a separation of concerns. It might be that either type of question ends up evolving in a different direction, since one is programming-oriented and the other is unopinionated.
If you want to save on some class definitions, maybe it's possible to just have `IsaacReorderQuestion` extend `IsaacParsonsQuestion` but set `disableIndentation=true` in the constructor?

src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -116,25 +116,25 @@ export const IsaacParsonsQuestion = ({doc, questionId, readonly} : IsaacQuestion
...
             moveItem(items, result.source.index, items, result.destination.index, 0);
             setAvailableItems(items);
-        } else if (result.source.droppableId === 'availableItems' && result.destination.droppableId === 'answerItems' && currentAttempt) {
+        } else if (result.source.droppableId === 'availableItems' && result.destination.droppableId === 'answerItems') {
             // Move from availableItems to currentAttempt
             const srcItems = [...availableItems];
             const dstItems = [...(currentAttempt?.items || [])];
             moveItem(srcItems, result.source.index, dstItems, result.destination.index, currentIndent || 0);
-            dispatchSetCurrentAttempt({...currentAttempt, ...{ items: dstItems }});
+            dispatchSetCurrentAttempt({type: "parsonsChoice", items: dstItems});
```
jsharkey13, line 130: Why does this spell out the type explicitly and no longer use the `currentAttempt`? Could you add a comment to explain it? Because above and below _do_ use the current attempt and the existing comment for this case doesn't really explain what this bit does.

src/app/components/content/IsaacReorderQuestion.tsx
```diff
@@ -0,0 +1,169 @@
...
+            // Reorder availableItems
+            const items = [...availableItems];
+            moveItem(items, result.source.index, items, result.destination.index);
+            setAvailableItems(items);
+        } else if (result.source.droppableId === 'availableItems' && result.destination.droppableId === 'answerItems') {
+            // Move from availableItems to currentAttempt
+            const srcItems = [...availableItems];
+            const dstItems = [...(currentAttempt?.items || [])];
+            moveItem(srcItems, result.source.index, dstItems, result.destination.index);
+            dispatchSetCurrentAttempt({type: "itemChoice", items: dstItems});
```
jsharkey13, line 70: Maybe any comment from the Parsons question as noted above could apply here too; the comment here is just as unhelpful.

jsharkey13 (approved):
Maybe improve the comments for clarity, since you clearly fixed something that you consider a bug but it's not clear from what's here why it was wrong. Otherwise looks okay!

PR #575, opened 2022-05-20
My Progress bars out of total questions in each Isaac book

chrisjpurdy:
This makes each Isaac books progress bar show the percentage of correct answers out of total questions in that book. The total questions in each book is calculated by searching for, and then counting all questions that contain that books tag.

I also added a secondary bar that shows the number of questions attempted, which sits behind the "correct" bar. Each bar has a title on hover (and on screenread) that indicates its statistic: x [attempted|correct] out of y.

image

### DIFF
--- a/src/IsaacAppTypes.tsx
+++ b/src/IsaacAppTypes.tsx
@@ -839,15 +839,15 @@ export enum ATTENDANCE {
 }
 
 export interface QuestionSearchQuery {
-    searchString: string;
-    tags: string;
+    searchString?: string;
+    tags?: string;
     levels?: string;
     stages?: string;
     difficulties?: string;
     examBoards?: string;
-    fasttrack: boolean;
-    startIndex: number;
-    limit: number;
+    fasttrack?: boolean;
+    startIndex?: number;
+    limit?: number;
 }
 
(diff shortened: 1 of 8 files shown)


jsharkey13 (changes requested):
This makes a request for all book questions (_all_ 3000+ of them!) and then counts them for every page load. It's over 1.2MB of JSON! Aside from the (enormous!) computational overhead and expense of us doing this every time, and the bandwidth cost to us, and the bandwidth cost to the end users, it takes over 4s locally which makes the page perform even more poorly than it already did.
We cannot use this approach.

chrisjpurdy:
> This makes a request for all book questions (_all_ 3000+ of them!) and then counts them for every page load. It's over 1.2MB of JSON! Aside from the (enormous!) computational overhead and expense of us doing this every time, and the bandwidth cost to us, and the bandwidth cost to the end users, it takes over 4s locally which makes the page perform even more poorly than it already did. We cannot use this approach.

Fair enough, I totally agree. I built it this way as an MVP really - for it to be sustainable we could precompute this data and put in in elasticsearch when ETL is run, or something similar?

jsharkey13 (approved):

PR #576, opened 2022-05-24
Remove hardcoded special cases for regression test page

mwtrew:
This is no longer required, as arbitrary pages may be tagged with 'regression_test' for the same behaviour.


jsharkey13 (approved):

jsharkey13:
Maybe remove the unused import, though?

PR #579, opened 2022-05-27
Implements suspense and lazy loading examples

bh412:
This PR pulls out of the main bundle the following libraries:

- leaflet (and react-leaflet)
- inequality (and inequality-grammar)
- highlight.js
- p5
- react-beautiful-dnd

Along with a bunch of components that aren't used on the initial page load. Specifically, all of the question components are now loaded lazily to make the types nicer (a lot/most of them need to be loaded lazily for the bigger libraries to be split off anyway).

This improves the performance of the site on initial load by a fair amount, displayed in the attached lighthouse reports:
[before-code-splitting-lighthouse-report.pdf](https://github.com/isaacphysics/isaac-react-app/files/8798371/before-code-splitting-lighthouse-report.pdf)
[after-code-splitting-lighthouse-report.pdf](https://github.com/isaacphysics/isaac-react-app/files/8798370/after-code-splitting-lighthouse-report.pdf)

The final chunk diagram:
image

For interest, here's a (slightly outdated) chunk diagram using webpacks `MinChunkSizePlugin`:
image
It isn't much better (if better at all) than without the plugin, because modules that are required simultaneously are bundled separately still, but also have a bunch of unneeded modules bundled alongside them (for example, `MyProgress` is bundled alongside `highlight.js`, which is doesn't use, but not `billboard.js`, which it does use).


jsharkey13:
FWIW, I did review this before it got merged!

PR #580, opened 2022-05-31
Remove question type duplication in `quiz.ts`

chrisjpurdy:
The circular dependency no longer seems to be an issue (maybe since some files are lazy loaded)


mlt47 (approved):
The recent change to `questions.ts`, which makes the loading of each Question component lazy, has very likely broken the circular dependency problem - that explains why this change is now plausible. I suggest we merge but keep it in mind if we come across any unusual behaviour during regression testing.

PR #582, opened 2022-06-01
Hack to get keyboard interaction working on cloze questions

chrisjpurdy:
This takes `useKeyboardSensor` from `react-beautiful-dnd`, and adds some extra (nasty) stuff so that it behaves as expected with cloze questions. Since you can only interact with one cloze question at a time, this should work fine in theory, and is certainly better than the non-functioning keyboard interaction we had already. It is now actually possible to answer a cloze question only using the keyboard.

If the question re-renders underneath a drag, the drag will cancel if it can't find any elements - this shouldn't really happen much, if at all.

I also changed around some ids so that `react-beautiful-dnd` gives nice screenreader messages. There is a small chance that multiple cloze questions on a single page could break because of this, but looking at the docs it seems like `react-beautiful-dnd` separates different drag-and-drop contexts and the items inside them without us having to worry.

### DIFF
--- a/src/app/components/content/IsaacClozeQuestion.tsx
+++ b/src/app/components/content/IsaacClozeQuestion.tsx
@@ -26,7 +26,7 @@ const IsaacClozeQuestion = ({doc, questionId, readonly}: IsaacQuestionProps<Isaa
     const cssFriendlyQuestionPartId = questionId?.replace(/\|/g, '-') ?? ""; // Maybe we should clean up IDs more?
     const withReplacement = doc.withReplacement ?? false;
 
-    const itemsSectionId = "Non-selected-items"; // `${cssFriendlyQuestionPartId}-items-section`;
+    const itemsSectionId = "Non-selected-items";
 
     const [nonSelectedItems, setNonSelectedItems] = useState<ClozeItemDTO[]>(doc.items ? [...doc.items].map(x => ({...x, replacementId: x.id})) : []);
 
--- a/src/app/components/content/IsaacClozeQuestion.tsx
+++ b/src/app/components/content/IsaacClozeQuestion.tsx
@@ -10,11 +10,14 @@ import {
     DragUpdate,
     Droppable,
     DropResult,
-    ResponderProvided
+    ResponderProvided, useKeyboardSensor,
+    useMouseSensor,
+    useTouchSensor
 } from "react-beautiful-dnd";
 import {ClozeDropRegionContext, ClozeItemDTO, IsaacQuestionProps} from "../../../IsaacAppTypes";
 import {v4 as uuid_v4} from "uuid";
 import {Item} from "../elements/markup/portals/InlineDropZones";
+import {buildUseKeyboardSensor} from "../../services/clozeQuestionKeyboardSensor";
 
 const IsaacClozeQuestion = ({doc, questionId, readonly}: IsaacQuestionProps<IsaacClozeQuestionDTO>) => {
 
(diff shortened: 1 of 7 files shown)


src/app/components/content/IsaacClozeQuestion.tsx
```diff
@@ -23,7 +26,7 @@ const IsaacClozeQuestion = ({doc, questionId, readonly}: IsaacQuestionProps<Isaa
     const cssFriendlyQuestionPartId = questionId?.replace(/\|/g, '-') ?? ""; // Maybe we should clean up IDs more?
     const withReplacement = doc.withReplacement ?? false;
 
-    const itemsSection = `${cssFriendlyQuestionPartId}-items-section`;
+    const itemsSectionId = "Non-selected-items";
```
jsharkey13, line 29: Does this not need to be unique across the whole page? Does it work okay with multiple cloze questions next to one another?

src/app/components/elements/markup/portals/Tables.tsx
```diff
@@ -55,8 +55,8 @@ export const useExpandContent = (expandable: boolean, el?: HTMLElement, unexpand
     const show = SITE_SUBJECT === SITE.CS && expandable && !isMobile() && above["md"](deviceSize) && !expandableParent;
 
     const expandButton = (show && <div className={"expand-button position-relative"}>
-        <button type={"button"} onClick={toggleExpanded}>
-            <div><span><img aria-hidden alt={"Button to expand content"} src={"/assets/expand-arrow.svg"}/> {expanded ? "Close" : "Expand"}</span></div>
+        <button type={"button"} aria-label={"Expand content"} onClick={toggleExpanded}>
+            <div><span><img aria-hidden src={"/assets/expand-arrow.svg"} alt={"Expand icon"}/> {expanded ? "Close" : "Expand"}</span></div>
```
jsharkey13, line 59: This probably shouldn't have `aria-hidden` and alt text. I'd get rid of the alt-text.

src/app/services/clozeQuestionKeyboardSensor.ts
```diff
@@ -0,0 +1,300 @@
+import {DraggableId, FluidDragActions, PreDragActions, SensorAPI} from "react-beautiful-dnd";
+import {useCallback, useLayoutEffect, useMemo, useRef} from "react";
+
+/**
+ * This entire file is essentially a hacked version of https://github.com/atlassian/react-beautiful-dnd/blob/master/src/view/use-sensor-marshal/sensors/use-keyboard-sensor.js
```
jsharkey13, line 5: Could this link either use a tag or commit SHA in place of `master`, so that we can keep track of which version of the file it was and review it if/when we update the library?

src/app/services/katex-a11y.js
```diff
@@ -569,9 +569,14 @@ const handleObject = (tree, a11yStrings, atomType) => {
                     modifier = "";
             }
             buildRegion(a11yStrings, function(regionStrings) {
-                regionStrings.push(`start ${modifier} text`.replace(/\s+/, " "));
-                buildA11yStrings(tree.body, regionStrings, atomType);
-                regionStrings.push(`end ${modifier} text`.replace(/\s+/, " "));
+                const dropZoneRegex = /\[drop-zone(?<params>\|(?<width>w-\d+?)?(?<height>h-\d+?)?)?]/g;
+                if (tree.body.map(a => a.hasOwnProperty("text") ? a.text : "").join("").search(dropZoneRegex) !== -1) {
+                    regionStrings.push("clickable drop zone in la-tech");
```
jsharkey13, line 574: Most users won't know what LaTeX is, so I'd probably not mention that in the screenreader text. Also, this'd be super confusing for Braille text users; don't try to spell things phonetically like this! Surely "cloze question drop zone" or suchlike would do?

src/app/components/elements/markup/portals/Tables.tsx
chrisjpurdy: Yeah fair enough, it was WebStorm wanting me to put it there in-case the image doesn't load.

src/app/services/katex-a11y.js
chrisjpurdy: The drop zones get read out as "clickable drop zone" by the screenreader, so I'll just remove the "la-tech" bit!

src/app/services/clozeQuestionKeyboardSensor.ts
chrisjpurdy: Can do - the library is currently (for the last couple years) not being maintained apart from bug fixes and dependency updates though, so I imagine this code will probably stay the same.

src/app/components/content/IsaacClozeQuestion.tsx
chrisjpurdy: Yep it works fine! `react-beautiful-dnd` also internally labels draggables and droppables with the context they exist in while they're being dragged - not entirely sure how but it worked fine when I tested it.

mlt47 (approved):
This is a clear improvement on the old keyboard interaction experience with cloze questions. It's a shame that we've had to "fork" the keyboard sensor stuff but it's necessary and the library's repo status looks like the code won't change too much in the immediate future.

PR #585, opened 2022-06-14
Move invalid format feedback to app

chrisjpurdy:
Also put in additions requested by phy


jsharkey13 (approved):
This is good, but I don't think the tags should be so terse/cryptic. I think `unrecognised_format` and `invalid_std_form` would be clearer.

jsharkey13 (approved):

PR #586, opened 2022-06-15
Make use of isPhy, isCS, and siteSpecific everywhere

chrisjpurdy:
The focus of this is to have a standard interface for deciding which content to show based on the current site. It's big, but the code should be nicer to read from now on.
This also stops a large amount of objects being created (not sure if this is an optimisation or not).

### DIFF
--- a/src/app/components/content/IsaacAccordion.tsx
+++ b/src/app/components/content/IsaacAccordion.tsx
@@ -11,16 +11,16 @@ import {
 } from "../../services/userContext";
 import {useSelector} from "react-redux";
 import {selectors} from "../../state/selectors";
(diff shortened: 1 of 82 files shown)


chrisjpurdy (commented):
Self-review to check for mistakes and possible improvements. Found a mistake and have fixed it, along with changing some other non-mistakes that are now written more clearly.

chrisjpurdy (commented):
Re-reviewed to check for mistakes in latest changes, none found 🎆

mlt47 (approved):
Looks fine. I wouldn't say this is a particularly important change and probably not worth worrying too much about the performance implications of the small objects.
In github it is not clear that the first argument is the physics but in a modern IDE it'll tell you the argument name.

PR #589, opened 2022-06-20
Question confidence research gameboards

chrisjpurdy:
This does the following:
 - Implements confidence questions (pre and post) for arbitrary question types
 - Conditionally enables the use of these based on whether the gameboard is tagged with `CONFIDENCE_RESEARCH_BOARD` (option added to CS gameboard builder)
 - Record hint views on a question with confidence buttons as a log event with the tag `QUESTION_CONFIDENCE_HINT`, along with an id relating it to the current "confidence session"
 - Generates a new confidence session ID when "a new attempt is started" for a given question. This is different based on whether the question is a question or quick question:
   - Question: when the current attempt changes with the post-question being shown, or when the post-question is answered
   - Quick question: whenever the answer is hidden, or when the post-question is answered
 - Remove the confidence session ID, so data must be grouped by timestamp and user ID

Decent effort has been made to separate out the logic for confidence buttons as much as possible, so they can be easily modified in isolation, or removed completely if needed.
 
See the Google doc attached to the card for details on requirements for this change.


src/app/components/content/IsaacQuickQuestion.tsx
```diff
@@ -1,67 +1,85 @@
...
 
-    const hideAnswer = () => {
-        setVisible(false);
-        setHideOptions(false);
-        attemptUuid.current = uuid_v4().slice(0, 8);
-    };
+    // We have 3 possible styles for the Show/Hide options (default, fast-track and confidence questions)
 
-    const defaultOptions = <Row>
+    const defaultOptions = () => <Row>
```
mlt47, line 36: Why not make these functional components rather than functions that return JSX (even here within the other component)? Although they are not likely to be reused, it might be helpful for dev tools to record each layer of the react tree.

src/app/components/elements/inputs/ConfidenceQuestions.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const confidenceDisabled = type === "question" && (!canSubmit || !currentAttempt || !currentAttempt.value);
+    const showQuestionFeedback = confidenceState !== "initial" || !show || correct;
+
+    // Reset question confidence on attempt change
+    useEffect(() => {
+        if (type === "question") {
+            // If user had answered the question, but is now changing their answer, then generate a new session id
+            if (confidenceState === "followUp") {
+                confidenceSessionUuid.current = uuid_v4().slice(0, 8);
+            }
```
mlt47, line 188: I'm not sure if we're benefitting much from doing this within Isaac. We will need to group log events chronologically during analysis outside of Isaac either way, i.e. to get information such as "was the next attempt correct". We could instead just send a flag into hint views or get hints components to check whether they are part of a research board to decide whether or not to also send the research hint view log event (maybe wrap it all up in a hook).
It doesn't do much harm to keep it in as long as we do at least think about whether or not we've got the correct groupings when we're doing the analysis. There may be unexpected behaviours like [Comparison](https://dl.acm.org/doi/10.1145/2876034.2893379), where students look at hints after getting the answer correct, that are interesting valid behaviours that we might miss by pre-emptive grouping of the log actions (I think Comparison would get caught with the session as it is, btw).

mlt47 (changes requested):

src/app/components/elements/inputs/ConfidenceQuestions.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const toggle = (confidence: string, state: ActiveConfidenceState) => {
+        const stateAndType: `${ActiveConfidenceState} & ${ConfidenceType}` = `${state} & ${type}`;
+        switch (stateAndType) {
+            case "initial & question":
+                dispatch(logAction({
+                    type: "QUESTION_CONFIDENCE_BEFORE",
+                    questionId: identifier,
+                    confidenceSessionUuid: confidenceSessionUuid.current,
+                    answer,
+                    answerCorrect: correct,
```
mlt47, line 107: My local database is a little messed up for testing. I can't see how this can be the value after the response comes back from submitting the form and getting the attempt marked by the server?

chrisjpurdy: Yep this is definitely going to be the old answer - this was already here when I picked up the work so I assumed it just worked, my bad! I think I've got an idea to fix this, it might be a little hard to do though

src/app/components/elements/inputs/ConfidenceQuestions.tsx
chrisjpurdy: I've thought about the pairing of hints to previous attempts but its quite difficult as you would need two session IDs in use concurrently. 
WRT to the benefit of doing this, it makes comparing confidence pre and post feedback much easier - you can just merge entries with the same session ID, but this approach is only relevant to that specific research question.
I think you're right in that a single session identifier would be better to answer almost any other questions, and that the analysis can be done using timestamps within a single session for most questions and given a bit more work with scripting.

chrisjpurdy: On reflection I've removed the session ID all-together - it provides no benefit over a combined key of [day + user ID]. Maybe if it was much more sophisticated in what event are associated with each session then it would be better, but this is hard to do in a way that is flexible to answering every question.

src/app/components/elements/inputs/ConfidenceQuestions.tsx
```diff
@@ -0,0 +1,205 @@
...
+            initial: {
+                question: "What is your level of confidence that your own answer is correct?",
+                options: [
+                    {label: "Low", color: "negative"},
+                    {label: "Medium", color: "neutral"},
+                    {label: "High", color: "positive"}
+                ]
+            },
+            followUp: {
+                question: "Having read the feedback, how do you rate your confidence in answering this question again?",
```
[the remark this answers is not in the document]
chrisjpurdy: I've asked Jane and Eirini to double check this and the following comment.

src/app/components/content/IsaacQuestion.tsx
```diff
@@ -92,12 +105,12 @@ export const IsaacQuestion = withRouter(({doc, location}: {doc: ApiTypes.Questio
             </Suspense>
 
             {/* CS Hints */}
-            {isCS && <React.Fragment>
+            {isCS && <ConfidenceContext.Provider value={{recordConfidence}}>
```
mlt47, line 108: Rather than having two ConfidenceContext.Providers in this component, would there be a problem with just wrapping the IsaacQuestion component in one around the Form?

src/app/components/elements/inputs/ConfidenceQuestions.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    confidence
+                }));
+                setState("hidden");
+                break;
+        }
+    };
+
+    useEffect(() => {
+        // The validation response gets set to undefined whenever the current attempt changes, so I am very sure
+        // that this log action will only run once for a given attempt.
```
mlt47, line 133: This feels long-term fragile, but there isn't an easy alternative solution. Perhaps there exists an option using redux middleware where a state machine of sorts can make decisions based on which actions it sees and that might be a bit more explicit? Seeing as this feature is not fundamental to the use of Isaac, I don't think it's worth making the effort to rearchitect it yet.

mlt47 (approved):

src/app/components/content/IsaacQuestion.tsx
chrisjpurdy: No problem at all - that's sorted 👍

src/app/components/elements/inputs/ConfidenceQuestions.tsx
chrisjpurdy: I agree that it's fragile, but building a long term solution is a bit out of scope as you say, since this might be temporary and in any case is not a major part of Isaac.

mlt47 (approved):

PR #592, opened 2022-06-24
Rename waiting list to request a place for waiting list only events

mwtrew:
This PR removes references to the waiting list for WAITING_LIST_ONLY events, so it should be clearer to guests that they are actually requesting one of the available places (this is how WAITING_LIST_ONLY is used in practice).


chrisjpurdy (approved):
LGTM - the functions for formatting each booking message make things nice and clean too!

chrisjpurdy:
It might be worth looking at `AddUsersToBooking.tsx` and `UserBookingModal.tsx` as well in case wording needs changing there too

PR #594, opened 2022-06-28
Allow registration for 10-12 year olds on physics

chrisjpurdy:
I've used placeholder text in the "ask your parents for consent" pop-up. 

There is another card about the proper wording of this pop-up, so this PR will have to wait for that to be figured out and copy-edited before being merged.

**It would still be useful for this to be reviewed**, so that the only thing left to do is work out the wording of the consent box.

### DIFF
--- a/src/app/components/pages/Registration.tsx
+++ b/src/app/components/pages/Registration.tsx
@@ -309,7 +309,7 @@ export const Registration = withRouter(({location}:  RouteComponentProps<{}, {},
                         <CustomInput
                             id="consent-checkbox" name="consent-checkbox" type="checkbox"
                             checked={parentalConsentCheckboxChecked}
-                            label="Please confirm that you have read and understood this message."
+                            label="Please check the box to confirm that you have read and understood this message."
                             onChange={(e) => setParentalConsentCheckboxChecked(e?.target.checked)}
                         />
                     </Alert>}
--- a/src/app/components/pages/Registration.tsx
+++ b/src/app/components/pages/Registration.tsx
@@ -3,6 +3,7 @@ import {useDispatch, useSelector} from "react-redux";
 import {Link} from "react-router-dom";
 import ReactGA from "react-ga";
 import {
+    Alert,
     Card,
     CardBody,
     CardTitle,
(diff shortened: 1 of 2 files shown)


src/app/components/pages/Registration.tsx
```diff
@@ -237,31 +244,37 @@ export const Registration = withRouter(({location}:  RouteComponentProps<{}, {},
...
                                             required
                                             label="I am at least 13 years old"
-                                            disabled={registrationUser.dateOfBirth}
-                                            // TODO: Look at DateInput null vs undefined for updating DoB and maybe
-                                            // change this in future
-                                            onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
-                                                setDobCheckboxChecked(!dobCheckboxChecked);
-                                            }}
+                                            disabled={(isPhy && dob10To12CheckboxChecked) || registrationUser.dateOfBirth}
+                                            // TODO: Look at DateInput null vs undefined for updating DoB and maybe change this in future
```
mwtrew, line 266: I'm not quite sure what this todo is suggesting, can we clarify (or ideally address/remove) it?

src/app/components/pages/Registration.tsx
```diff
@@ -278,12 +291,31 @@ export const Registration = withRouter(({location}:  RouteComponentProps<{}, {},
...
                             </h4>
                         </Col>
                     </Row>
+
+                    {/* 10-12 parental consent box */}
+                    {isPhy && confirmedTenToTwelve && <Alert color={"warning"}>
+                        <h2>Attention young'un</h2>
+                        <hr/>
+                        <p>
+                            You're under 13 and you wanna use Isaac Phy? You better make sure that you ask your mum or dad or guardian or whoever because there's some really scary stuff on this platform that they might want to know about beforehand.
```
mwtrew, line 306: I don't think we need to wait for text from the content team when this is already perfect.

mwtrew (approved):

src/app/components/pages/Registration.tsx
chrisjpurdy: I had no idea either, I'll git blame it!

chrisjpurdy: Yeah it can almost certainly be safely removed

mwtrew: Ah - I thought you added it there, my mistake. It's good to get these things tidied up regardless!

PR #598, opened 2022-07-01
Move `validationPending` state up to allow post question to show safely

chrisjpurdy:
The issue was that the question flow could be corrupted by the user answering twice in quick succession (within the window of the server returning a response to an attempt), so if a validation was pending, the we needed to be able to cancel the pending state change to the confidence buttons if another answer was submitted. 
Beforehand I tried to do this with a crude work-around that ended up not working.
Doing this properly just involved allowing the `useConfidenceQuestionsValues` hook to manage the `validationPending` state, meaning it could reset it if the current attempt changes.

This PR also fixed the confidence questions not working on item questions.


mlt47 (approved):
This looks OK

PR #600, opened 2022-07-06
Add content renderer containers for editor preview functionality

mlt47:


jsharkey13 (approved):
I'm not sure I like that a dev-only service lives in the live configuration, but it can't easily go in `./compose` so maybe this is easier than a new script.

Still all looks okay, looking forward to doubling our build times 😂

mlt47:
> Still all looks okay, looking forward to doubling our build times 😂
Yes... the build times is a shame! If it becomes too annoying we can look into ways of speeding it up through reuse.

PR #602, opened 2022-07-13
Changes to item section ID for screen reader messages

chrisjpurdy:
This also improves readability a bit. 

I've reverted the IDs, so they're no longer unique, but have removed the `htmlFor` attribute from the item section `Label` so that browsers don't get confused if there are multiple item sections on screen. We don't really need that label for screen reader purposes since the item section element has an `aria-label`, and `react-beautiful-dnd` produces nice messages anyway.


mlt47 (approved):
This looks good and does improve the readability of the code.

PR #603, opened 2022-07-14
Gameboards "View more" incorrect ordering fix (and refactor)

chrisjpurdy:
The primary goal of this was to make sure the "View more" button added new boards to the end of the list. That is solved by [this commit](https://github.com/isaacphysics/isaac-react-app/commit/4f4203499c8c8e82ac0a105f63000bfaf0dd475e).

I also made a couple of larger changes since the code for these components really needed a makeover:

---

A visual adjustment to `SetAssignments` and `MyGameboards` (see first commit), which makes the card view for both pages look much neater. James and Meurig approved of this change when demoing it to them.

---

The 6th commit pulls out the gameboard fetching logic for `MyGameboards` and `SetAssignments` into a hook defined in `gameboards.tsx`, and refactors the code in a few other ways:
 - Gets rid of `actualBoardLimit` (this name confused me for about an hour) and replaces it with `numberOfBoards`
 - Consistent variable name style for enums (CamelCase)
 -  Adds a bunch of comments

### DIFF
--- a/src/app/components/pages/MyGameboards.tsx
+++ b/src/app/components/pages/MyGameboards.tsx
@@ -304,7 +304,7 @@ export const MyGameboards = () => {
                 <div className="mt-4 mb-2">
                     {boards && boards.totalResults > 0 && <h4>You have completed <strong>{completed}</strong> of <strong>{boards.totalResults}</strong> gameboard{boards.totalResults > 1 && "s"},
                         with <strong>{inProgress}</strong> on the go and <strong>{notStarted}</strong> not started</h4>}
-                    {!boards && <h4>You have <IsaacSpinner size="sm" /> saved gameboards...</h4>}
+                    {!boards && <h4>You have <IsaacSpinner size="sm" inline /> saved gameboards...</h4>}
                 </div>
                 <div>
                     {boardView !== boardViews.table && <Row>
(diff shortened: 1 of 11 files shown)


src/app/components/pages/MyGameboards.tsx
```diff
@@ -164,7 +140,7 @@ const Board = (props: BoardTableProps) => {
                         {`Stage${boardStages.length !== 1 ? "s" : ""}: `}<strong>{boardStages.map(s => stageLabelMap[s]).join(', ') || "N/A"}</strong>
                     </CardSubtitle>
                     <CardSubtitle>
-                        {`Difficult${boardStages.length !== 1 ? "ies" : "y"}: `}
+                        {`Difficult${boardDifficulties.length !== 1 ? "ies" : "y"}: `}
```
mlt47, line 143: 🙃

mlt47 (commented):
This looks fine and works fine from what I could see.
I think bringing the set assignment panel out as a modal is nicer but we should let the content teams know in advance, they might feel like they want to update their documentation to match the new behaviour.

It is also a bit of a shame that rows including one card now stretch to fill the row rather than being horizontally the same width as the other cards.

chrisjpurdy:
> It is also a bit of a shame that rows including one card now stretch to fill the row rather than being horizontally the same width as the other cards.

This is fixed - the cards all stay the same width now! I've also sent a message to the Physics content team to ask for opinions on the change to using the modal.

PR #605, opened 2022-07-15
Allow the top-right search bar to work on the /search page

mlt47:
This PR fixes the search bar in the top-right on the /search page.
This is achieved by making the query in the URL query params the definitive source for queries sent to the server. Interacting with filters and input fields changes the URL; then the URL fields are used for sending to the server.


Morpheu5 (approved):

mwtrew (commented):
One thing that might be a bit counter-intuitive: after a search if you edit the filters then search for something else with the top-right search bar, the filters are reset as far as the URL/query is concerned but this is not reflected in the "filters" box, so it looks out of sync with the results.

mwtrew (approved):

PR #606, opened 2022-07-15
User context (stage, exam board, school) reconfirmation modal

chrisjpurdy:
Modal will show to anyone who hasn't updated their user context details since the most recent 1st August.

**After final copyedit:**

Teacher relevant text on Phy:
image

Student relevant text on CS
image

### DIFF
--- a/src/app/state/middleware/utils.ts
+++ b/src/app/state/middleware/utils.ts
@@ -1,23 +1,10 @@
-// Returns true if lastConfirmedDate is before *last August*, and it's at least the 24th of August. False otherwise
-export const needToUpdateUserContextDetails = (lastConfirmedDate: Date | undefined) => {
-    // First calculate when the beginning of last August before today was...
+export const needToUpdateUserContextDetails = (lastConfirmedDate: Date | undefined): boolean => {
     let date = new Date();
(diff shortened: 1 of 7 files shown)


src/app/state/middleware/utils.ts
```diff
@@ -0,0 +1,10 @@
+export const needToUpdateUserContextDetails = (lastConfirmedDate: Date | undefined): boolean => {
+    let date = new Date();
+    if (date.getMonth() < 7) { // Month is 0-indexed for some stupid reason
+        date.setFullYear(date.getFullYear() - 1);
+    }
+    date.setMonth(7, 1);
```
Morpheu5, line 6: Just a comment, don't mind me, but am I the only one enraged that months start at 0 and days start at 1?!

Morpheu5 (approved):

src/app/state/middleware/utils.ts
chrisjpurdy: Not at all, I also lost my cool a bit and added a comment in that snippet 😅

Morpheu5: Ah yes, completely overlooked that in my rage! 😁

PR #609, opened 2022-07-19
RTK refactor: Account MFA

chrisjpurdy:
This adds RTK to the app, along with refactoring the setup MFA logic to use RTK query. This is a fully-worked basic use case for RTK, both to get the boilerplate for the refactor implemented and to provide an example to anyone who wants to help.

I also changed all occurrences of `useDispatch` => `useAppDispatch` and `useSelector` => `useAppSelector`. These `..App...` versions are (almost) correctly typed, and we can change these types to restrict/help infer parameters to these hooks.


chrisjpurdy:
I can't fix the `actions` tests and can't figure out where it's breaking - it looks like it's because `thunk` middleware isn't being used, but AFAIK it is being added to the `store` by `getDefaultMiddleware` so IDK what's going on.

src/app/components/elements/panels/UserMFA.tsx
```diff
@@ -16,14 +16,17 @@ interface UserMFAProps {
...
-    const totpSharedSecret = useSelector((state: AppState) => state?.totpSharedSecret?.sharedSecret);
-    const [updateMFARequest, setUpdateMFARequest] = useState(false);
+    const segueEnvironment = useAppSelector(selectors.segue.environmentOrUnknown);
+    const totpSharedSecret = useAppSelector((state: AppState) => state?.totpSharedSecret?.sharedSecret);
+    const [showMFAConfig, setShowMFAConfig] = useState(false);
     const [successfulMFASetup, setSuccessfulMFASetup] = useState(false);
     const [mfaVerificationCode, setMFAVerificationCode] = useState<string | undefined>(undefined);
     const [qrCodeStringBase64SVG, setQrCodeStringBase64SVG] = useState<string | undefined>(undefined);
 
+    const [ newMFASecret, { isSuccess: receivedNewMFASecret } ] = isaacApi.endpoints.newMFASecret.useMutation();
```
mlt47, line 26: I was wondering, as the second value from the `useMutation()`'s returned tuple is the result. Could we then get totpSharedSecret from that and get rid of the need for the totpSharedSecret slice? That slice's reducer logic would then live in the `transformResponse` for the endpoint.
It is very possible I am missing something here so please test if you do try it out!

chrisjpurdy: Have done this in the most recent commit - it makes the code so much nicer! 🎉

src/app/components/elements/panels/UserMFA.tsx
```diff
@@ -16,21 +16,23 @@ interface UserMFAProps {
...
     const [mfaVerificationCode, setMFAVerificationCode] = useState<string | undefined>(undefined);
     const [qrCodeStringBase64SVG, setQrCodeStringBase64SVG] = useState<string | undefined>(undefined);
 
+    const [ newMFASecret, { data: totpSharedSecret, reset: resetMFASecret } ] = isaacApi.endpoints.newMFASecret.useMutation();
+    const [ setupAccountMFA ] = isaacApi.endpoints.setupAccountMFA.useMutation();
+    const [ disableAccountMFA ] = isaacApi.endpoints.disableAccountMFA.useMutation();
+
     const authenticatorURL: string | null = useMemo(() => {
-        if (totpSharedSecret) {
+        console.log("New secret! Creating QR code...", totpSharedSecret);
```
mlt47, line 29:
```suggestion
```

src/app/components/elements/panels/UserMFA.tsx
mlt47: Great!

src/app/components/elements/panels/UserMFA.tsx
chrisjpurdy: 🙃

PR #613, opened 2022-08-01
Add dummy "null item" to cloze choices

chrisjpurdy:
This means that spaces left in cloze questions are persisted when passed between the frontend and API

### DIFF
--- a/src/app/components/content/IsaacClozeQuestion.tsx
+++ b/src/app/components/content/IsaacClozeQuestion.tsx
@@ -19,13 +19,21 @@ import {v4 as uuid_v4} from "uuid";
 import {Item} from "../elements/markup/portals/InlineDropZones";
 import {buildUseKeyboardSensor} from "../../services/clozeQuestionKeyboardSensor";
 
-const augmentInlineItemWithUniqueReplacementID = (idv: ClozeItemDTO) => idv ? ({...idv, replacementId: `${idv?.id}-${uuid_v4()}`}) : undefined;
+const augmentInlineItemWithUniqueReplacementID = (idv: ClozeItemDTO | undefined) => idv ? ({...idv, replacementId: `${idv?.id}-${uuid_v4()}`}) : undefined;
 const augmentNonSelectedItemWithReplacementID = (item: ClozeItemDTO) => ({...item, replacementId: item.id});
-const itemNotInAttempt = (currentAttempt: ItemChoiceDTO) => (i: ClozeItemDTO) => !currentAttempt.items?.map(si => si?.id).includes(i.id);
+const itemNotNullAndNotInAttempt = (currentAttempt: {items?: (ItemDTO | undefined)[]}) => (i: ClozeItemDTO | undefined) => i ? !currentAttempt.items?.map(si => si?.id).includes(i.id) : false;
+
+const NULL_CLOZE_ITEM_ID = "NULL_CLOZE_ITEM" as const;
+const NULL_CLOZE_ITEM: ItemDTO = {
+    type: "itemChoice",
+    id: NULL_CLOZE_ITEM_ID
+};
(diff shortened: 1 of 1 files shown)


src/app/components/content/IsaacClozeQuestion.tsx
```diff
@@ -19,13 +19,21 @@ import {v4 as uuid_v4} from "uuid";
 import {Item} from "../elements/markup/portals/InlineDropZones";
 import {buildUseKeyboardSensor} from "../../services/clozeQuestionKeyboardSensor";
 
-const augmentInlineItemWithUniqueReplacementID = (idv: ClozeItemDTO) => idv ? ({...idv, replacementId: `${idv?.id}-${uuid_v4()}`}) : undefined;
+const augmentInlineItemWithUniqueReplacementID = (idv: ClozeItemDTO | undefined) => idv ? ({...idv, replacementId: `${idv?.id}-${uuid_v4()}`}) : undefined;
```
Morpheu5, line 22: This may also appear elsewhere. I know checking for definedness in this way is idiomatic JS, and it's fine in this case, but it shouldn't because a value of `0` or `""`, for example, is defined but would test false-y. Using the `isDefined()` utility is better because it checks for `null` and `undefined` specifically.

```suggestion
import {isDefined} from "../../services/miscUtils";

const augmentInlineItemWithUniqueReplacementID = (idv: ClozeItemDTO | undefined) => isDefined(idv) ? ({...idv, replacementId: `${idv?.id}-${uuid_v4()}`}) : undefined;
```

Morpheu5 (approved):
Looks good, other than that suggested change.

src/app/components/content/IsaacClozeQuestion.tsx
chrisjpurdy: I do use that utility a fair amount, not sure why I didn't add it here!

src/app/components/content/IsaacClozeQuestion.tsx
```diff
@@ -19,13 +19,23 @@ import {v4 as uuid_v4} from "uuid";
...
+import {isDefined} from "../../services/miscUtils";
+
+const augmentInlineItemWithUniqueReplacementID = (idv: ClozeItemDTO | undefined) => isDefined(idv) ? ({...idv, replacementId: `${idv?.id}-${uuid_v4()}`}) : undefined;
 const augmentNonSelectedItemWithReplacementID = (item: ClozeItemDTO) => ({...item, replacementId: item.id});
-const itemNotInAttempt = (currentAttempt: ItemChoiceDTO) => (i: ClozeItemDTO) => !currentAttempt.items?.map(si => si?.id).includes(i.id);
+const itemNotNullAndNotInAttempt = (currentAttempt: {items?: (ItemDTO | undefined)[]}) => (i: ClozeItemDTO | undefined) => i ? !currentAttempt.items?.map(si => si?.id).includes(i.id) : false;
+
+const NULL_CLOZE_ITEM_ID = "NULL_CLOZE_ITEM" as const;
+const NULL_CLOZE_ITEM: ItemDTO = {
+    type: "itemChoice",
```
jsharkey13, line 30:
```suggestion
    type: "item",
```

This is incorrect! And I didn't review this before merging it because it was approved 🤦

jsharkey13: I'll fix it now!

jsharkey13: It turns out that, since both types inherit from `Content` and since we don't explicitly check the type of the submitted items (I think maybe we should?!), it does work whatever value this has. But it's fixed now anyway . . .

Morpheu5: Woops, I'd never have spotted that…

PR #630, opened 2022-08-25
Remove P3 and C3 from difficulty filters for CS

mlt47:
This does not remove P3 and C3 from the iconography nor the help tooltip.


mlt47:
This PR has been changed to draft as we are waiting for feedback from the RPF CS team for further input.

chrisjpurdy (approved):

PR #633, opened 2022-09-01
Add a map on /events showing all the events (loaded on the page)

Morpheu5:
This is **waiting** on Fergus for the moment.


mlt47:
If this request gets asked for again we will resurrect. Currently there are no events related to Ada Computer Science.

PR #635, opened 2022-09-02
Front-end dependencies update, September 2022

Morpheu5:
There is still a type error that says

```
[tsl] ERROR in /Users/morpheu5/src/isaac/isaac-react-app/src/app/components/elements/inputs/SchoolInput.tsx(102,17)
      TS2322: Type 'DebouncedFunc<(schoolSearchText: string) => Promise<unknown>>' is not assignable to type '(inputValue: string, callback: (options: OptionsOrGroups<{ value: string; label: string | undefined; }, GroupBase<{ value: string; label: string | undefined; }>>) => void) => void | Promise<...>'.
  Type 'Promise<unknown> | undefined' is not assignable to type 'void | Promise<OptionsOrGroups<{ value: string; label: string | undefined; }, GroupBase<{ value: string; label: string | undefined; }>>>'.
    Type 'Promise<unknown>' is not assignable to type 'void | Promise<OptionsOrGroups<{ value: string; label: string | undefined; }, GroupBase<{ value: string; label: string | undefined; }>>>'.
      Type 'Promise<unknown>' is not assignable to type 'Promise<OptionsOrGroups<{ value: string; label: string | undefined; }, GroupBase<{ value: string; label: string | undefined; }>>>'.
        Type 'unknown' is not assignable to type 'OptionsOrGroups<{ value: string; label: string | undefined; }, GroupBase<{ value: string; label: string | undefined; }>>'.
```

### DIFF
--- a/src/app/components/elements/cards/ActionCard.tsx
+++ b/src/app/components/elements/cards/ActionCard.tsx
@@ -29,7 +29,7 @@ export const ActionCard = ({title, amountText, linkDestination, linkText, coming
                 {comingSoon ? comingSoonMessage : amountText}
             </p>
             <p>
-                {children}
+                <>{children}</>
             </p>
             <RS.Button tag={Link} to={linkDestination} color="secondary" block disabled={comingSoon}>
                 {linkText}
--- a/src/app/components/elements/inputs/ConfidenceQuestions.tsx
+++ b/src/app/components/elements/inputs/ConfidenceQuestions.tsx
@@ -191,7 +191,7 @@ export const useConfidenceQuestionsValues = (show: boolean | undefined, type: Co
     // Confidence question specific things
     const [confidenceState, setConfidenceState] = useState<ConfidenceState>("initial");
     const [validationPending, setValidationPending] = useState<ValidationPendingState>({pending: false});
-    const confidenceDisabled = type === "question" && (!canSubmit || !currentAttempt || (currentAttempt.value === "") || ((currentAttempt as ItemChoiceDTO).items === []));
+    const confidenceDisabled = type === "question" && (!canSubmit || !currentAttempt || (currentAttempt.value === "") || (Array.isArray((currentAttempt as ItemChoiceDTO).items) && (currentAttempt as ItemChoiceDTO).items?.length === 0));
     const showQuestionFeedback = confidenceState !== "initial" || !show || correct;
 
     // Reset question confidence on attempt change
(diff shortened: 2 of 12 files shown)


src/app/components/elements/cards/ActionCard.tsx
```diff
@@ -29,7 +29,7 @@ export const ActionCard = ({title, amountText, linkDestination, linkText, coming
                 {comingSoon ? comingSoonMessage : amountText}
             </p>
             <p>
-                {children}
+                <>{children}</>
```
mlt47, line 32: Mildly annoying that this is now necessary! 🤦

src/app/components/elements/inputs/ConfidenceQuestions.tsx
```diff
@@ -191,7 +191,7 @@ export const useConfidenceQuestionsValues = (show: boolean | undefined, type: Co
     // Confidence question specific things
     const [confidenceState, setConfidenceState] = useState<ConfidenceState>("initial");
     const [validationPending, setValidationPending] = useState<ValidationPendingState>({pending: false});
-    const confidenceDisabled = type === "question" && (!canSubmit || !currentAttempt || (currentAttempt.value === "") || ((currentAttempt as ItemChoiceDTO).items === []));
+    const confidenceDisabled = type === "question" && (!canSubmit || !currentAttempt || (currentAttempt.value === "") || (Array.isArray((currentAttempt as ItemChoiceDTO).items) && (currentAttempt as ItemChoiceDTO).items?.length === 0));
```
mlt47, line 194:
```suggestion
    const confidenceDisabled = type === "question" && (!canSubmit || !currentAttempt || (currentAttempt.value === "") || ((currentAttempt as ItemChoiceDTO).items?.length === 0));
```
I'm also happy to lose some of the brackets around the last two evaluations

src/app/components/elements/inputs/SchoolInput.tsx
```diff
@@ -20,7 +20,7 @@ interface SchoolInputProps {
 const NOT_APPLICABLE = "N/A";
 
 
-const getSchoolPromise = (schoolSearchText: string) =>
+const getSchoolPromise = (schoolSearchText: string): Promise<any> =>
```
mlt47, line 23: There's already a lot of missing types here so I think the any here is fine.
We should instead revisit this Input now that we are more familiar with React as I seem to remember the throttling doesn't quite work.

mlt47 (approved):
Minor suggestions, but I'm ok with the new `any`s as those files already suffer from not being well typed. Getting the updates is more important.

src/app/components/elements/cards/ActionCard.tsx
Morpheu5: Indeed! Any alternative solution is welcome, but it seems like this is the preferred (or only) way.

PR #636, opened 2022-09-05
Prevent unnecessary user updates

jsharkey13:
On login success, the current user object is returned and so no further current_user requests are necessary. However, I think in order to avoid duplicating the logic for getting auth settings and preferences, we then dispatched a current_user request anyway.

This causes the Redux "user" object to change, even though its contents are identical, which causes many useEffects to fire unnecessarily. These unwanted firings caused many duplicate WebSocket requests, and duplicate auth/preference settings requests, which are bad for our server performance as much as the app performance.

After these changes, there's a bug with login. Previously, a login failure would have caused a pointless current_user request which would have reset the "user" state object back and removed the "requesting" property. Now we no longer make that unwanted request, we have to make sure we reset this, else without reloading you can't make another attempt to login if have an error this first time.

---

I am sure that the _best_ approach here would be a new action that does the settings/preferences requests, but we're likely to refactor all this anyhow so this was simpler and we can cope with the duplication.

### DIFF
--- a/src/app/components/navigation/IsaacApp.tsx
+++ b/src/app/components/navigation/IsaacApp.tsx
@@ -106,23 +106,25 @@ export const IsaacApp = () => {
         dispatch(fetchGlossaryTerms());
     }, [dispatch]);
 
+    const loggedInUserId = isLoggedIn(user) ? user.id : undefined;
     useEffect(() => {
-        if (isLoggedIn(user)) {
+        if (loggedInUserId) {
             dispatch(requestNotifications());
-            checkForWebSocket(user);
+            checkForWebSocket(loggedInUserId);
         }
         return () => {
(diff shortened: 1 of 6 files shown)


mlt47 (approved):
I traced through the changes to redux's dispatched actions and changes to state / middleware triggers as a result of that and am happy that everything should work the same with less current user requests.
I added a commit that alters the useEffects with `user` as a dependency to match our actual intentions by using the user ID rather than a change to the user object to trigger the re-running of the effect.

jsharkey13:
Your changes seem sensible, so I'm going to merge this! (I'd tried them initially, and seen that using just the ID number value worked in my testing, but wanted to avoid the extra actions too so didn't go that route. Doing both seems good too!).

PR #639, opened 2022-09-09
Set scheduled assignments at about 7AM in the timezone of the browser

Morpheu5:


src/app/components/pages/SetAssignments.tsx
```diff
@@ -113,6 +113,14 @@ const AssignGroup = ({groups, board, assignBoard}: BoardProps) => {
     const yearRange = range(currentYear, currentYear + 5);
     const currentMonth = (new Date()).getMonth() + 1;
 
+    function setScheduledStartDateAtSevenAM(e: ChangeEvent<HTMLInputElement>) {
+        const utcDate = e.target.valueAsDate as Date;
+        const sevenAM = 7 * 60 * 60 * 1000;
+        const scheduledDate = new Date(utcDate.getTime() + sevenAM);
```
jsharkey13, line 119: I don't think this is correct? For me the timestamp this creates is always 8am BST, because the `utcDate` is midnight UTC (so _1am_ BST) and then adding 7 hours leads to 8am BST. I have tested this suggestion (whilst trying to work out why the change never reached the API).
```suggestion
        const scheduledDate = new Date(utcDate.getFullYear(), utcDate.getMonth(), utcDate.getDate(), 7);
```

jsharkey13 (changes requested):
This doesn't work because the `assignBoard` action in `actions.tsx` ([specifically this line](https://github.com/isaacphysics/isaac-react-app/blob/2c4ae77a0ef4073277791594ab3228cd2fe1f2cb/src/app/state/actions.tsx#L1589)), then removes the timestamp you add here and replaces it with midnight UTC. I think the line in the action can just be deleted, but then the check probably should use the "midnight" value to make sure you don't choose "today" (and risk the scheduler not noticing it if we've messed up the timezones and you do it at just before 7am your time).

src/app/components/pages/SetAssignments.tsx
Morpheu5: Much more readable, thanks!

src/app/state/actions.tsx
```diff
@@ -1586,7 +1586,7 @@ export const assignBoard = (board: GameboardDTO, groups: Item<number>[] = [], du
     }
 
     if (scheduledStartDate != undefined) {
-        scheduledStartDate?.setUTCHours(0, 0, 0, 0);
+        // Unline with the due date, we want to preserve the hour assigned at the UI level, unless we want to move that logic here.
```
chrisjpurdy, line 1589: You might want to check that this logic of making sure the scheduled start date is still in the future checks out still. 
To be honest, I think the logic to set the hour to 7am should be in this file because other logic here (the two comparisons against `scheduledStartDate`) could be affected.
```suggestion
        scheduledStartDate?.setUTCHours(7, 0, 0, 0);
```

chrisjpurdy (commented):
I feel like the logic shouldn't be at the UI level since we don't allow the user to select the hour anyway. Unless we're going to add this functionality soon?

src/app/state/actions.tsx
```diff
@@ -1586,7 +1586,7 @@ export const assignBoard = (board: GameboardDTO, groups: Item<number>[] = [], du
...
         if ((scheduledStartDate.valueOf() - today.valueOf()) < 0) {
```
jsharkey13, line 1590: I think this check is now wrong, because `today` is set to UTC midnight and so you could schedule it for in the past on the current day. At the very least, it should be the current timestamp we're checking against to make sure it's not in the past (say `new Date().valueOf()` instead maybe?).

jsharkey13 (approved):

PR #640, opened 2022-09-14
Patch rtk query bugs

chrisjpurdy:
This fixes the bugs found so far in regression testing the RTK query assignment and gameboard refactor, those being:
 - Assignment progress cache isn't updated when new assignment is made
 - Fast track navigation being "totally broken"
 - Gameboard builder selecting random questions and breaking weirdly when "Create another gameboard" is clicked after creating one first
 - "My assignments" and "My tests" counter overflowing to the next line
 - Gameboard filter loading forever on filter combo without questions, instead of showing "No questions found..."
 - Non-existent gameboard loading forever instead of displaying "Not found"

A couple of the fixes are a bit hacky, the patch to `isaacBaseQuery` in particular needing a revisit to fix the types - ideally, it should be clear from the RTK Query hook types that data might come back as `NOT_FOUND`

### DIFF
--- a/src/app/components/pages/Gameboard.tsx
+++ b/src/app/components/pages/Gameboard.tsx
@@ -140,7 +140,7 @@ export const GameboardViewer = ({gameboard, className}: {gameboard: GameboardDTO
 export const Gameboard = withRouter(({ location }) => {
     const dispatch = useAppDispatch();
     const gameboardId = location.hash ? location.hash.slice(1) : null;
-    const gameboardQuery = isaacApi.endpoints.getGameboardById.useQuery(gameboardId ?? skipToken);
+    const gameboardQuery = isaacApi.endpoints.getGameboardById.useQuery(gameboardId || skipToken);
     const {data: gameboard} = gameboardQuery;
     const user = useAppSelector(selectors.user.orNull);
 
--- a/src/app/components/pages/Gameboard.tsx
+++ b/src/app/components/pages/Gameboard.tsx
@@ -1,5 +1,12 @@
-import React, {useEffect} from "react";
-import {isaacApi, logAction, selectors, useAppDispatch, useAppSelector} from "../../state";
+import React, {useEffect, useState} from "react";
+import {
+    extractDataFromQueryResponse,
+    isaacApi,
+    logAction,
+    selectors,
+    useAppDispatch,
+    useAppSelector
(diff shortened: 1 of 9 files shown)


src/app/components/pages/Gameboard.tsx
```diff
@@ -149,31 +159,7 @@ export const Gameboard = withRouter(({ location }) => {
         }
     }, [dispatch, gameboard]);
 
-    const userButtons = user && isTeacher(user) ?
```
chrisjpurdy, line 152: These lines are just for inlineing `userButtons`

jsharkey13 (approved):
I don't pretend to understand all of this, but none of it looks questionable, so I think it's okay to go.

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -114,7 +126,19 @@ const GameboardBuilder = withRouter((props: { location: { search?: string } }) =
         existing gameboard to duplicate and edit it.
     </span>;
 
+    const sentinal = useRef<HTMLDivElement>(null);
```
chrisjpurdy, line 129:
```suggestion
    const sentinel = useRef<HTMLDivElement>(null);
```

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -114,7 +126,19 @@ const GameboardBuilder = withRouter((props: { location: { search?: string } }) =
...
+
+    const resetBuilder = () => {
+        if (baseGameboard) {
+            cloneGameboard(baseGameboard);
+        } else {
+            initialise();
+        }
+        setTimeout(() => sentinal.current?.scrollIntoView(), 50);
```
chrisjpurdy, line 137:
```suggestion
        setTimeout(() => sentinel.current?.scrollIntoView(), 50);
```

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -114,7 +126,19 @@ const GameboardBuilder = withRouter((props: { location: { search?: string } }) =
...
+    };
+
     return <Container id="gameboard-builder">
+        <div ref={sentinal}/>
```
chrisjpurdy, line 141:
```suggestion
        <div ref={sentinel}/>
```

PR #642, opened 2022-09-20
Assignment Schedule

chrisjpurdy:
This implements a new page to manage assignments based on their start and due dates. It's accessed via `/assignment_schedule`.

The idea is that this is released to live and anyone with the link can access it, and then we add it to the navbar later on. It is probably going to remain separate to the set assignments page since it represents a different part of the assignment workflow.

I have yet to add buttons to add gameboard from books and "boards for lessons", but this shouldn't hold back merging because the whole page is a hidden beta feature subject to change. There also isn't yet functionality to modify your current assignments - this can also be done later since it requires an API change (probably a new endpoint).

There are a couple of changes to gameboards and dates which are transient bug fixes - the commits containing these are separate from the new additions.

### DIFF
--- a/src/app/components/handlers/ShowLoadingQuery.tsx
+++ b/src/app/components/handlers/ShowLoadingQuery.tsx
@@ -1,4 +1,4 @@
-import React, {ReactElement} from "react";
+import React from "react";
 import {FetchBaseQueryError} from "@reduxjs/toolkit/dist/query/fetchBaseQuery";
 import {SerializedError} from "@reduxjs/toolkit";
 import {IsaacSpinner} from "./IsaacSpinner";
--- a/src/app/components/handlers/ShowLoadingQuery.tsx
+++ b/src/app/components/handlers/ShowLoadingQuery.tsx
@@ -2,31 +2,79 @@ import React, {ReactElement} from "react";
 import {FetchBaseQueryError} from "@reduxjs/toolkit/dist/query/fetchBaseQuery";
 import {SerializedError} from "@reduxjs/toolkit";
 import {IsaacSpinner} from "./IsaacSpinner";
-import {isDefined} from "../../services";
+import {isDefined, isFound, SITE_SUBJECT_TITLE, WEBMASTER_EMAIL} from "../../services";
+import {getRTKQueryErrorMessage} from "../../state";
+import {Alert} from "reactstrap";
+import {NOT_FOUND_TYPE} from "../../../IsaacAppTypes";
 
 const loadingPlaceholder = <div className="w-100 text-center pb-2">
     <h2 aria-hidden="true" className="pt-5">Loading...</h2>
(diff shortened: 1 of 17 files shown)


src/app/services/assignments.ts
```diff
@@ -14,8 +14,8 @@ export const filterAssignmentsByStatus = (assignments: AssignmentDTO[] | undefin
     const fourWeeksAgo = new Date(now.valueOf() - (4 * 7 * 24 * 60 * 60 * 1000));
     // Midnight five days ago:
     const fiveDaysAgo = new Date(now);
-    fiveDaysAgo.setDate(now.getDate() - 5);
-    fiveDaysAgo.setHours(0, 0, 0, 0);
+    fiveDaysAgo.setUTCDate(now.getDate() - 5);
+    fiveDaysAgo.setUTCHours(0, 0, 0, 0);
```
Morpheu5, line 18: There are other similar changes and the whole PR seems to adopt UTC for hours. It may be fine but I wonder if this is going to have weird effects with the fact that we schedule assignments at a certain time in the timezone of the browser instead of UTC.

src/app/components/handlers/ShowLoadingQuery.tsx
```diff
@@ -1,32 +1,80 @@
...
+    return <Alert color={"warning"}>
+        {title ?? "Error fetching data from server"}: {errorDetails.message}
+        {errorDetails.status ? <><br/>Status code: {errorDetails.status}</> : ""}
+        <br/>
+        You may want to refresh the page, or <a href={`mailto:${WEBMASTER_EMAIL}`}>email</a> us if
+        this continues to happen.
+        Please include in your email the name and email associated with this
+        Isaac {SITE_SUBJECT_TITLE} account, alongside the details of the error given above.
+    </Alert>;
+};
```
Morpheu5, line 26: I'm assuming this (and related changes) isn't strictly related to this PR. Any chance these changes can be separated out into another PR? OK if not, it just feels a bit unclean, but I think we can survive.

Morpheu5 (commented):
Nice job overall, the UI looks very neat and seems to do what it's supposed to do. I'm just going to leave those two comments there.

src/app/services/assignments.ts
chrisjpurdy: Sorted this 👍

Morpheu5 (approved):

src/app/components/handlers/ShowLoadingQuery.tsx
chrisjpurdy: This is not _strictly_ related to this PR, but I had to build it in order to show errors and loading status for the assignment schedule in a clean way. I'd rather not remove it since I'd end up having to reimplement a version of it in `AssignmentSchedule` anyway

PR #643, opened 2022-09-23
Lazy loading outdated chunks causing errors

chrisjpurdy:
This seems to fix the problem, but I think it only works in the case that the main chunk (and whatever chunk(s) fetches the outdated chunk(s)) changes as well.

To test this, we need to: 
 - bring this branch up on `test`
 - login to the test site in a browser and navigate to the homepage (don't visit My Progress)
 - change something in `MyProgress.tsx`
 - build and deploy to `test` again
 - without reloading the tab, navigate to My Progress
 
The page should show a "Loading..." message and reload, and either My Progress should render as expected, or the site goes in an infinite loop of reloading and the thing that I mentioned above is the case.

This needs to be done with an actual nginx instance (and not just with a python server locally) so all the assets are given the same cache control headers as on live.


mlt47 (approved):
I have tested this by deploying a change to the my progress page test.isaacphysics.org and the reload was surprisingly smooth.

PR #652, opened 2022-10-07
Implementation of cloze questions with dnd-kit

chrisjpurdy:
With a custom autoscroller because the built-in one was buggy.

Things this PR fixes/adds:

- Grid view of items instead of single scrolling line
- Dragging items between inline drop-zones
- All printing issues (to date)
- Better screenreader instructions and support


package.json
```diff
@@ -8,6 +8,9 @@
   },
   "engineStrict": true,
   "dependencies": {
+    "@dnd-kit/core": "^6.0.5",
+    "@dnd-kit/sortable": "^7.0.1",
+    "@dnd-kit/utilities": "^3.2.0",
```
mlt47, line 13: Can probably also remove the old dnd kit here :)

mlt47 (approved):
The code looks to be fine to me. Perhaps you need to write more code to achieve the same thing with this library but the extensibility is ultimately a good thing as hopefully we won't bump into as many walls as compared to the last one. The library does have only about 20% of the old library's weekly downloads but the license is permissable and although it has been a busy week for you, a week to undo the library choice is manageable making the decision relatively easily reversible.

I'm sure the content teams are going to be super glad that you've fixed the bugs that they reported! Bring it up for them to trial and merge it if they are happy.

src/app/components/content/IsaacClozeQuestion.tsx
```diff
@@ -30,26 +61,128 @@ const NULL_CLOZE_ITEM: ItemDTO = {
...
-    const [borderMap, setBorderMap] = useState<{[dropId: string]: boolean}>({});
-
     const [inlineDropValues, setInlineDropValues] = useState<(Immutable<ClozeItemDTO> | undefined)[]>(() => currentAttempt?.items || []);
     // Whenever the inlineDropValues change or a drop region is added, computes a map from drop region id -> drop region value
     const inlineDropValueMap = useMemo(() => Array.from(registeredDropRegionIDs.entries()).reduce((dict, [dropId, i]) => Object.assign(dict, {[dropId]: inlineDropValues[i]}), {}), [inlineDropValues]);
 
+    // Compute map used to highlight each inline drop-zone with whether it is correct or not
+    const itemsCorrect = validationResponse?.itemsCorrect;
+    const dropZoneValidationMap = useMemo<{[p: string]: boolean | undefined}>(() => isDefined(itemsCorrect)
+        ? Array.from(registeredDropRegionIDs.entries()).reduce((dict, [dropId, i]) => Object.assign(dict, {[dropId]: itemsCorrect.at(i)}), {})
```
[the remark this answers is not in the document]
chrisjpurdy: Yeah that was for readability - we use spread operators elsewhere so I thought it would be marginally better practice 🤷

package.json
chrisjpurdy: We do need that actually - it's the latest version of the utilities module which we use in one place to parse some object the core library creates into CSS.

chrisjpurdy: Their versioning is a little wacky 😄

mlt47: I forgot that we're still using this for Parsons and Reorder questions.

PR #653, opened 2022-10-10
Upgrade KaTeX to 0.16.2

chrisjpurdy:
Fixes matrix determinant rendering issues: `/concepts/cm_matrices_determinants` is a great example page to compare before and after.

Having read the changelogs for the [recent KaTeX releases](https://github.com/KaTeX/KaTeX/releases) it doesn't look like there are any breaking changes for us.


mwtrew (approved):
Nice, that did it.

PR #655, opened 2022-10-17
Turn `InequalityModal` into a functional component 

chrisjpurdy:
This is useful for future maintenance and updates, bringing this code up to date with the rest of the codebase. It is also a necessary prerequisite for any hook-based management of active modals. 

I've tried to separate out large pieces of logic from rendering as much as possible, break off menu-related components from the main `InequalityModal` component, and add some types for easier future modification if needed.


src/app/components/elements/modals/inequality/utils.ts
```diff
@@ -0,0 +1,690 @@
...
+    editorMode?: EditorMode;
+    logicSyntax?: LogicSyntax;
+    inequalityModalRef: React.RefObject<HTMLDivElement>;
+    isTrashActive: React.MutableRefObject<boolean>;
+    sketch: React.MutableRefObject<Nullable<Inequality>>;
+    initialEditorSymbols: { type: string; properties: any }[];
+    onEditorStateChange?: (state: any) => void;
+    setEditorState: (state: any) => void;
+}
+export function prepareInequality({editorMode, inequalityModalRef, initialEditorSymbols, isTrashActive, sketch, logicSyntax, setEditorState, onEditorStateChange}: PrepareInequalityArgs) {
```
Morpheu5, line 647: This one seems to be called when closing the modal as well as when opening it, and it creates a new instance of Inequality.

Morpheu5 (commented):
In trying to solve a performance issue, I found that we end up with multiple, mostly unintentionally so, instances of Inequality, most of which aren't really in use and by all accounts should be garbage-collected, and yet they aren't.

I don't think this is specifically a problem with the new functional style because something similar happened when the modal was a class-style component, but it does feel like it got a little bit worse. It **could** be an issue with React's references across multiple re-renders not handling replacement/nullification correctly.

One thing is for sure: `prepareInequality` appears to be called upon closing the modal too, and that does create a new instance of Inequality for no reason.

Something I **can** do is prevent the `draw()` loop from running when Inequality is in headless mode for text entry. That should mitigate the problem a little bit, but it doesn't look like the headless instance is necessarily the problem (oceans, drops) although it does contribute. There'll be a new Inequality release for this at some point soon.

NOTE: There's a `[remove()](https://p5js.org/reference/#/p5/remove)` method in p5 that might help, but I couldn't get it to work.

src/app/components/elements/modals/inequality/utils.ts
chrisjpurdy: On the other hand, I've added that p5 `remove()` method in the cleanup for `prepareInequality` and it seems to have helped...?

PR #658, opened 2022-10-21
Display event cancelled status

chrisjpurdy:
This makes it clear that events are cancelled, and stops users from booking cancelled events through the UI.

### DIFF
--- a/src/app/components/pages/EventDetails.tsx
+++ b/src/app/components/pages/EventDetails.tsx
@@ -1,5 +1,6 @@
 import React, {useEffect, useState} from "react";
 import * as RS from "reactstrap";
+import {Alert} from "reactstrap";
 import dayjs from "dayjs";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
 import {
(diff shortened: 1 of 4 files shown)


src/app/components/pages/EventDetails.tsx
```diff
@@ -234,7 +236,7 @@ const EventDetails = ({match: {params: {eventId}}, location: {pathname}}: EventD
                                                 }
                                             </td>
                                         </tr>}
-                                    {event.bookingDeadline &&
+                                    {(isEventLeader(user) || isAdminOrEventManager(user) || event.isNotClosed) && event.bookingDeadline &&
```
jsharkey13, line 239: What is this change for? Won't it prevent the "deadline has passed" message that's inside it from ever showing to ordinary users?

chrisjpurdy: I guess it depends on when events are marked as "closed" and/or "expired". Maybe `event.isNotClosed` should be replaced with `!event.isCancelled`?
The intention was to hide irrelevant data from normal users if the event is cancelled (and closed as well?) to prevent cluttering, but still show it to event leaders and above.

chrisjpurdy: I did that replacement in the most recent commit, to keep previous behaviour intact.

PR #660, opened 2022-10-26
Avoid capturing Delta, delta, and d as differentials if they aren't followed by variables

Morpheu5:
This avoids the case where you couldn't have `Delta`, `delta`, and `d` as individual letters instead of differentials without argument.


chrisjpurdy (approved):
Tested it locally and works as expected! 👍

PR #662, opened 2022-10-27
Cloze question item alt text

chrisjpurdy:
Add `altText` to items and use these for better screenreader messages in cloze questions. 

Alt text is preferred over item value, and if there is no text and no item value then default screenreader text will be displayed to screenreaders.


mwtrew (approved):
Alt text changes look good and work as intended.

PR #665, opened 2022-10-28
Log "view github code" event when that button is clicked on code snippets

chrisjpurdy:
I pushed the addition of the new log type in the API straight to master (it was just adding another value to an enum)


mlt47 (approved):
LGTM 👍

PR #666, opened 2022-10-28
Make user objects immutable everywhere

chrisjpurdy:
This should eventually be done for the entire Redux state. 

This refactor caught a possible accident waiting to happen in `SchoolInput` where a prop user object had an attribute mutated directly.


mlt47 (approved):
Nice!

PR #667, opened 2022-10-31
Fix Parsons indentation using mouse drag

chrisjpurdy:
Just a case of re-registering mouse/key event listeners when their deps change, and unregistering old ones.

This also fixes the issue of the current attempt not updating when indentation is changed (but I'm not 100% certain why)


mlt47 (approved):
The changes, in themselves, look sensible and it does seem to fix both problems.

PR #671, opened 2022-11-04
RTK Query questions and concepts refactor

chrisjpurdy:
~~Currently having issues with quiz attempts~~ It seems that was just because I didn't have the symbolic logic checker running. 
~~Also, rate limiting questions doesn't seem to work as it used to~~ Another non-issue that turned out to be fine after more manual testing.

I need to write tests to replace those for the old question reducer. There seems to be some quite major performance issues locally, but that might be because immutability and serializability middleware take a while to process large actions (like those produced by querying documents).


chrisjpurdy:
Closing since I'll probably do this better now and there are wayyyyy to many merge conflicts

PR #679, opened 2022-12-07
Stop cloze item dropping into off-screen drop-zone causing entire screen to offset

chrisjpurdy:
There is an easy fix on newer browsers which consists of setting `overflow: clip` for the root div on both sites.

For all other browsers, we need to set `scrollLeft = 0` whenever the root div tries to scroll. This plays nicely with the fix for newer browsers and is fairly lightweight and self-contained. 

If the browser doesn't have `overflow: clip` support, there is a small discrepancy between where an item is dropped and where its drop animation indicates it will be dropped, but this only happens where the bug would occur previously and is a big improvement.

### DIFF
--- a/src/app/components/content/IsaacClozeQuestion.tsx
+++ b/src/app/components/content/IsaacClozeQuestion.tsx
@@ -186,7 +186,7 @@ const IsaacClozeQuestion = ({doc, questionId, readonly, validationResponse}: Isa
         return () => {
             root?.removeEventListener("scroll", resetScrollLeft);
         };
-    });
+    }, []);
 
     useEffect(function updateStateOnCurrentAttemptChange() {
         if (currentAttempt?.items) {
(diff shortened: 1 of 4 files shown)


src/app/components/content/IsaacClozeQuestion.tsx
```diff
@@ -175,6 +175,19 @@ const IsaacClozeQuestion = ({doc, questionId, readonly, validationResponse}: Isa
...
+    // This is an issue with how dnd-kit manages scroll contexts when `overflow: hidden`, can be fixed by setting
+    // `overflow: clip` but only very modern browsers support that.
+    useEffect(() => {
+        const root = document.querySelector<HTMLDivElement>("#root");
+        const resetScrollLeft = () => {if (root) {root.scrollLeft = 0;}};
+        root?.addEventListener("scroll", resetScrollLeft);
+        return () => {
+            root?.removeEventListener("scroll", resetScrollLeft);
+        };
+    });
```
mlt47, line 189: Is not having a list of dependencies for the useEffect here intentional?

mlt47: No worries!

mlt47 (approved):

PR #680, opened 2022-12-07
Order gameboards client-side if we have all of them cached

chrisjpurdy:


mlt47 (approved):
Looks good to me. I was sure we did the front-end sorting in the past but it hasn't been done since at least the useGameboards refactor. Seemed to work correctly for me this morning.

PR #682, opened 2022-12-09
Fix question finder loading infinitely on shuffle and reload

chrisjpurdy:
Fixed by not unsubscribing from the gameboard-fetching promise. I can't believe I missed this before... it doesn't look like this causes a ridiculous number of boards to be stored in Redux, so we can just remove the `unsubscribe` call.


jsharkey13 (approved):
Your explanation of this seems solid!

PR #684, opened 2022-12-12
Use subject colour in test page UI

chrisjpurdy:
This makes the buttons etc. in the test attempt pages use the subject colour of the test. 
It defaults to physics if no subject tag is present - the subject tag needs to be added at the quiz DTO level.

As a side effect, this PR stops Physics UI elements from using the `phy_extra_force_yellow` colour, and makes them use `a11y_green` instead.

### DIFF
--- a/src/app/services/quiz.ts
+++ b/src/app/services/quiz.ts
@@ -1,6 +1,6 @@
 import {useEffect, useMemo} from "react";
 import {deregisterQuestions, registerQuestions, selectors, useAppDispatch, useAppSelector} from "../state";
-import {API_PATH, isDefined, isQuestion} from "./";
+import {API_PATH, isDefined, isQuestion, tags} from "./";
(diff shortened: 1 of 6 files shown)


src/app/services/quiz.ts
```diff
@@ -65,6 +65,8 @@ export function useCurrentQuizAttempt(studentId?: number) {
     const [attempt, error] = studentId
         ? [studentAttempt, studentError]
         : [currentUserAttempt, currentUserError];
+    // Augment quiz object with subject id
+    const attemptWithQuizSubject = {...attempt, quiz: attempt?.quiz && tags.augmentDocWithSubject(attempt.quiz)}
```
mlt47, line 69:
```suggestion
    const attemptWithQuizSubject = {...attempt, quiz: attempt?.quiz && tags.augmentDocWithSubject(attempt.quiz)};
```

mwtrew (approved):

PR #690, opened 2022-12-16
Admin user misuse statistics

chrisjpurdy:
Adds an admin interface for inspecting user site misuse statistics. Also allows the admin to reset counters for each misuse event type. 

This will be helpful in particular if content managers are attempting to do mass operations like adding loads of group managers to groups in a short time period etc., because it means we can just reset the misuse events through this UI instead of faffing around with swagger or curl.


src/app/state/slices/api/index.ts
```diff
@@ -19,7 +19,7 @@ import {
     GameboardListDTO,
     IsaacConceptPageDTO,
     IsaacPodDTO,
-    IsaacWildcard,
+    IsaacWildcard, MisuseStatisticDTO,
```
jsharkey13, line 22: Shouldn't this be on a line by itself?

src/app/components/pages/Admin.tsx
```diff
@@ -129,8 +130,13 @@ export const Admin = ({user}: {user: RegisteredUserDTO}) => {
                         }
                     </React.Fragment>}
 
-                    <h3 className={"mt-2"}>Demonstration Mode</h3>
+                    <h3 className={"mt-3"}>Demonstration Mode</h3>
                     <AnonymiseUsersCheckbox/>
+
+                    {isAdmin(user) && <>
+                        <h3 className={"mt-3"}>Misuse statistics</h3>
+                        <MisuseStats/>
```
jsharkey13, line 138: Is cramming this all onto the main admin page really a good idea? It'll mean lots of work to load the stats every time someone wants a content push. Does it try (and fail) to make the request for non-admin users, or does this conditional prevent that at least?

jsharkey13 (commented):
The merge conflict needs resolving, too.

src/app/components/pages/Admin.tsx
chrisjpurdy: The conditional prevents the `MisuseStats` component from rendering at all, and since that component does the data fetching, it won't make the request if the user isn't an admin. 

I think having all on the same page is fine, because IMO there isn't really that much on the page already 🤷 we can move it elsewhere if it causes any issues, but in principle not many users will access this page.

PR #691, opened 2023-01-05
Feature/clearer gameboard staged difficulties labelling

mlt47:
This replaces the aggregate difficulty icons with a table where each row contains a stage and a list of difficulties in short label form (i.e. "P1", "C3").
The old aggregate difficulty icons were confusing as they were similar to the difficulty icons shown for individual questions but the shading rules were subtly different. It also, in the case of a gameboard with one question labelled as GCSE: Challenge 3 and A Level: Practice 1, would not be able to distinguish if the gameboard contained any challenging A level questions.

The new implementation solves this through the use of tables.
The use of tables is usually good for accessibility but in the table view of gameboards/set assignments this solution involves using a nested table, they are generally not as great for accessibility/screen readers... If anyone has an alternative solution I would be happy to try it. My first implementation used `rowspan` (`rowSpan` property in react) but it proved quite difficult to get the css to do what I wanted it to do as we explicitly set the height of each `<tr>`.

There is perhaps scope for code reuse for the very similar tables used here but I thought it would be worth waiting until the broader task of reusing components between Set Assignments and My Gameboards which are very similar pages.

### DIFF
--- a/src/app/components/elements/Assignments.tsx
+++ b/src/app/components/elements/Assignments.tsx
@@ -1,32 +1,29 @@
-import {AssignmentDTO, Difficulty, Stage} from "../../../IsaacApiTypes";
-import React, {MouseEvent} from "react";
+import {AssignmentDTO} from "../../../IsaacApiTypes";
+import React, {MouseEvent, useMemo} from "react";
 import {Col, Row} from "reactstrap";
 import {Link} from "react-router-dom";
 import {
-    allPropertiesFromAGameboard,
+    determineGameboardStagesAndDifficulties,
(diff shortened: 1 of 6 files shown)


chrisjpurdy (approved):
Seems good! 👍 I kind of wish there was a more visual way of representing this but the information is complicated so this approach is understandable.

Only other thing I noticed is that, in table view, the hexagon is not vertically centred, but the text/buttons in every other row are. This is only noticeable if the stage/difficulty list is long enough to cause the row to expand vertically. It's probably not worth fixing before merging this I think!

PR #694, opened 2023-01-13
Allow deploy script to execute commands itself (after prompting user)

chrisjpurdy:
Also adds commands that check for "penultimate" api version(s), new db migrations, and new config changes.


mlt47 (approved):
My careful reading of this looks like it should work. I guess we'll test it and see!

PR #695, opened 2023-01-16
De-emphasise scheduled assignments in assignment progress

chrisjpurdy:
To make it clear which assignments are live.

image

### DIFF
--- a/src/app/components/pages/AssignmentProgress.tsx
+++ b/src/app/components/pages/AssignmentProgress.tsx
@@ -353,7 +353,13 @@ const AssignmentDetails = ({assignment}: {assignment: EnhancedAssignment}) => {
             <Button color="link" className="gameboard-title align-items-center" onClick={() => setIsExpanded(!isExpanded)}>
                 <span className={classNames({"text-muted": assignmentHasNotStarted})}>
                     {assignment.gameboard?.title}
-                    {assignment.dueDate && <span className="gameboard-due-date">({assignmentHasNotStarted ? <>Scheduled:&nbsp;{formatDate(getAssignmentStartDate(assignment))},{" "}</> : ""}Due:&nbsp;{formatDate(assignment.dueDate)})</span>}
+                    {assignmentHasNotStarted && <span className="gameboard-due-date">
+                        (Scheduled:&nbsp;{formatDate(getAssignmentStartDate(assignment))})
+                    </span>}
+                    {assignmentHasNotStarted && assignment.dueDate && " "}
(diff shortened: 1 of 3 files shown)


src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -343,13 +346,18 @@ const AssignmentDetails = ({assignment}: {assignment: EnhancedAssignment}) => {
...
+    const assignmentHasNotStarted = !hasAssignmentStarted(assignment);
+
     return <div className="assignment-progress-gameboard" key={assignment.gameboardId}>
-        <div className="gameboard-header" onClick={() => setIsExpanded(!isExpanded)}>
+        <div className={classNames("gameboard-header", {"text-muted": assignmentHasNotStarted})} onClick={() => setIsExpanded(!isExpanded)}>
             <Button color="link" className="gameboard-title align-items-center" onClick={() => setIsExpanded(!isExpanded)}>
-                <span>{assignment.gameboard?.title}{assignment.dueDate && <span className="gameboard-due-date">(Due:&nbsp;{formatDate(assignment.dueDate)})</span>}</span>
+                <span className={classNames({"text-muted": assignmentHasNotStarted})}>
+                    {assignment.gameboard?.title}
+                    {assignment.dueDate && <span className="gameboard-due-date">({assignmentHasNotStarted ? <>Scheduled:&nbsp;{formatDate(getAssignmentStartDate(assignment))},{" "}</> : ""}Due:&nbsp;{formatDate(assignment.dueDate)})</span>}
```
mlt47, line 356: If there is an assignment start date but no due date (I agree probably a strange case) there will not be an indication for why the assignment is de-emphasised.

I am between two minds as to whether we should disable the `setIsExpanded` on clicking the containing button. Viewing the progress of an assignment which has not yet been assignment doesn't really make sense but it would give them the opportunity to spot if most of their students have already answered the question in the past, so probably leave that as is for now.

mlt47 (approved):
I'm happy with this except for the minor edge case I have added as a comment.

src/app/components/pages/AssignmentProgress.tsx
chrisjpurdy: Ahhh that's a good point, surprised I missed that 🙄 
WRT disabling the expansion I'd rather not for now since that is how it is currently, but James also reckons that we should do this as it would be arguably more consistent with our privacy policy. The card is marked for review to discuss this point with Lisa in the tech meeting.

PR #699, opened 2023-01-26
Add "Reset" button to Inequality 

chrisjpurdy:
Design choices:
- Reset button only shows if the question has an editor seed value
- Button immediately resets, without prompting the user ("are you sure?")
- The icon (below)

image

~~I am happy to [update the help modal](https://editor.isaacphysics.org/edit/master/content/general_pages/eqn_editor_help_modal_chemistry.json) (add an extra step and make a new PNG with number labels etc.) once I get some feedback on design choices.~~ Lee is happy with the design currently and says that he'll update the help modal! (we should let him know when this is merged)

---

I personally prefer no prompt, because I doubt that much work will be lost in practice and students will just learn what it does. In the long run undo/redo functions would be great, and alleviate this issue, but they're way out of scope at the moment. 

For the record, here's an example of the button with a popup, allowing the user to reconsider in case they pressed the reset button by accident:
image


src/app/components/elements/modals/inequality/InequalityModal.tsx
```diff
@@ -480,7 +487,12 @@ const InequalityModal = ({availableSymbols, logicSyntax, editorMode, close, onEd
             Help
         </div>
 
-        <div id="inequality-trash" className="inequality-ui trash button">Trash</div>
+        {isDefined(editorSeed) && <div id="inequality-reset" className="inequality-ui reset button"
+             role="button" tabIndex={-1} onClick={resetToInitialState} onKeyUp={resetToInitialState}
```
mlt47, line 491: What is the reason for the button not to be tab indexable?

mlt47 (changes requested):
Cool. I've included a small suggestion but the main feedback is that the content team would also like this button to be visible to general equation questions, not just chemistry questions. Other than that I'm happy for this to be merged.

src/app/components/elements/modals/inequality/InequalityModal.tsx
chrisjpurdy: I don't know, but I noticed that the others weren't either so assumed it was best to add 🤷

PR #704, opened 2023-02-08
Run Google Analytics 4 alongside current Universal Analytics

chrisjpurdy:
This just runs both versions of ReactGA alongside one another. 

The new version is a bit crusty but does the job fine (it's essentially just a light wrapper around `gtag.js`) - by this I mean the convenience methods `pageview` and `exception` now no longer exist but you can just use the base `gtag` method to send GA the same message.

I've created a GA4 "property" for each site - the measurement ids in this PR are not placeholders.

### DIFF
--- a/src/app/components/navigation/TrackedRoute.tsx
+++ b/src/app/components/navigation/TrackedRoute.tsx
@@ -1,11 +1,13 @@
 import React, {useEffect} from "react";
 import {Redirect, Route, RouteComponentProps, RouteProps} from "react-router";
 import ReactGA, {FieldsObject} from "react-ga";
+import ReactGA4 from "react-ga4";
 import {FigureNumberingContext, PotentialUser} from "../../../IsaacAppTypes";
 import {ShowLoading} from "../handlers/ShowLoading";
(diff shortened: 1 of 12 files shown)


src/app/components/navigation/TrackedRoute.tsx
```diff
@@ -15,11 +17,15 @@ import {Unauthorised} from "../pages/Unauthorised";
...
 ReactGA.initialize(GOOGLE_ANALYTICS_ACCOUNT_ID);
+ReactGA4.initialize(GOOGLE_ANALYTICS_4_MEASUREMENT_ID);
 ReactGA.set({ anonymizeIp: true });
+ReactGA4.set({ anonymizeIp: true });
 
 const trackPage = (page: string, options?: FieldsObject) => {
     ReactGA.set({ page, ...options });
+    ReactGA4.set({ page, ...options });
     ReactGA.pageview(page);
+    ReactGA4.send({ hitType: "pageview", page });
```
jsharkey13, line 28: How is this a valid JS object? If `page` is a string, does this automagically do `page: page`?

jsharkey13 (approved):

PR #705, opened 2023-02-23
Wrap month and year properly when setting day in date inputs

chrisjpurdy:
New behaviour: 
- If the day is 09/12/2023 and you input 5 as the day, the date input defaults to 05/01/2024
- If the day is 09/05/2023 and you input 5 as the day, the date input defaults to 05/06/2023
- If the day is 01/09/2023 and you input 5 as the month, the date input defaults to --/05/2024

This was harder than we thought! IMO we need a new date input method (I actually have a "nicer" one in a branch somewhere) because rolling your own is tricky.

Would benefit from someone testing it again locally

### DIFF
--- a/src/app/components/elements/inputs/DateInput.tsx
+++ b/src/app/components/elements/inputs/DateInput.tsx
@@ -5,8 +5,7 @@ import {range} from 'lodash';
 // @ts-ignore This value definition is a bit dodgy but should work.
 export interface DateInputProps extends InputProps {
     labelSuffix?: string;
-    defaultYear?: number;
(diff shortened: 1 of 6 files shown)


mlt47 (approved):
This looks OK. Perhaps accepting a minDate as a prop would be more generic in that it would allow you to run this logic for assignment due date needing to be greater than the assignment scheduled date rather than just greater than "now". That would require a little more restructuring of the component as there is overlap in duties with the yearRange prop.
As there is a good chance we move away from this date input in the future it is probably not worth the effort now, especially while there are higher priority tasks.

PR #709, opened 2023-03-28
Merge `the-analytical-engine/main`

chrisjpurdy:
Conflicts:

- package.json - version number, I kept the one from `the-analytical-engine` (3.0.3)
- src/app/components/navigation/WarningBanner.tsx - just the Isaac CS downtime banner, I changed the date range to end yesterday
- src/app/components/site/cs/HomepageCS.tsx - just some imports 

I had to fix the tests (which luckily are pretty solid and failed for understandable reasons), and make sure that `highlightjs-vba` is transformed when the tests are run (because it contains some ES6 syntax).

Changes I made: https://github.com/isaacphysics/isaac-react-app/compare/cc4ea1e320aeadec16ddc1d93a6ac3656ac601d1..3d92b9db96482d5fd9b6b039891d3ec95e7d3694


src/app/services/highlightJs.ts
```diff
@@ -4,11 +4,16 @@ import php from 'highlight.js/lib/languages/php'
 import csharp from 'highlight.js/lib/languages/csharp'
 import plaintext from 'highlight.js/lib/languages/plaintext'
 import sql from 'highlight.js/lib/languages/sql'
+import java from 'highlight.js/lib/languages/java'
 import {LanguageFn} from "highlight.js";
+const vba = require("highlightjs-vba");
```
mlt47, line 9: Why doesn't the ES6 Module Import not work here?

mlt47 (commented):
Thanks for also fixing the tests to work with Ada, that's great.
It is not clear to me why the highlight.js vba module needed importing via require rather than an ES6 import, what was it that made this necessary?

src/app/services/highlightJs.ts
chrisjpurdy: It was just that `highlightjs-vba` doesn't have it's own module definition, so Matthew T had included a `declare module "highlightjs-vba"` in `react-app-env.tsx`.
I wasn't sure whether that had something to do with the fact that the tests we're failing (there were errors about this package) so I changed it to this, but it turned out to be an issue with out Jest webpack config.  Looking at `react-app-env.tsx` again we do the same thing for `math-random-seed` so I'll change it back.

jsharkey13: But the "library" to define the language is super simple: https://github.com/dullin/highlightjs-vba/blob/master/src/vba.js
We could define it ourselves by copying (with attribution!) that code into the file, in the same way I declare the pseudocode language, which feels more flexible and much simpler!

chrisjpurdy: Up to you 🤷 I don't see any issue with importing it like this, but I suppose we'll know for sure that code-splitting etc. will work if we add it into the codebase directly.

jsharkey13: I have an even better fix for this, just waiting to confirm that the language on the site is actually Visual Basic .NET (rebranded as "Visual Basic" by Microsoft but not the original Visual Basic language). If so, that's built in to HighlightJS since it's very common and we don't need any of this 🙃

chrisjpurdy:
I added the code font-size thing straight to this - it's a small visual tweak and I would have committed it to `master` if I could.

build.sh
```diff
@@ -28,14 +28,14 @@ fi
...
+echo Building Ada CS and Physics apps in $BUILD_DIR: $VERSION_TO_DEPLOY
 
 rm -rf $BUILD_DIR
 mkdir -p $BUILD_DIR
 cd $BUILD_DIR
 
-git clone -b $VERSION_TO_DEPLOY --depth 1 https://github.com/isaacphysics/isaac-react-app.git
-cd isaac-react-app
+git clone -b $VERSION_TO_DEPLOY --depth 1 https://github.com/isaacphysics/the-analytical-engine.git
+cd the-analytical-engine
```
jsharkey13, line 38: These are actually incorrect changes, will undo this now.

build.sh
```diff
@@ -57,40 +57,40 @@ else
...
   docker pull "docker.isaacscience.org/isaac-phy-app:${VERSION_TO_DEPLOY}"
-  docker pull "docker.isaacscience.org/isaac-cs-app-renderer:${VERSION_TO_DEPLOY}"
+  docker pull "docker.isaacscience.org/isaac-ada-app-renderer:${VERSION_TO_DEPLOY}"
   docker pull "docker.isaacscience.org/isaac-phy-app-renderer:${VERSION_TO_DEPLOY}"
 fi
 
 
 cd ..
-rm -rf isaac-react-app
+rm -rf the-analytical-engine
```
jsharkey13, line 93: Ditto; to undo.

PR #711, opened 2023-03-31
Add group event booking contact form to manage group reservations modal

chrisjpurdy:
Needs copy.

---

The `RESERVED_ONLY` event status displays the following for students:

image

and makes no difference to the teacher view. An event that is both `RESERVED_ONLY` but not group booking doesn't make any sense, but I haven't handled this case because it's fundamentally a content error (and a by-product of the overcomplicated way we categorise events)


mlt47 (approved):
Waiting on copy. Code works and is fine.
Will affect all reservation modals not just reservation-only modals. As it is only Physics that are running events for the foreseeable future that is fine.

PR #721, opened 2023-04-21
Optional group name anonymisation

chrisjpurdy:
If we want even finer grain anonymisation then the `anonymiseIfNeededWith` functions should be rewritten, but this should do for now.

This also gives staff the ability to see group ids by hovering over the group name on the sidebar in "Manage groups"


mlt47 (approved):
Seems fine.
Perhaps all the `persistence.load(KEY.ANONYMISE_GROUPS) === "YES"` could be moved out of index.ts into utils.ts to be slightly cleaner.

mlt47 (approved):

PR #722, opened 2023-04-21
Fix news not loading after logout

chrisjpurdy:
This loads news on the homepage whenever the user object changes, which fixes the issue. 

It seems to be caused by a race condition between the homepage component mounting and fetching the news, and the logout action triggering the store to be cleared. I'm guessing that the RTK query API reducer needs a moment to recover after being set to `undefined` 🤷

### DIFF
--- a/src/app/components/site/cs/HomepageCS.tsx
+++ b/src/app/components/site/cs/HomepageCS.tsx
@@ -14,7 +14,11 @@ import {MetaDescription} from "../../elements/MetaDescription";
 export const HomepageCS = () => {
     useEffect( () => {document.title = SITE_TITLE;}, []);
     const user = useAppSelector(selectors.user.orNull);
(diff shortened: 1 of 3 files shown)


mlt47 (changes requested):
At the stand up we had discussed a hard redirect to "/" using `window.open(...)` as a solution (that we applied in the AngularJS app) for this problem.
That does not work now because the redux state is needed to display the user consistency error modal in any other open tabs on signing out.

This solution fixes the particular case for news but the underlying problem also affects glossary terms and any other similarly loaded content. You can check this by visiting a question's Hint 1 on Ada after logging out, the glossary terms will be continuously "loading".

If you get to the cause of the problem by tracing redux actions there might be a fix that sorts this out for all this type of data. If not we will need to apply similar fixes for the other instances.

Have another think about it in case you can find a nicer solution.

PR #724, opened 2023-04-25
Get groups in SetQuizzes component so augmenting assignments succeeds

mwtrew:


chrisjpurdy (approved):
Seems fine, should be revisited when we move tests/quizzes into RTK query

PR #726, opened 2023-04-28
Refresh on logout to clear the Redux cache entirely

chrisjpurdy:
User consistency errors now cause the user to be hard-redirected to `/consistency-error`, which is just a page version of the consistency modal we had previously.

This fixes the "news and glossary terms not loading after logout" issue.

It does make the code a bit simpler - there's probably a few other pieces of code that aren't strictly required any more (any reducers that handle the logout/user consistency actions for example).


src/app/components/pages/ConsistencyError.tsx
```diff
@@ -0,0 +1,21 @@
...
+        <p className="pb-2">This can happen if you have logged out or logged in as another user via another browser window. Please click continue to avoid any issues.</p>
+        <p>
+            <em>Note:</em>
+            {" If you would like to be logged in with two accounts at one time you will need to use your browser's "}
+            <a href="https://en.wikipedia.org/wiki/Privacy_mode" target="_blank" rel="noopener noreferrer">
+                private browsing
+            </a>
+            {" feature."}
+        </p>
+        <Button color="secondary" href="/login">Continue</Button>
```
jsharkey13, line 19: I wonder if this shouldn't just be the homepage? In case it's someone who doesn't want to log in who lands there?

src/test/pages/Groups.test.tsx
```diff
@@ -41,6 +41,15 @@ const switchGroupsTab = async (activeOrArchived: "active" | "archived", expected
     return groups;
 };
 
+const closeActiveModal = async (modal: HTMLElement) => {
```
jsharkey13, line 44: Is this related at all? :/

src/test/state/userConsistencyChecker.test.tsx
```diff
@@ -27,13 +29,15 @@ describe("userConsistencyCheckerMiddleware", () => {
 
         // @ts-ignore
         setUserId.mockImplementation(() => true);
+        // @ts-ignore
+        redirectTo.mockImplementation(() => true);
```
jsharkey13, line 33: There's a FIXME in `reducers.test.tsx` to test that it does a refresh and clears the state. But if a mock like this is used, will the redirect do anything?

jsharkey13 (commented):
Another nice thing about this is that we will end up with pageview logging of the consistency warning now, too.

src/test/state/userConsistencyChecker.test.tsx
chrisjpurdy: If the redirect (via the `window`) object is done in the tests _at all_ without being mocked, then the tests will crash because `window` doesn't exist (since the tests are done in node, not in the browser). 
So when that test in `reducers.test.tsx` is implemented, this function will need to be mocked - you can check whether a mocked function has been called.

src/test/pages/Groups.test.tsx
chrisjpurdy: This is just a small refactor of the tests when I was fixing them after these changes - I couldn't leave the tests broken 🤷

src/app/components/pages/ConsistencyError.tsx
chrisjpurdy: I guess - I was thinking the same thing, but this is what we had before (see the deleted `ConsistencyErrorModal.tsx` file)  so I thought I'd leave it as is.

src/test/pages/Groups.test.tsx
chrisjpurdy: But yeah I have no idea why these tests broke in particular, since they have nothing to do with the changes

PR #731, opened 2023-05-09
Stop changing of page title if previewing question in gameboard builder

chrisjpurdy:
Adds a `preview` prop that determines whether "page level state modifying changes" happen when rendering a `Concept` or `Question` component.

At the moment, any time a `PageTitle` component is rendered it changes `document.title` - this stops that from happening in "previewed" pages.


mlt47 (approved):
I was worried the prop-drilling would be excessive but it is not, it is useful at every level.

PR #737, opened 2023-05-19
Redesign manage groups UI

chrisjpurdy:
Adds additional managers to the group overview, and moves buttons around to be more intuitive.

See pics for the (medium/large screen) layouts on Ada and physics - it is nicely responsive on small screens.
image
image

### DIFF
--- a/src/app/components/pages/Groups.tsx
+++ b/src/app/components/pages/Groups.tsx
@@ -16,7 +16,7 @@ import {
     Nav,
     NavItem,
     NavLink,
-    Row,
+    Row, Table,
     UncontrolledButtonDropdown,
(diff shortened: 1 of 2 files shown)


jsharkey13:
A question; when you are an additional manager, would it not make sense to also show the owner without having to click "More Info"? Perhaps at the top of the manager list with `(owner)` after their name?

Do we need the email addresses in that list? They make it quite big, and you can see them by clicking "More Info" anyhow?

And is the light grey of `(you)` accessible; it looks quite light on Ada particularly, is there enough contrast?

chrisjpurdy:
Updated:

image

jsharkey13 (approved):

PR #738, opened 2023-05-23
Coordinate questions

chrisjpurdy:
Adds support for coordinate questions to the site:

image

There might be a couple of things that need changing going forwards, but I'd like to merge them as they are currently (modulo code review) so content can test them out and decide what they want from question types like this.


mlt47 (approved):
This looks fine. Good enough to get some feedback on the question type. I feel a request for a 3rd dimension will be likely but we'll see :)

PR #741, opened 2023-06-02
Redirect users to feedback page (if allowed) after quiz submission

chrisjpurdy:
The generifying of `augmentDocWithSubject` was necessary for the types to work out in some other code to do with this PR - I removed that code (it was wrong) but kept the extra typing because it improves type inference and autocompletion if you're working with objects that have passed through that function. Beforehand, calling `augmentDocWithSubject(obj)` erases the type of `obj` to `ContentDTO & {subjectId: string}` (if `obj` is of type `QuizDTO` for example then this is annoying).

### DIFF
--- a/src/app/services/tagsAbstract.ts
+++ b/src/app/services/tagsAbstract.ts
@@ -5,7 +5,7 @@ import {ContentDTO} from "../../IsaacApiTypes";
 export abstract class AbstractBaseTagService {
     abstract getTagHierarchy(): TAG_LEVEL[];
     abstract getBaseTags(): BaseTag[];
-    abstract augmentDocWithSubject(doc: ContentDTO): ContentDTO & {subjectId: string};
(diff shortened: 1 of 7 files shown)


src/app/services/tagsAbstract.ts
```diff
@@ -5,7 +5,7 @@ import {ContentDTO} from "../../IsaacApiTypes";
 export abstract class AbstractBaseTagService {
     abstract getTagHierarchy(): TAG_LEVEL[];
     abstract getBaseTags(): BaseTag[];
-    abstract augmentDocWithSubject(doc: ContentDTO): ContentDTO & {subjectId: string};
+    abstract augmentDocWithSubject<T extends ContentDTO>(doc: T): T & {subjectId: string};
```
mlt47, line 8: 👍

PR #742, opened 2023-06-02
Add SSO buttons to registration page, use signup param for RPF

mwtrew:
Also, make RPF button text more concise when space is limited and adds 'friendly names' for authentication providers.


src/app/components/pages/Registration.tsx
```diff
@@ -349,6 +351,22 @@ export const Registration = withRouter(({location}:  RouteComponentProps<{}, {},
...
+                                <RaspberryPiSignInButton isSignup={true} />
+                            </Col>
+                        </Row>
+                        <Row className={"mb-3 justify-content-center"}>
+                            <Col md={{size: 7}} lg={{size: 5}}>
+                                <GoogleSignInButton />
+                            </Col>
+                        </Row>
+                    </div>
+                }
```
chrisjpurdy, line 369: Just wondering why this doesn't add the "sign up with Google" option to the physics registration page? I can't see any issues with this, since you can already do that via the login page.

chrisjpurdy (approved):
Cool! Just had one question about Physics, other than that seems good 🙌

src/app/components/pages/Registration.tsx
mwtrew: Physics are interested in this after all, so I've added it there too.

PR #745, opened 2023-06-06
Split off related groups of RTKQ endpoints into separate files

chrisjpurdy:
https://github.com/isaacphysics/isaac-react-app/pull/744 should be reviewed and merged first - it takes care of the first two commits.

This (the third commit) also changes how we export the query/mutation hooks to be more robust to endpoints moving around. The code also reads a bit nicer after this change too.

Most of the changes are just replacements of `isaacApi.endpoints.foo.useBar(...)` with `useFooBar(...)` all over the codebase.


src/app/components/navigation/IsaacApp.tsx
```diff
@@ -87,7 +87,7 @@ export const IsaacApp = () => {
     const dispatch = useAppDispatch();
     const serverError = useAppSelector((state: AppState) => state && state.error && state.error.type == "serverError" || false);
     const goneAwayError = useAppSelector((state: AppState) => state && state.error && state.error.type == "goneAwayError" || false);
-    const segueEnvironment = useAppSelector((state: AppState) => state && state.constants && state.constants.segueEnvironment || "unknown");
+    const {data: segueEnvironment} = useGetSegueEnvironmentQuery();
```
mlt47, line 90: This loses the `|| "unknown"` state. Unsure if that is important later on in the code.
If it is, the same behaviour could be achieved by setting a default value on the destructuring/unpacking. (Technically not exactly the same as it is `|| "unknown"` rather than `?? "unknown` but hopefully we weren't operating on falsy values.)

mlt47: Just looked through the file and it is safe not to have this state.

mlt47 (approved):
Assuming you copied and pasted the slices out correctly all other file changes look correct to me. Nice!

PR #746, opened 2023-06-06
RTKQ Admin endpoints

chrisjpurdy:
Doesn't include the email endpoints - I'm going to do those separately.

~~Not going to mark this as ready until the 2 previous RTKQ PRs are sorted.~~

The only complicated part was the user search caching, and the cache being updated correctly by user deletion and modification.

I've added some comprehensive tests for the admin user manager (deletion, changing role and verification status, etc.), so I'm confident about the refactor.


mlt47 (approved):
Looks OK

PR #747, opened 2023-06-07
Graph sketcher curve debug mode

chrisjpurdy:
This adds a debug button to the graph sketcher modal, that when toggled on displays the current graph features.

The PR also fixes a minor logical error where graph sketcher question attempts were dispatched every time the graph is changed (when really they should be dispatched when the modal is closed).

I took this opportunity to move the "generate answer specification" endpoint into RTKQ as well.


src/app/components/elements/modals/GraphSketcherModal.tsx
```diff
@@ -67,15 +92,6 @@ const GraphSketcherModal = (props: GraphSketcherModalProps) => {
...
     }, []);
 
-    useEffect(() => {
-        if (isDefined(modalSketch)) {
-            modalSketch.state = {
-                ...modalSketch.state,
-                curves: initialState?.curves ?? modalSketch.state.curves ?? []
-            };
-        }
-    }, [initialState]);
```
chrisjpurdy, line 77: I removed this because changing the "initial state" should not overwrite the current state of the modal. 
I tested this manually and can't see it breaking anything currently - the only way it would be an issue is if the loading of the users current attempt happens asynchronously after the question page is loaded.

chrisjpurdy:
Deprecated in favour of https://github.com/isaacphysics/isaac-react-app/pull/748 (this one didn't merge nicely after the major RTKQ code splitting changes)

PR #777, opened 2023-08-10
Allow upgrades to teacher accounts without manual approval

mwtrew:
Adds a page to Ada through which users can upgrade themselves to teacher accounts without manual approval.


jsharkey13 (approved):
In the interest of getting this tested by the team, I am going to merge it; but there are some issues about the UX that will need work.
 - There is no redirect after success, just an easily-missed toast pop-up.
 - The page shows the same content if you already have a teacher account, and after a successful upgrade if the page is reloaded.

PR #788, opened 2023-09-01
Return to previous Y position on equation-editor close

mlt47:


src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
```diff
@@ -27,13 +27,15 @@ const IsaacSymbolicChemistryQuestion = ({doc, questionId, readonly}: IsaacQuesti
...
-    };
+    const closeModalAndReturnToScrollPosition = useCallback(function(previousYPosition: number) {
+        return function() {
+            document.body.style.overflow = "initial";
+            setModalVisible(false);
+            if (isDefined(previousYPosition)) {
+                window.scrollTo(0, previousYPosition);
+            }
+        };
+    }(window.scrollY), [modalVisible]);
```
mlt47, line 38: This will produce a callback function for the InequalityModal to use when closing. The produced callback will have the window.scrollY position captured in its closure from whenever the last time the modalVisible parameter changed, no matter how many times this component is re-rendered in the meantime.

jsharkey13 (approved):

PR #790, opened 2023-09-05
Confirm expected API exists for frontend-only release

mwtrew:


deploy.py
```diff
@@ -200,7 +210,7 @@ def deploy_live(ctx):
         ask_to_run_command(f"./compose-live {ctx['site']} {ctx['app']} up -d {ctx['site']}-api-live-{ctx['api']}")
 
         print("# Wait until the api is up:")
-        api_endpoint = f"https://isaac{'computerscience' if ctx['site'] == Site.ADA else 'physics'}.org/api/{ctx['api']}/api/info/segue_environment"
+        api_endpoint = f"https://{'adacomputerscience' if ctx['site'] == Site.ADA else 'isaacphysics'}.org/api/{ctx['api']}/api/info/segue_environment"
```
mlt47, line 213: Good spot!

deploy.py
```diff
@@ -183,8 +183,18 @@ def deploy_live(ctx):
...
+    while response not in ["front-end-only", "n"]:
+        response = input("Please respond with one of:\n - front-end-only \n - n\n").lower()
+    front_end_only_release = response == "front-end-only"
+
+    if front_end_only_release:
+        print("# Front-end-only release - confirm which API this app image expects:")
+        expected_api = ask_to_run_command(f"docker inspect --format '{{{{ index .Config.Labels \"apiVersion\"}}}}' docker.isaacscience.org/isaac-{ctx['site']}-app:{ctx['app']}")
+
+        print("# Front-end-only release - confirm the expected API is running:")
+        ask_to_run_command(f"docker ps --format '{{{{.Names}}}}' | grep {ctx['site']}-api-live-{expected_api}")
```
mlt47, line 196: I see that grep finished with exit code 1 if there are no matches which would then cause `ask_to_run_command(...)` to show the error and ask the deployer if they want to continue or abort. That should be enough of a prompt for the deployer to notice that something isn't quite right.

mlt47 (approved):
👍

PR #791, opened 2023-09-06
Skip queries for set Tests if user is tutor

mlt47:
For tutors we were requesting their tests on the Assignment Progress page even though they cannot set test and this was resulting in an error toast being displayed on each visit.
Using a `skipToken` is the idiomatic way of conditionally querying an endpoint (`skipToken` preferred over `{skip: true}` options for easier typing): https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching.

### DIFF
--- a/src/app/components/pages/AssignmentProgress.tsx
+++ b/src/app/components/pages/AssignmentProgress.tsx
@@ -493,11 +493,11 @@ const QuizDetails = ({quizAssignment}: { quizAssignment: QuizAssignmentDTO }) =>
     </div> : null;
 };
 
-const GroupDetails = ({group}: {group: AppGroup}) => {
+const GroupDetails = ({group, user}: {group: AppGroup, user: RegisteredUserDTO}) => {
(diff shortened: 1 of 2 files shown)


src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -493,11 +493,11 @@ const QuizDetails = ({quizAssignment}: { quizAssignment: QuizAssignmentDTO }) =>
     </div> : null;
 };
 
-const GroupDetails = ({group}: {group: AppGroup}) => {
+const GroupDetails = ({group, user}: {group: AppGroup, user: RegisteredUserDTO}) => {
```
mlt47, line 496: Could `useSelector(...)` in the `useGroupAssignments(...)` hook to grab the user from the redux store rather than "prop-drill" the user down each component but this way avoids needing to check that the user is signed in so I decided to choose that option.

jsharkey13 (approved):

PR #792, opened 2023-09-12
Hotfix/mobile friendly graph sketcher

mlt47:
This alters the Graph Sketcher modal to behave like the Equation Editor modal - it fills the viewport and stops the touch-action from scrolling the page in any direction. It also records the scroll y position on modal open so that it can return to it on modal close.
This has been tested on a Android Pixel and iPhone where graphs were sketchable, hopefully it works for other smaller devices too. I suggest we release and ask for further feedback from the content teams and their devices.

### DIFF
--- a/src/app/components/content/IsaacGraphSketcherQuestion.tsx
+++ b/src/app/components/content/IsaacGraphSketcherQuestion.tsx
@@ -1,4 +1,4 @@
-import React, {useState, useEffect, useRef} from "react";
+import React, {useState, useEffect, useRef, useCallback} from "react";
 import {GraphChoiceDTO, IsaacGraphSketcherQuestionDTO} from "../../../IsaacApiTypes";
 import GraphSketcherModal from "../elements/modals/GraphSketcherModal";
 import {GraphSketcher, makeGraphSketcher, LineType, GraphSketcherState} from "isaac-graph-sketcher";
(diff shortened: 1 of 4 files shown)


src/app/components/content/IsaacGraphSketcherQuestion.tsx
```diff
@@ -29,8 +29,16 @@ const IsaacGraphSketcherQuestion = ({doc, questionId, readonly}: IsaacQuestionPr
         !readonly && setModalVisible(true);
     }
 
+    const returnToScrollYPosition = useCallback(function(previousYPosition: number) {
+        return function() {
+            document.body.style.overflow = "initial";
+            window.scrollTo(0, previousYPosition);
+        }
+    }(window.scrollY), [modalVisible]); // Capture y position whenever modalVisible changes.
```
mlt47, line 37: The useCallback and immediately invoked function expression is a little strange but it was the neatest way I could think of for capturing the `window.scrollY` value at the time that we need to store it, without it getting overwritten on re-renders when the modal is full-screen and `window.scrollY` is `0`.

src/app/components/elements/modals/inequality/utils.ts
```diff
@@ -367,6 +367,28 @@ export const generateDefaultMenuItems = (parsedAvailableSymbols: string[], logic
 
 // --- Callback/effect definitions ---
 
+export function fillScreenWithModal(fullScreen: boolean) {
+    if (fullScreen) {
+        document.documentElement.style.overflow = "hidden";
+        document.documentElement.style.width = '100vw';
+        document.documentElement.style.height = '100vh';
```
mlt47, line 374: There is a `dvh` for device view height and the same for width that we might choose to prefer once more browsers can support it - `dvh` and `dvw` take mobile menus, etc. into account. If we used CSS rather than directly setting the style of the DOM element, we could state both and the browser could use `100dvh` if it can support it and ignore it as an invalid value if it can't.

jacbn:
LGTM. There's some flickering whenever you make a line for the first time on the modal (though not if you delete a line first?), though this is definitely an improvement!

jacbn (approved):

PR #793, opened 2023-09-13
Warn on duplicate group name

jacbn:
Will warn the user if they enter a group name when creating a new group or editing an existing one that already exists; this does not block them from adding it should they insist.

### DIFF
--- a/src/app/components/pages/Groups.tsx
+++ b/src/app/components/pages/Groups.tsx
@@ -228,8 +228,8 @@ const GroupEditor = ({group, allGroups, user, createNewGroup, groupNameInputRef}
                             <Input
                                 innerRef={groupNameInputRef} length={50} placeholder="Group name" value={newGroupName}
                                 onChange={e => setNewGroupName(e.target.value)} aria-label="Group Name" disabled={isDefined(group) && !(isUserGroupOwner || group.additionalManagerPrivileges)}
-                                invalid={newGroupName !== undefined && newGroupName.length > 0 && allGroups?.some(g => g.groupName == newGroupName && (group ? group.id != g.id : true))}
-                                valid={newGroupName !== undefined && newGroupName.length > 0 && !allGroups?.some(g => g.groupName == newGroupName && (group ? group.id != g.id : true))}
+                                invalid={newGroupName !== undefined && newGroupName.length > 0 && allGroups?.some(g => g.groupName == newGroupName && (isDefined(group) ? group.id != g.id : true))}
+                                valid={newGroupName !== undefined && newGroupName.length > 0 && !allGroups?.some(g => g.groupName == newGroupName) && (isDefined(group) ? newGroupName !== group.groupName : true)}
                             />
                             {(!isDefined(group) || isUserGroupOwner || group.additionalManagerPrivileges) && <InputGroupAddon addonType="append">
                                 <Button
--- a/src/app/components/pages/Groups.tsx
+++ b/src/app/components/pages/Groups.tsx
@@ -402,8 +402,9 @@ export const Groups = ({user}: {user: RegisteredUserDTO}) => {
     const [showArchived, setShowArchived] = useState(false);
     const groupQuery = useGetGroupsQuery(showArchived);
     const { currentData: groups, isLoading, isFetching } = groupQuery;
+    const otherGroups = useGetGroupsQuery(!showArchived);
 
-    const allGroups = [...(groups ?? []) , ...(useGetGroupsQuery(!showArchived).currentData ?? [])];
+    const allGroups = [...(groups ?? []) , ...(otherGroups.currentData ?? [])];
 
     const [createGroup] = useCreateGroupMutation();
     const [deleteGroup] = useDeleteGroupMutation();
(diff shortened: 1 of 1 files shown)


src/app/components/pages/Groups.tsx
```diff
@@ -364,8 +369,16 @@ const MobileGroupCreatorComponent = ({className, createNewGroup}: GroupCreatorPr
...
         </Col>
         <Col size={12} className="mb-2">
-            <Input length={50} placeholder="Enter the name of your group here" value={newGroupName}
-                onChange={e => setNewGroupName(e.target.value)} aria-label="Group Name"/>
+            <Form>
+                <InputGroup>
+                    <Input length={50} placeholder="Enter the name of your group here" value={newGroupName}
+                        onChange={e => setNewGroupName(e.target.value)} aria-label="Group Name"
+                        invalid={newGroupName !== undefined && newGroupName.length > 0 && allGroups?.some(g => g.groupName == newGroupName)}
+                        valid={newGroupName !== undefined && newGroupName.length > 0 && !allGroups?.some(g => g.groupName == newGroupName)}
```
mlt47, line 377: To avoid a little duplication and help with readability, I think here would be a good place to pull out this logic into a descriptively named variable (before the returned JSX).

```
const nameAlreadyInUse = isDefined(newGroupName) && newGroupName.length > 0 && allGroups?.some(g => g.groupName == newGroupName);
```
(There is also a useful utility function called `isDefined(...)`.)
The input can then have the props `invalid={nameAlreadyInUse} valid={!nameAlreadyInUse}`.


Or as it is logic shared across two components you could pull it out into a `function isNameAlreadyInUse(newGroupName, allGroups)`. As it is only this page that will need this functionality I would recommend keeping the function in this page (and module) rather than moving it into a service.

```
const nameAlreadyInUse = isNameAlreadyInUse(newGroupName, allGroups);
```

src/app/components/pages/Groups.tsx
```diff
@@ -381,6 +394,8 @@ export const Groups = ({user}: {user: RegisteredUserDTO}) => {
     const groupQuery = useGetGroupsQuery(showArchived);
     const { currentData: groups, isLoading, isFetching } = groupQuery;
 
+    const allGroups = [...(useGetGroupsQuery(false).currentData ?? []) , ...(useGetGroupsQuery(true).currentData ?? [])];
```
mlt47, line 397: It's a shame that useGetGroupsQuery() doesn't have an option to return all groups. We could then filter them appropriately in the front-end code.
Perhaps we look at altering/fixing that endpoint as your first API PR.

However, before that, we can reduce the number of queries by one by doing something like:
```
const otherGroups = useGetGroupsQuery(!showArchived);

const allFroups = [...groups, ...(otherGroups.currentData ?? [])];
```

mlt47 (changes requested):
It's looking good. I have some small suggestions.
We also talked about maybe showing a different feedback message if the group name matches an archived group name rather than matching an active group's name because matching an archived group name might be something that people are more likely to do on purpose.

src/app/components/pages/Groups.tsx
jacbn: The group valid/invalid conditions are not quite opposites, as there is a third null state if the name is valid but has not changed. Otherwise agreed, changes made!

src/app/components/pages/Groups.tsx
```diff
@@ -359,13 +368,26 @@ const MobileGroupCreatorComponent = ({className, createNewGroup}: GroupCreatorPr
...
-            <Input length={50} placeholder="Enter the name of your group here" value={newGroupName}
-                onChange={e => setNewGroupName(e.target.value)} aria-label="Group Name"/>
+            <Form>
+                <InputGroup>
+                    <Input length={50} placeholder="Enter the name of your group here" value={newGroupName}
+                        onChange={e => setNewGroupName(e.target.value)} aria-label="Group Name"
+                        invalid={isGroupNameInvalid}
+                        valid={isGroupNameValid}
+                        />
+                    <FormFeedback>A{isGroupNameInvalid && getNameConflictingGroup?.archived ? <>n archived</> : <></>} group with that name already exists.</FormFeedback>
```
mlt47, line 388: I don't think you need the `isGroupNameInvalid &&` here because <FormFeedback> is by default only shown when the input group is invalid, but please correct me if I'm wrong.

src/app/components/pages/Groups.tsx
```diff
@@ -215,6 +217,10 @@ const GroupEditor = ({group, user, createNewGroup, groupNameInputRef}: GroupCrea
     const canArchive = group && (isUserGroupOwner || group.additionalManagerPrivileges);
     const canEmailUsers = isStaff(user) && usersInGroup.length > 0;
 
+    const getNameConflictingGroup = allGroups?.find(g => g.groupName == newGroupName && (isDefined(group) ? group.id != g.id : true));
```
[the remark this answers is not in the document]
mlt47: `find`-ing the conflicting group to check the archived status later is neat though :)

mlt47 (changes requested):
Looking better and I'm quite sure the archived notification will be useful to users. Small tweak suggested.

src/app/components/pages/Groups.tsx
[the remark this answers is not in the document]
jacbn: Good spot, this was a function the first time I wrote it, hence the name.

src/app/components/pages/Groups.tsx
jacbn: Seems to work without, you're right. Good to know!

mlt47 (approved):
🙌

PR #794, opened 2023-09-13
Update button text for adding group managers

jacbn:
If there are no managers to remove, change 'Add / remove group managers' text to just 'Add group managers'.


src/app/components/pages/Groups.tsx
```diff
@@ -260,7 +260,7 @@ const GroupEditor = ({group, user, createNewGroup, groupNameInputRef}: GroupCrea
                         {/* Only teachers and above can add group managers */}
                         <Button outline={isAda} className="w-100 w-sm-auto d-inline-block text-nowrap" size="sm" color={siteSpecific("tertiary", "secondary")} onClick={() => dispatch(showGroupManagersModal({group, user}))}>
                             {isUserGroupOwner
-                                ? <>Add / remove<span className="d-none d-xl-inline">{" "}group managers</span></>
+                                ? <>Add {additionalManagers.length > 1 ? <>/ remove</> : <></>}<span className="d-none d-xl-inline">{" "}group managers</span></>
```
mlt47, line 263: It is possible to just return a string from these expressions which can sometimes be a little neater.

mlt47 (approved):

src/app/components/pages/Groups.tsx
mlt47: But I think this PR is good to go 👍

PR #796, opened 2023-09-15
Undo gameboard question deletion

jacbn:
Adds undo and redo buttons to gameboard editing for easier workflow.

Just the undo was requested, but it's then equally easy to make mistakes using that, so a redo was added too.

### DIFF
--- a/src/app/components/elements/GameboardBuilderRow.tsx
+++ b/src/app/components/elements/GameboardBuilderRow.tsx
@@ -11,8 +11,7 @@ import {
     siteSpecific,
     stageLabelMap,
     TAG_ID,
-    TAG_LEVEL,
-    GAMEBOARD_UNDO_STACK_SIZE_LIMIT
+    TAG_LEVEL
 } from "../../services";
 import React from "react";
 import {AudienceContext} from "../../../IsaacApiTypes";
--- a/src/app/components/elements/GameboardBuilderRow.tsx
+++ b/src/app/components/elements/GameboardBuilderRow.tsx
@@ -39,7 +39,9 @@ interface GameboardBuilderRowInterface {
 }
 
 const GameboardBuilderRow = (
-    {provided, snapshot, question, selectedQuestions, setSelectedQuestions, questionOrder, setQuestionOrder, previousQuestionOrderStack: previousQuestionOrderStack, setPreviousQuestionOrderStack: setPreviousQuestionOrderStack, previousSelectedQuestionsStack: previousSelectedQuestionsStack, setPreviousSelectedQuestionsStack: setPreviousSelectedQuestionsStack, resetRedoStacks, creationContext}: GameboardBuilderRowInterface
+    {provided, snapshot, question, selectedQuestions, setSelectedQuestions, questionOrder, setQuestionOrder, 
+        previousQuestionOrderStack, setPreviousQuestionOrderStack, previousSelectedQuestionsStack, 
+        setPreviousSelectedQuestionsStack, resetRedoStacks, creationContext}: GameboardBuilderRowInterface
 ) => {
     const dispatch = useAppDispatch();
 
--- a/src/app/components/elements/GameboardBuilderRow.tsx
+++ b/src/app/components/elements/GameboardBuilderRow.tsx
@@ -28,20 +28,18 @@ interface GameboardBuilderRowInterface {
     question: ContentSummary;
     selectedQuestions: Map<string, ContentSummary>;
     setSelectedQuestions: (m: Map<string, ContentSummary>) => void;
-    previousSelectedQuestionsStack: Array<Map<string, ContentSummary>>;
-    setPreviousSelectedQuestionsStack: (m: Array<Map<string, ContentSummary>>) => void;
+    setPreviousSelectedQuestionsStack: React.Dispatch<React.SetStateAction<Map<string, ContentSummary>[]>>;
     questionOrder: string[];
     setQuestionOrder: (a: string[]) => void;
     previousQuestionOrderStack: string[][];
-    setPreviousQuestionOrderStack: (a: string[][]) => void;
+    setPreviousQuestionOrderStack: React.Dispatch<React.SetStateAction<string[][]>>;
     resetRedoStacks: () => void;
     creationContext?: AudienceContext;
 }
 
 const GameboardBuilderRow = (
-    {provided, snapshot, question, selectedQuestions, setSelectedQuestions, questionOrder, setQuestionOrder, 
-        previousQuestionOrderStack, setPreviousQuestionOrderStack, previousSelectedQuestionsStack, 
-        setPreviousSelectedQuestionsStack, resetRedoStacks, creationContext}: GameboardBuilderRowInterface
+    {provided, snapshot, question, selectedQuestions, setSelectedQuestions, questionOrder, setQuestionOrder, previousQuestionOrderStack,
+        setPreviousQuestionOrderStack, setPreviousSelectedQuestionsStack, resetRedoStacks, creationContext}: GameboardBuilderRowInterface
 ) => {
     const dispatch = useAppDispatch();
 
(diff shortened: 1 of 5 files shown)


src/app/components/elements/GameboardBuilderRow.tsx
```diff
@@ -28,13 +28,18 @@ interface GameboardBuilderRowInterface {
...
     setQuestionOrder: (a: string[]) => void;
+    previousQuestionOrderStack: string[][];
+    setPreviousQuestionOrderStack: (a: string[][]) => void;
+    resetRedoStacks: () => void;
     creationContext?: AudienceContext;
 }
 
 const GameboardBuilderRow = (
-    {provided, snapshot, question, selectedQuestions, setSelectedQuestions, questionOrder, setQuestionOrder, creationContext}: GameboardBuilderRowInterface
+    {provided, snapshot, question, selectedQuestions, setSelectedQuestions, questionOrder, setQuestionOrder, previousQuestionOrderStack: previousQuestionOrderStack, setPreviousQuestionOrderStack: setPreviousQuestionOrderStack, previousSelectedQuestionsStack: previousSelectedQuestionsStack, setPreviousSelectedQuestionsStack: setPreviousSelectedQuestionsStack, resetRedoStacks, creationContext}: GameboardBuilderRowInterface
```
jsharkey13, line 42: Aside from the length of this line, do you need to be explicit in the mapping here? Most of the rest use the shorthand, and one of the five params you have added does too, but the rest do not despite looking to be mapping to the same name?

src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -52,9 +52,14 @@ interface QuestionSearchModalProps {
...
     setOriginalQuestionOrder: (a: string[]) => void;
+    previousQuestionOrder: string[][];
+    setPreviousQuestionOrder: (a: string[][]) => void;
+    previousSelectedQuestions: Array<Map<string, ContentSummary>>;
+    setPreviousSelectedQuestions: (m: Array<Map<string, ContentSummary>>) => void;
+    resetRedoStacks: () => void;
     eventLog: object[];
 }
-export const QuestionSearchModal = ({originalSelectedQuestions, setOriginalSelectedQuestions, originalQuestionOrder, setOriginalQuestionOrder, eventLog}: QuestionSearchModalProps) => {
+export const QuestionSearchModal = ({originalSelectedQuestions, setOriginalSelectedQuestions, originalQuestionOrder, setOriginalQuestionOrder, previousQuestionOrder: previousQuestionOrderStack, setPreviousQuestionOrder: setPreviousQuestionOrderStack, previousSelectedQuestions: previousSelectedQuestionsStack, setPreviousSelectedQuestions: setPreviousSelectedQuestionsStack, resetRedoStacks, eventLog}: QuestionSearchModalProps) => {
```
jsharkey13, line 62: Ditto here regarding the mix of explicit vs implicit mappings?

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -64,7 +64,12 @@
     const [gameboardTags, setGameboardTags] = useState<Item<string>[]>([]);
     const [gameboardURL, setGameboardURL] = useState<string>();
     const [questionOrder, setQuestionOrder] = useState<string[]>([]);
+    const [previousQuestionOrderStack, setPreviousQuestionOrderStack] = useState<string[][]>([]);
     const [selectedQuestions, setSelectedQuestions] = useState(new Map<string, ContentSummary>());
+    const [previousSelectedQuestionsStack, setPreviousSelectedQuestionsStack] = useState(new Array<Map<string, ContentSummary>>());
+    const [redoQuestionOrderStack, setRedoQuestionOrderStack] = useState<string[][]>([]);
+    const [redoSelectedQuestionsStack, setRedoSelectedQuestionsStack] = useState(new Array<Map<string, ContentSummary>>());
+    const resetRedoStacks = () => {setRedoQuestionOrderStack([]); setRedoSelectedQuestionsStack([]);}
```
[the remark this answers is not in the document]
jsharkey13: This function should probably come after all the hooks, and be on several lines so that the semicolon mess looks clearer!

src/app/components/elements/GameboardBuilderRow.tsx
jacbn: This is a VSC config option, good spot. `typescript.preferences.useAliasesForRenames` is now false :)

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -255,36 +269,73 @@ const GameboardBuilder = ({user}: {user: RegisteredUserDTO}) => {
...
+                                                        <div className="col-md-3 d-flex justify-content-center justify-content-md-start">
+                                                            {previousQuestionOrderStack.length > 0 && <Button
+                                                                className="mb-1 mb-md-0"
+                                                                color="primary" outline
+                                                                onClick={() => {
+                                                                    setQuestionOrder(previousQuestionOrderStack.pop() || []);
+                                                                    setSelectedQuestions(previousSelectedQuestionsStack.pop() || new Map<string, ContentSummary>());
+                                                                    setRedoQuestionOrderStack([...redoQuestionOrderStack, questionOrder]);
+                                                                    setRedoSelectedQuestionsStack([...redoSelectedQuestionsStack, selectedQuestions]);
+                                                                }}
```
jacbn, line 282: The redo stack is limited in size by the undo stack, so don't need to check size

src/app/components/elements/GameboardBuilderRow.tsx
```diff
@@ -88,6 +95,15 @@ const GameboardBuilderRow = (
...
                         setQuestionOrder(newQuestionOrder);
+                        if (provided) {
+                            if (previousQuestionOrderStack.length > 9) {
+                                setPreviousQuestionOrderStack(previousQuestionOrderStack.slice(1));
+                                setPreviousSelectedQuestionsStack(previousSelectedQuestionsStack.slice(1));
+                            }
+                            setPreviousQuestionOrderStack([...previousQuestionOrderStack, questionOrder]);
+                            setPreviousSelectedQuestionsStack([...previousSelectedQuestionsStack, selectedQuestions]);
+                            resetRedoStacks();
+                        }
```
mlt47, line 106: In this section `setPreviousQuestionOrderStack(...)` is called multiple times, but `setState(...)` does not update `previousQuestionOrderStack` immediately, it updates it at the end of the render ready for the next "snapshot" so (I think) it will take the last value each time.
It is possible to fix this by rearranging the logic or using an "updater function": https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -255,36 +269,73 @@ const GameboardBuilder = ({user}: {user: RegisteredUserDTO}) => {
...
-                                                        >
-                                                            {siteSpecific("Add Questions", "Add questions")}
-                                                            {isAda && <img className={"plus-icon"} src={"/assets/cs/icons/add-circle-outline-pink.svg"}/>}
-                                                        </Button>
-                                                    </ShowLoading>
+                                                    <div className="row w-100">
+                                                        <div className="col-md-3 d-flex justify-content-center justify-content-md-start">
+                                                            {previousQuestionOrderStack.length > 0 && <Button
+                                                                className="mb-1 mb-md-0"
+                                                                color="primary" outline
```
mlt47, line 276: Perhaps have a play around with the style. Currently undo and redo are emphasised as much as the add questions button. `/gameboards/new`'s "Shuffle questions" might be a good example to try and copy (`btn-sm`). If there are appropriate icons that might look better included on the buttons too.

mlt47 (changes requested):
This seems to function well (although perhaps the stack length of 9 is not being enforced due to multiple `setStates(...)` in a render).
I've included some suggestions.

We might also want to investigate creating a custom hook to wrap some of this logic in one place and expose a reduced API to the components. This will likely require also looking at how the `useContext()` hook works.

mlt47 (approved):
This feature has quite a lot of logic involved in getting it working.
Works well, the buttons look a lot better now and the code is improved with the refactor.

PR #798, opened 2023-09-18
Add checkbox in search to hide book questions

jacbn:
Adds a checkbox toggle to show/hide book questions (default show) when searching for questions in gameboard creation.


mlt47 (approved):
This implementation is good.

I would recommend sending a quick screenshot over to Nicki or Lisa to make sure it looks like what they had in mind.

The card also mentions the teacher gameboard builder but there is already a select option for that.

jacbn:
Confirmed wording with Nicki

PR #799, opened 2023-09-19
Add dropdown for managing group members on small devices

jacbn:
On the group management page, compress the 'Reset Password' and 'Remove [member]' buttons into a dropdown on xs resolutions.


jsharkey13 (approved):
Looks okay, I haven't tested it but you have!

PR #800, opened 2023-09-19
Immediate invalid numeric question feedback

jacbn:
Give feedback under numeric question attempts before submitting if the formatting is incorrect (similar functionality exists already for symbolic questions).


src/app/components/content/IsaacNumericQuestion.tsx
```diff
@@ -123,6 +123,46 @@ const IsaacNumericQuestion = ({doc, questionId, validationResponse, readonly}: I
         dispatchSetCurrentAttempt(attempt);
     }
```
jacbn, line 125: The bulk of this code (minus the regexes/errors) was reused from IsaacSymbolicQuestion

mlt47 (approved):
This looks good and works. I think a refactor to reduce the duplication is a good idea though.

PR #801, opened 2023-09-22
Instructional text for graph sketcher

jacbn:
Adds instructional text beneath the graph sketcher, as requested by the content team, for ease of use. The text does not display when printing.


mlt47 (approved):
Nice

PR #802, opened 2023-09-25
Move common debouncing code into shared validator

jacbn:
Refactors Equality and the Numeric, Symbolic, SymbolicLogic IsaacQuestions to reuse shared debouncing code.

### DIFF
--- a/src/app/components/content/IsaacQuestionValidator.tsx
+++ b/src/app/components/content/IsaacQuestionValidator.tsx
@@ -1,0 +1,31 @@
+import { useRef, useState } from "react";
+import { isDefined } from "../../services";
+import React from "react";
+
+interface QuestionInputValidationProps<T> {
+    userInput: T;
+    validator: (i: T) => string[];
+}
+
+const QuestionInputValidation = <T extends string | number>({userInput, validator}: QuestionInputValidationProps<T>) => {
+    const [errors, setErrors] = useState<string[]>([]);
+    const debounceTimer = useRef<number|null>(null);
+
+    if (debounceTimer.current) {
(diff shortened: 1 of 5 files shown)


src/app/components/content/IsaacQuestionValidator.tsx
```diff
@@ -0,0 +1,31 @@
+import { useRef, useState } from "react";
+import { isDefined } from "../../services";
+import React from "react";
+
+interface QuestionInputValidationProps<T> {
+    userInput: T;
+    validator: (i: T) => string[];
+}
+
+const QuestionInputValidation = <T extends string | number>({userInput, validator}: QuestionInputValidationProps<T>) => {
```
mlt47, line 10:
```suggestion
const QuestionInputValidation = <T>({userInput, validator}: QuestionInputValidationProps<T>) => {
```
I would have thought that making T extend string or number is not necessary on this occasion as we don't use the user input's stringy or numbery-ness in QuestionInputValidation we just pass it to the validation function that needs things of shape `T`.

mlt47 (approved):
Always good to remove duplication. Small suggested change to generic type of the new component. Happy for it to be merged otherwise.

src/app/components/content/IsaacQuestionValidator.tsx
jacbn: I was trying to do this but it's invalid syntax in tsx. ESLint suggested `<T extends string | number>` based on current usage originally, though `<T,>` seems to be what we're after! ([relevant SO](https://stackoverflow.com/questions/32308370/what-is-the-syntax-for-typescript-arrow-functions-with-generics))

PR #805, opened 2023-09-26
Scroll connections lists when too large

jacbn:
On My Accounts page, teacher/student connections list length is unbounded; this seems to cause performance issues on lower-end devices. This change creates a window to contain and hide elements outside a given size range.


mlt47 (approved):
This is a useful library.
A tiny tweak would be to set `scrollbar-gutter: stable` so that Revoke is aligned even if one of the lists is not overflowing.

As we discussed at the office we do now lose the ability to Ctrl+F search for a particular student or teacher on this page so I would suggest adding a simple filter. That can be a separate PR.

PR #807, opened 2023-09-27
Fix contrast issue with `text-muted` class

mwtrew:
This brings the contrast of `text-muted` up to WCAG requirements for Ada CS. 

It then became apparent that the placeholder text for certain question types used `text-muted`, so for consistency I changed that and made sure that all placeholder text is now `$input-placeholder-color` as defined by each site.

### DIFF
--- a/src/scss/common/forms.scss
+++ b/src/scss/common/forms.scss
@@ -65,7 +65,6 @@
 
 input {
   &.form-control {
-    @include placeholder($gray-500);
     border: solid 1px $black;
     padding: 0 1.2em;
   }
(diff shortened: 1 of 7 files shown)


src/scss/common/forms.scss
```diff
@@ -65,7 +65,6 @@
 
 input {
   &.form-control {
-    @include placeholder($gray-500);
```
mlt47, line 68: I think we need to keep this (for physics at least).

mlt47 (changes requested):
The `color` needs to be set on the `::placeholder` pseudo element for it to be applied to placeholders. If this is not set it seems to take the browser's default placeholder colour rather than the input element's `color` and the browser default was not WCAG AAA for me.

PR #809, opened 2023-09-29
Show subtitles in search, question pages, and gameboard viewer

jacbn:
Displays a question's subtitle in various places. In search, hides when a summary is present.

(requires API change: https://github.com/isaacphysics/isaac-api/pull/561)


mlt47 (approved):
I'm happy with this once we confirm with Nicki when to display the subtitle and when to display the summary.

PR #810, opened 2023-09-29
Prevent question select clipping on tiny devices

jacbn:
Stops hexagons from clipping off the edge of the screen for very small devices (min supported width now 343px).


mlt47 (approved):
Neat! 🙂

PR #812, opened 2023-10-02
Redesign Isaac Gameboard card view

jacbn:
Reordered a lot of the information on the card to follow an expected order.

### DIFF
--- a/src/app/components/elements/cards/BoardCard.tsx
+++ b/src/app/components/elements/cards/BoardCard.tsx
@@ -36,6 +36,7 @@ import React from "react";
 import {Link} from "react-router-dom";
 import {BoardAssignee, Boards} from "../../../../IsaacAppTypes";
 import indexOf from "lodash/indexOf";
+import { Spacer } from "../Spacer";
 
 
 interface HexagonGroupsButtonProps {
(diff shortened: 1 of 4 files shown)


src/app/components/elements/cards/BoardCard.tsx
```diff
@@ -110,7 +111,9 @@ export const BoardCard = ({user, board, boardView, assignees, toggleAssignModal,
 
     const basicCellClasses = `align-middle ${siteSpecific("text-center", "text-left")}`;
 
-    const phyHexagon = <div className={classNames("board-subject-hexagon-container", {"table-view": boardView === BoardViews.table})}>
+    const isTable = boardView === BoardViews.table;
+    const phyHexagon = <div className={classNames("board-subject-hexagon-container", {"table-view": isTable})} 
+            style={{right: `${isTable ? "" : "45px"}`}}>
```
mwtrew, line 116: This should probably go in the `table-view` class instead of using inline style (which we generally avoid).

mwtrew (changes requested):
This looks a lot better, nice work! One minor comment about the use of inline style.

mwtrew (approved):

PR #816, opened 2023-10-04
Add scroll-to-top button

skyepurchase:
ScrollToTop component added to the IsaacApp component. It remains at the bottom of the screen (out of site) until the user scrolls beyond the top of the main content at which point it sticks to the bottom right of the screen.

Currently only on the Ada site but removing the conditional will add it to the Physics site as well. The scroll-button.scss will also need to be imported in the relevant isaac.scss file.

### DIFF
--- a/src/app/components/navigation/IsaacApp.tsx
+++ b/src/app/components/navigation/IsaacApp.tsx
@@ -39,7 +39,7 @@ import {
     KEY,
     showNotification,
     isTutorOrAbove,
-    PATHS, isAda
+    PATHS
 } from "../../services"
 import {Generic} from "../pages/Generic";
 import {ServerError} from "../pages/ServerError";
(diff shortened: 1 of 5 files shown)


skyepurchase:
Button added to the physics site as well. In the process fixed a minor bug where the button was visible at the bottom of the screen even though the button was not in use.

src/app/components/navigation/IsaacApp.tsx
```diff
@@ -39,7 +39,7 @@ import {
     KEY,
     showNotification,
     isTutorOrAbove,
-    PATHS
+    PATHS, isAda
```
mlt47, line 42:
```suggestion
    PATHS
```

mlt47 (approved):
This looks good. Although it is not typical to need to use useRef as much, in this case I think it is the correct thing to do.

PR #817, opened 2023-10-05
Reword additional managers and ownership change dialogue

skyepurchase:
Clearer wording about the inability for demoted owners to add or remove managers (which they previously could do).

The warning dialogue about converting a manager to an owner distinguishes between whether the privileges are allowed or not.


mlt47 (approved):
I've tested this and it works.

PR #824, opened 2023-10-12
My Assignments card text overflow

jacbn:
The card title text was overflowing for long strings with no soft breaks.


mlt47 (approved):
Looks good to me!

PR #827, opened 2023-10-19
New Ada content headers

skyepurchase:
With the updated headers in Ada content accordions H3 titles will be underlined (the largest title in a content accordion) and H4 titles (nested underneath H3) are not.

Note that the accordion title and internal headers all have the same style now (that is the old accordion title <span>).


jacbn:
LGTM

PR #828, opened 2023-10-20
Add filters to the "Manage Tests" page

jacbn:
Adds title, group, and date filtering to the Manage Tests page.

Ideally once the ability to set tests to multiple groups at once is in place, cards in this page could be grouped in a similar way.

### DIFF
--- a/src/app/components/pages/quizzes/SetQuizzes.tsx
+++ b/src/app/components/pages/quizzes/SetQuizzes.tsx
@@ -12,7 +12,7 @@ import {ShowLoading} from "../../handlers/ShowLoading";
 import {QuizAssignmentDTO, QuizSummaryDTO, RegisteredUserDTO} from "../../../../IsaacApiTypes";
 import {TitleAndBreadcrumb} from "../../elements/TitleAndBreadcrumb";
 import {Spacer} from "../../elements/Spacer";
-import {formatDate} from "../../elements/DateString";
+import {formatDate, formatISODateOnly} from "../../elements/DateString";
 import {AppQuizAssignment} from "../../../../IsaacAppTypes";
 import {
     below, confirmThen,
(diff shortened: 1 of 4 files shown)


src/scss/phy/quiz.scss
```diff
@@ -0,0 +1,11 @@
+@import "../common/quiz.scss";
+
+.quiz-date-filter-type {
+  button {
+    min-width: unset;
+  }
+}
+
+.quiz-filter-date-span {
+  transform: translateY(3px);
```
jacbn, line 10: I'm not convinced this is the best way to do this, if there are any other suggestions. Text inside buttons in Isaac have different heights to those in Ada -- this does alleviate the problem, but forcing it down feels a little ugly.

src/app/components/pages/quizzes/SetQuizzes.tsx
```diff
@@ -177,16 +186,122 @@ const SetQuizzesPageComponent = ({user, location}: SetQuizzesPageProps) => {
                 </ShowLoading>,
 
                 [siteSpecific("Manage Tests", "Previously set tests")]:
+                <>
+                    <div className="d-none d-md-flex justify-content-center">
+                        <RS.Button color="tertiary" size="sm" onClick={() => setShowFilters(s => !s)}>
+                            {showFilters ? "Hide filters" : "Show filters"}
+                        </RS.Button>
+                    </div>
+                    {showFilters && <RS.Row>
```
mlt47, line 195: Once opened, the filters are not responsive to screen size changes. I think we would want these to be accessible to mobile too. Anecdotally, teachers do sometimes do classroom management tasks on mobile devices - on their commute, for example.

src/app/components/pages/quizzes/SetQuizzes.tsx
```diff
@@ -177,16 +186,122 @@ const SetQuizzesPageComponent = ({user, location}: SetQuizzesPageProps) => {
...
+                                        switch (quizDueDateFilterType) {
+                                            case 'after':
+                                                return assignment.creationDate >= nthHourOf(24, quizDueDate);
+                                            case 'before':
+                                                return assignment.creationDate <= nthHourOf(0, quizDueDate);
+                                            case 'on':
+                                                return assignment.creationDate >= nthHourOf(0, quizDueDate) && assignment.creationDate <= nthHourOf(24, quizDueDate);
+                                        }
+                                    });
+                                }
```
mlt47, line 302: Some shared filter code here that could be lifted out into a function.

mlt47 (changes requested):
This works well.
I would suggest changing it to work responsively and then sending a screenshot to the content teams to check that they approve (as they did not specifically ask for this and they might want to add/alter documentation).

mlt47 (approved):

PR #829, opened 2023-10-24
Small, high-value updates to tests and assignments

jacbn:
Many small changes for consistency between tests and assignments:
- Removed date of creation column in Set Assignments table;
- Reverses the order of the Set / Manage tests page to show the most recent at the top;
- In the same page, maintains the current tab when moving through the browser history stack;
- Link to the test preview in Assignment Progress;
- Summarise class-wide performance over each question in the tests table on the same page.


src/app/components/pages/quizzes/SetQuizzes.tsx
```diff
@@ -87,25 +88,15 @@ function QuizAssignment({user, assignment}: QuizAssignmentProps) {
     </div>;
 }
 
-const SetQuizzesPageComponent = ({user, location}: SetQuizzesPageProps) => {
+const SetQuizzesPageComponent = ({user}: SetQuizzesPageProps) => {
     const dispatch = useAppDispatch();
     const deviceSize = useDeviceSize();
-    const hashAnchor = location.hash?.slice(1) ?? null;
-    const [activeTab, setActiveTab] = useState(MANAGE_QUIZ_TAB.set);
+    const [activeTab, setActiveTab] = useHistoryState("currentTab", MANAGE_QUIZ_TAB.set);
```
mwtrew, line 94: Removing the hash anchor means we can no longer link directly to a tab, e.g. https://staging.adacomputerscience.org/set_tests#manage. Have you checked that we don't do that anywhere on the site or in the content?

mwtrew (approved):
Looks good overall, a couple of points/questions before we merge.

PR #830, opened 2023-10-25
Test setting to multiple groups

skyepurchase:
Relies on [Test setting to multiple groups #564](https://github.com/isaacphysics/isaac-api/pull/564) in the backend as it anticipates the new POST and return types.

Single quiz assignments are still possible by selecting only one group.

### DIFF
--- a/src/app/components/elements/modals/QuizSettingModal.tsx
+++ b/src/app/components/elements/modals/QuizSettingModal.tsx
@@ -6,9 +6,9 @@ import {
     useGetGroupsQuery,
     useAssignQuizMutation,
     showSuccessToast,
-    mutationSucceeded, closeActiveModal
+    mutationSucceeded, closeActiveModal, useAppSelector, selectors
 } from "../../../state";
-import {isDefined, Item, nthHourOf, selectOnChange, TODAY} from "../../../services";
+import {assignMultipleQuiz, isDefined, Item, nthHourOf, selectOnChange, TODAY} from "../../../services";
 import range from "lodash/range";
 import {currentYear, DateInput} from "../inputs/DateInput";
 import {IsaacSpinner} from "../../handlers/IsaacSpinner";
@@ -48,6 +48,7 @@ interface QuizSettingModalProps {
 export function QuizSettingModal({allowedToSchedule, quiz, dueDate: initialDueDate, scheduledStartDate: initialScheduledStartDate, feedbackMode: initialFeedbackMode}: QuizSettingModalProps) {
     const dispatch: AppDispatch = useAppDispatch();
     const groupsQuery = useGetGroupsQuery(false);
+    const user = useAppSelector(selectors.user.loggedInOrNull);
 
     const [assignQuiz, {isLoading: isAssigning}] = useAssignQuizMutation();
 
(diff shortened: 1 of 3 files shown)


src/app/components/elements/modals/QuizSettingModal.tsx
```diff
@@ -45,17 +46,18 @@ interface QuizSettingModalProps {
...
 
-    const [assignQuiz, {isLoading: isAssigning}] = useAssignQuizMutation();
+    const [_assignQuiz, {isLoading: isAssigning}] = useAssignQuizMutation();
 
     const [validated, setValidated] = useState<Set<ControlName>>(new Set());
     const [selectedGroups, setSelectedGroups] = useState<Item<number>[]>([]);
     const [dueDate, setDueDate] = useState<Date | null>(initialDueDate ?? null);
     const [scheduledStartDate, setScheduledStartDate] = useState<Date | null>(initialScheduledStartDate ?? null);
-    const [feedbackMode, setFeedbackMode] = useState<QuizFeedbackMode | null>(initialFeedbackMode ?? null);
+    const [feedbackMode, setFeedbackMode] = useState<QuizFeedbackMode>();
```
mlt47, line 60: Is there a reason we're dropping the ability to set the initial feedback mode? I see that we're not making use of it so far but neither are we making use of the initialScheduledStartDate or initialDueDate.

src/app/services/quiz.ts
```diff
@@ -1,26 +1,192 @@
...
+            appDispatch(showErrorToast(
+                `${siteSpecific("Quiz", "Test")} assignment failed`,
+                "Error: Please choose one or more groups."
+            ));
+            return rejectWithValue(null);
+        }
+
+        const today =  TODAY();
+        const isDue = dueDate !== undefined;
+        const isScheduled = scheduledStartDate !== undefined;
```
mlt47, line 76: I see we perform similar validation when setting bulk assignments and that is, correctly, the most similar part of code to copy. However, I'm not entirely convinced we should be doing validation of this sort in the action dispatching thunk. We have this validation logic in the back-end, where it matters, and we also have this logic encoded in the form in the front-end, for a more responsive user experience. I think that having the validation logic here again doesn't gain us anything except increasing the chance of introducing bugs when duplicated logic is not updated consistently in all places.
Let's not do it as part of this PR, but I would suggest creating a separate card to remove this validation logic for both quizzes and assignments. It might even mean we can replace the createAsyncThunk with more standard RTK Query data fetching (but that might not be possible). What do you think?

src/app/state/slices/api/quizApi.ts
```diff
@@ -165,9 +166,9 @@ export const quizApi = isaacApi.enhanceEndpoints({
             })
         }),
 
-        assignQuiz: build.mutation<QuizAssignmentDTO, QuizAssignmentDTO>({
+        assignQuiz: build.mutation<AssignmentFeedbackDTO[], QuizAssignmentDTO[]>({
```
mlt47, line 169: It is confusing (and obviously not your fault!) that the name AssignmentFeedbackDTO does not match the back-end's name for the same data transfer object, AssignmentStatusDTO. I would recommend altering one or the other so that they match again, but this can be a separate card on Trello.

mlt47 (commented):
A few small comments and maybe a few separate tasks for the future but I'm happy with this.

src/app/components/elements/modals/QuizSettingModal.tsx
skyepurchase: At one stage I had swapped these to being `undefined` rather than `null` which didn't work and I must have forgotten to reset feedback mode correctly. Will change this back

src/app/services/quiz.ts
skyepurchase: I think a separate card would be good. Would be something to discuss, but I agree that there are a lot of places where the same check/group of checks are done where they are not necessarily required.

src/app/state/slices/api/quizApi.ts
skyepurchase: Card created

src/app/components/elements/modals/QuizSettingModal.tsx
```diff
@@ -79,7 +100,7 @@ export function QuizSettingModal({allowedToSchedule, quiz, dueDate: initialDueDa
                 defaultErrorTitle={"Error fetching groups"}
                 thenRender={groups => {
                     const groupOptions: Item<number>[] = groups.map(g => ({label: g.groupName as string, value: g.id as number}));
-                    return <StyledSelect
+                    return <StyledSelect isMulti placeholder="Select groups"
```
mlt47, line 103: Sorry, what I meant by the label needing updating was the value in line 97 rather than the placeholder. "None" for the placeholder is fine (if a bit "techy") as it matches the group(s) input for setting assignments.

mlt47 (changes requested):
Tiny label tweak recommendation.

mlt47 (approved):

PR #833, opened 2023-10-27
Add screenreader only tables for "My progress" charts

skyepurchase:
The charts are now hidden from screen readers with a visually hidden table for the screenreaders. This table can later be made visible if desired.

Additional changes to comply with ESLint.

### DIFF
--- a/src/app/components/elements/views/QuestionProgressCharts.tsx
+++ b/src/app/components/elements/views/QuestionProgressCharts.tsx
@@ -155,12 +155,13 @@ export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
                         No data
                     </span> :
                     <table className={"sr-only"}>
-                    <tbody>
+                    <thead>
                     <tr>
                         <th>Subject</th>
                         <th>{subId}</th>
                     </tr>
-                    </tbody>
+                    </thead>
+					<tbody>
                     {categoryColumns.map((val, key) => {
                         return (
                             <tr key={key}>
@@ -169,6 +170,7 @@ export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
                             </tr>
                         );
                     })}
+					</tbody>
                 </table>}
             </div>
         </RS.Col>}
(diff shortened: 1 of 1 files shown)


src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -70,6 +70,8 @@ export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
         .sort(comparatorFromOrderedValues(difficultiesOrdered as string[]))
         .map((key) => [difficultyLabelMap[key as Difficulty], questionsByStageAndDifficulty[stageChoices[0].value][key]]) : [];
 
+    console.log(categoryColumns, topicColumns, difficultyColumns);
```
mlt47, line 73: We should remove unnecessary console.logs :)
```suggestion
```

src/app/components/elements/views/QuestionProgressCharts.tsx
```diff
@@ -159,15 +181,35 @@ export const QuestionProgressCharts = (props: QuestionProgressChartsProps) => {
...
-                <div id={`${subId}-topicChart`} className="text-center-width doughnut-binding  align-self-center">
+                <div aria-hidden={"true"} id={`${subId}-topicChart`} className="text-center-width doughnut-binding  align-self-center">
                     <strong>{isAllZero(topicColumns) ? "No data" : ""}</strong>
                 </div>
+                {isAllZero(topicColumns) ?
+                    <span className={"sr-only"}>
+                        No data
+                    </span> :
+                    <table className={"sr-only"}>
+                        <tbody>
```
mlt47, line 198: Headings should be in `<thead>` and it should be the data rows that are in `<tbody>`.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table

mlt47 (changes requested):
Small changes, but looks useful!

mlt47 (approved):

PR #839, opened 2023-11-09
Feature/my assignments layout change

mlt47:
Update My Assignments layout in accordance to designs agreed upon by the content team.
Updates the definition of a complete assignment by having attempted every question part rather than having answered each question part correctly.

I chose to do front-end pagination of the My Assignments results to handle the case where power users have hundreds of assignments and the default now is to not filter assignments but to order assignments in a "reasonable" way.
Front-end pagination was preferred to using `react-window` because the elements inside the window can be expanded making the size dynamically variable which would have complicated the implementation.

We **should not merge** until the content teams have had time to communicate the difference in the meaning of "completed assignment".


jacbn (approved):
Looks good. There's no "Show more" option for Ada's My Assignments cards, worth checking if they want it too.

Also just to relay the above, 'We **should not merge** until the content teams have had time to communicate the difference in the meaning of "completed assignment".'

src/app/components/pages/AssignmentProgress.tsx
```diff
@@ -148,7 +148,7 @@ export const ProgressDetails = ({assignment}: {assignment: EnhancedAssignmentWit
     }
 
     const tableHeader = <div className="progress-header">
-        <strong>{studentsCorrect}</strong> of <strong>{progress.length}</strong>
+        <strong>{noStudentsAttemptedAll}</strong> of <strong>{progress.length}</strong>
         {` students have completed the assignment `}
```
mlt47, line 152:
```suggestion
        {`  students attempted all questions in `}
```

src/app/services/gameboards.tsx
mwtrew, line 1: We should also change `BoardLimit.All` to be lowercase since it's now shown on the UI, and we don't use all caps for "All" elsewhere.

src/app/services/gameboards.tsx
```diff
@@ -208,15 +208,18 @@ export const BOARD_ORDER_NAMES: {[key in BoardOrder]: string} = {
     "-visited": "Date Visited Descending",
     "title": "Title Ascending",
     "-title": "Title Descending",
-    "completion": "Completion Ascending",
-    "-completion": "Completion Descending"
+    "attempted": "Completion Ascending",
```
mwtrew, line 211: I think we should use "Attempted" in this label, not "Completion", like we're now doing under "Status" in My Assignments.

mwtrew (approved):
Big improvement, a couple of minor comments relating to presentation. Nice to see some new tests.

src/app/services/gameboards.tsx
mlt47: I have not made this change yet. Will write it down to do as a small change for the next release. I was less confident that this would work as there was some logic that compared strings.

PR #843, opened 2023-11-15
Ada style glossary for isaacphysics.org/glossary

skyepurchase:


src/app/components/pages/Glossary.tsx
```diff
@@ -15,7 +15,8 @@ import {
     scrollVerticallyIntoView,
     TAG_ID,
     tags,
-    useUrlHashValue
+    useUrlHashValue,
```
jacbn, line 18: Unfortunately the URL hashes won't work anymore as the relevant subject isn't displayed when you load the page. This is a bit of a problem as e.g. concept pages won't be able to link back to their respective glossaries. Maybe we could save which subject is selected in the URL and load that too on start?
(as in, /glossary?stage=all&subject=maths#a-functions-absvalue)

src/app/components/pages/Glossary.tsx
```diff
@@ -223,7 +241,18 @@ export const Glossary = () => {
                                 onChange={e => setFilterTopic(topics.find(v => v.id === (e as Item<TAG_ID> | undefined)?.value)) }
                                 isClearable
                             />
-                        </Col>
+                        </Col>}
+                        {isPhy && <Col className="mt-3 mt-md-0">
+                            <Label for='subject-select' className='sr-only'>Subject</Label>
+                            <StyledSelect inputId="subject-select"
```
jacbn, line 247: From a UX perspective having this box (which you _have_ to use before you can use the search) be after the search box feels wrong. I'd either switch them around (and float the search bar to the right maybe?) or disable the search box if the subject is empty. Alternatively, we could make the search box work for all subjects if none is selected, but not sure how this would cope with the glossary sizes we have currently.

jacbn (approved):
Looks good, a bit slow to run on local dev but so is the Ada glossary in master so should be okay, just keep an eye on staging when it goes up. A couple of comments about usability for both content editors and users.

src/app/components/pages/Glossary.tsx
skyepurchase: Will move this to the cosmetic changes card

src/app/components/pages/Glossary.tsx
```diff
@@ -60,22 +65,78 @@
...
+function tagByValue(values: string[], tags: Tag[]) {
+    return tags.filter(option => values.includes(option.id)).at(0);
+}
+
+interface QueryStringResponse {
+    queryStages: Tag | undefined;
+    querySubjects: Tag | undefined;
+}
+function processQueryString(query: string): QueryStringResponse {
+    const {subjects, stages} = queryString.parse(query);
```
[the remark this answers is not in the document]
skyepurchase: This was added for stage selection in an upcoming card

src/app/components/pages/Glossary.tsx
skyepurchase: New commit should fix this issue

jacbn (approved):
Looks good, just committing on top for some tiny changes to (a) keep the hash in place while it's loading and and (b) stop tab crashes. Loop me in when you finish the card that builds on this!

PR #846, opened 2023-11-23
Move subtitles above ContentSummary breadcrumbs

jacbn:
Moves subtitles/descriptions (with priority on descriptions) above the breadcrumbs for large screens. For small (sm) screens, removes the breadcrumbs. For extra small (xs) screens, only show subtitles.

### DIFF
--- a/src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
+++ b/src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
@@ -1,6 +1,7 @@
 import {ContentSummaryDTO} from "../../../../IsaacApiTypes";
 import {
     AUDIENCE_DISPLAY_FIELDS,
+    below,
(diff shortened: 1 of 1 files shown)


src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
```diff
@@ -1,6 +1,7 @@
 import {ContentSummaryDTO} from "../../../../IsaacApiTypes";
 import {
     AUDIENCE_DISPLAY_FIELDS,
+    below,
```
skyepurchase, line 4: This is imported but never used

skyepurchase, line 4:
```suggestion
```

skyepurchase (changes requested):

PR #847, opened 2023-11-23
Support custom axis labels in sketcher

jacbn:
Passes through custom axis labels received from the API to the sketcher for display. Also reorders the sketcher UI to prevent button / text field overlaps, particularly in debug mode.


mlt47 (approved):
These new fields will be undefined in many cases (and until the ETL starts processing the new fields) but the library correctly defaults to "x" and "y" in that case.

PR #851, opened 2023-11-28
Add filtering by stage to the new glossary

skyepurchase:
Merge after https://github.com/isaacphysics/isaac-react-app/pull/843 has been merged


src/app/components/pages/Glossary.tsx
```diff
@@ -206,24 +277,45 @@ export const Glossary = () => {
...
                                 placeholder="All topics"
                                 onChange={e => setFilterTopic(topics.find(v => v.id === (e as Item<TAG_ID> | undefined)?.value)) }
                                 isClearable
                             />
-                        </Col>
+                        </Col>}
+                        {isPhy && <Col className="mt-3 mt-md-0">
+                            <Label for='stage-select' className='sr-only'>Stage</Label>
+                            <StyledSelect inputId="stage-select"
+                                options={ stages.map(e => ({ value: e, label: stageLabelMap[e]})) }
```
jacbn, line 311: I would just try to keep the naming of the anonymous parameters consistent. The convention is the singular of the array name / first letter of, `e` in particular is mostly reserved for events so looks a bit weird here.
```suggestion
                                options={ stages.map(s => ({ value: s, label: stageLabelMap[s]})) }
```

src/app/components/pages/Glossary.tsx
```diff
@@ -206,24 +277,45 @@ export const Glossary = () => {
...
+                                value={queryStages ? ({value: queryStages, label: stageLabelMap[queryStages]}) : undefined}
+                                name="stage-select"
+                                placeholder="Select a stage"
+                                onChange={e => setFilterStage(stages.find(v => v === e?.value)) }
```
jacbn, line 315:
```suggestion
                                onChange={e => setFilterStage(stages.find(s => s === e?.value)) }
```

jacbn (approved):
Looks good, I'm hesitant to merge it without testing it though. Can we add a glossary term with a stage and try it with that?

skyepurchase:
Will test! I believe there is one glossary term that has a stage and subject and so will be visible. It is unpublished but should appear on local dev.

skyepurchase:
Have another minor change to variables that I will add to this PR after testing

PR #852, opened 2023-11-30
Improvement glossary UI in the front end

skyepurchase:


src/app/components/pages/Glossary.tsx
```diff
@@ -278,13 +279,15 @@ export const Glossary = () => {
                     <Row className="no-print">
                         {isPhy && <Col className="mt-3 mt-md-0">
                             <Label for='subject-select' className='sr-only'>Subject</Label>
-                            <StyledSelect inputId="subject-select"
-                                options={ subjects.map(e => ({ value: e.id, label: e.title})) }
-                                value={filterSubject ? ({value: filterSubject.id, label: filterSubject.title }) : undefined}
+                            <Select inputId="subject-select"
```
jacbn, line 282:
```suggestion
                            <Select inputId="subject-select" className="glossary-select"
```

src/scss/phy/glossary.scss
```diff
@@ -0,0 +1,45 @@
+@import "../common/glossary";
+
+.biology {
```
jacbn, line 3:
```suggestion
.glossary-select {
    &.biology {
```

src/scss/phy/glossary.scss
```diff
@@ -0,0 +1,45 @@
...
+    .select__control {
+        background-color: $phy_biology !important;
+    }
+
+    .select__single-value {
+        color: #fff;
+    }
+}
+
+.chemistry {
```
jacbn, line 13: these three will need the same treatment

jacbn (approved):
Nice, works great. I'm a bit worried though about defining `.physics`/`.maths`/etc as global properties, as they get used in a few other places in the code. It's _probably_ okay here as you have another `.select` after anyway, but in the off chance someone changes this / adds a select elsewhere it's worth wrapping them in some tag that will definitely only get used by the glossary.

src/app/components/pages/Glossary.tsx
skyepurchase: There already exists a className prop so will put it in that instead

PR #853, opened 2023-11-30
Link to account if preferences do not match page

skyepurchase:
The warning when a users preferences do not match the page's audience now includes a link to change account preferences. Only on Ada.


mwtrew (approved):
Looks good, worth addressing the unused import though.

mwtrew (approved):

PR #856, opened 2023-12-04
Interstitial cookie screen

jacbn:
On loading a page with YouTube or Anvil content, this content should instead be replaced with a screen informing the user that the third-party site may be using cookies and that, by continuing, they agree to use said cookies. Once the user clicks agree, the iframe content can be loaded as normal.

Isaac will store a cookie for both YouTube's and Anvil's acceptance, such that the user does not need to click the accept button once done the first time. We also store this in redux state to allow multiple pieces of content in the same page to update once any of the buttons to accept cookies has been clicked.


jsharkey13 (commented):
This seems okay to me, though we should review the acceptance text. Can't really comment on the structure, but the code looks good!

src/app/components/handlers/InterstitialCookieHandler.tsx
```diff
@@ -0,0 +1,107 @@
...
+const youtubeCookieText = <p>We use YouTube to show you videos on our website. We ask for your permission before loading the content, as YouTube may be using cookies to help them track usage and improve their services.<br/><br/>You may wish to read the <a href="https://policies.google.com/technologies/cookies" target="_blank" rel="noopener noreferrer"><b>Google Cookie Policy</b></a> and <a href="https://policies.google.com/privacy" target="_blank" rel="noopener noreferrer"><b>Privacy Policy</b></a> before accepting.</p>;
+const anvilCookieText = <p>We use Anvil to show you interactive content on our website. We ask for your permission before loading the content, as Anvil may be using cookies to help them track usage and improve their services.<br/><br/>You may wish to read the <a href="https://anvil.works/privacy" target="_blank" rel="noopener noreferrer"><b>Anvil Privacy Policy</b></a> before accepting.</p>;
```
jacbn, line 9: These texts will need reviewing

mlt47 (approved):
This works well for me.

PR #858, opened 2023-12-06
Fix hash anchor on general pages

jacbn:
On general pages, the hash anchor was attempting to scroll the page before any content had loaded. This now rescrolls the content once it has loaded.


src/app/components/pages/Generic.tsx
```diff
@@ -15,6 +15,7 @@ import {PrintButton} from "../elements/PrintButton";
 import {WithFigureNumbering} from "../elements/WithFigureNumbering";
 import {MetaDescription} from "../elements/MetaDescription";
 import classNames from "classnames";
+import { useUntilFound } from "./Glossary";
```
mlt47, line 18: If there is common code, like this utility custom hook, we should move it into a service. It's partly to keep the code neater but the main benefit is that it reduces the chance of accidentally introducing cyclic imports causing an imported value to be undefined when used. That type of bug is very tricky to find when it is accidentally introduced!

PR #860, opened 2023-12-07
Changing schools modal

skyepurchase:
- Fix ESLint warnings
- Fix ESLint warnings
- Fix ESLint warnings
- Add a modal to help teachers when changing schools
- Fix ESLint warnings for yet another file

### DIFF
--- a/src/app/components/elements/inputs/SchoolInput.tsx
+++ b/src/app/components/elements/inputs/SchoolInput.tsx
@@ -62,7 +62,6 @@ export const SchoolInput = ({userToUpdate, setUserToUpdate, submissionAttempted,
     }, [userToUpdate]);
 
     // Set schoolId or schoolOther
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     function setUserSchool(school: any) {
         const {schoolId: _schoolId, schoolOther: _schoolOther, ...userWithoutSchoolInfo} = userToUpdate;
         if (school.urn) {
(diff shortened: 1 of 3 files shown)


src/app/components/elements/inputs/SchoolInput.tsx
```diff
@@ -58,11 +58,13 @@ export const SchoolInput = ({userToUpdate, setUserToUpdate, submissionAttempted,
 
     useEffect(() => {
         fetchSchool(userToUpdate.schoolId || "");
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [userToUpdate]);
 
     // Set schoolId or schoolOther
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jacbn, line 65: The ESLint problems that we _should_ fix but would take more effort (like this one, which is a problem because we really shouldn't be using `if (school.urn)` to infer type but `typeof` / `instanceof`) we probably shouldn't comment out, but the unfixable ones that e.g. cause a useEffect loop are fine. Otherwise they won't ever get fixed...

jacbn (approved):
Works great, nice one :)

src/app/components/elements/inputs/SchoolInput.tsx
skyepurchase: That is true. Need to do a bit sweep of all the files that have `any` and fix them or have some reason for it being unreasonable

PR #862, opened 2023-12-08
Print at scale without overflowing

jacbn:
- Unsets min-width when printing so the Scale parameter in print menus works as intended;
- Since the previous print layout was designed for 992px min-width, sets the page size to this plus the padding difference not added by @ page size (0.5in left/right/top/bottom).

This correction isn't perfect, and the scale of certain pages is very slightly different to before. Worth checking if this affects any pages in particular.


jsharkey13 (approved):
I don't like this print size calc business, but A4 does look weird and changing it to A3 to get back to what it used to be isn't great either.
It works, from what I can see in Chrome, Firefox and Safari, so let's go with it.

PR #864, opened 2023-12-11
Fix date input element size and borders

jacbn:
The DateInput element has removed some ReactStrap-default properties but keeping others, making the overall element look a little off, particularly when invalid.


skyepurchase (approved):
LGTM

PR #865, opened 2023-12-11
Underline footer links on hover

jacbn:
As to meet WCAG guideline G183.


jsharkey13 (commented):
Previously it did not include the bolded external links on the left hand side, due to a messy SCSS hierarchy.

PR #867, opened 2023-12-13
My Tests: same height grid elements

jacbn:
Adds a CardGrid component for correctly managing the heights of elements in a grid, as to replace `block-grid`. Applies this to My Tests.

### DIFF
--- a/src/app/components/elements/CardGrid.tsx
+++ b/src/app/components/elements/CardGrid.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { Container, ContainerProps, Row } from 'reactstrap';
+import { below, useDeviceSize } from '../../services';
+
+export const CardGrid = (props: ContainerProps) => {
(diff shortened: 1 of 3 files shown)


mlt47 (commented):
There is a Reactstrap CardDeck component which might achieve what you need here?
https://6-4-0--reactstrap.netlify.app/components/card/#Decks
https://getbootstrap.com/docs/4.0/components/card/#card-decks

mlt47 (approved):
Thanks for pointing out that Card Deck does not satisfy that each row needs to be the same size! I thought it sounded a little too easy to be true.

I'm happy to go with this implementation.

I feel like it might even be achievable using "just CSS" with display: grid, grid-template-rows, grid-template-columns and the `fr` fractional measure, if you fancy learning some more modern CSS! I don't think we've made use of it on the site yet.

mlt47 (approved):

PR #868, opened 2023-12-14
Remove unnecessary redux checks

skyepurchase:
When setting an assignment or quiz checks on whether the due date is valid is done in the modal, the redux thunk, and
the backend. This is inefficient, so the redux thunk checks have been removed.

- Add front end date checks to set assignment
- Remove unnecessary redux date checks


mlt47 (approved):
Removes duplicated logic covered by the back-end validation and UI. Looks good to me!

PR #870, opened 2023-12-14
Clean buggy behaviour on assignment progress table

jacbn:
- Normalise `isQuiz` terminology to be consistent with the rest of the codebase (i.e. is this a _test_), thus fixing discrepancy
- Prevent table overflowing bounding box, scroll instead
- Fix sort buttons being too low


mlt47 (approved):
Great! The assignment progress overflow problem has needed addressing for a while, glad you took the initiative to fix it!

PR #871, opened 2023-12-15
Remove text styling from terms in inline definition

jacbn:
Removes all text styling from glossary terms when they are inside another inline glossary term's definition. Does **not** remove this styling if the parent glossary term is displayed as a block, e.g. in the bottom glossary.


mlt47 (changes requested):
I am happy with the approach of passing through className as we already do so for "plaintext". However the CSS will need expanding to cover the case where there is a link in the glossary so that it is legible (second example here: https://staging.isaacphysics.org/questions/3ece8d8c-b56d-4e60-93f5-ac8a6e17ea8a).

jacbn:
> the CSS will need expanding to cover the case where there is a link in the glossary so that it is legible

Indeed, missed this case. Thanks.

mlt47 (approved):

PR #873, opened 2024-01-02
Replace sketcher UI buttons with divs

jacbn:
The \<button\>s were causing the top-level form to submit, i.e. submitting the question attempt. Replacing these with divs with the button role prevents this.


jsharkey13 (approved):
All still seems to work as expected!

PR #877, opened 2024-01-05
Prevent symbolic question response resetting on reload

skyepurchase:
When a user changes their input the response banner needs to be removed
to allow the user to reassess their attempt. This is done through
`dispatchSetCurrentAttempt` which sets `validationResponse` to
`undefined`.

The issue with symbolic questions is that there are two input types that
need to be kept in sync and both need to be able to set a current
attempt. Text input only does this when there is no input otherwise it
updates symbolic input's state. The symbolic input will always run the
dispatch when it's state changes.

However, the symbolic input's state is set on reload which triggered the
dispatch causing the brief flash of the banner. Adding conditions to
make sure that there has been a meaningful change prevents this.


mlt47 (approved):
Well done in chasing this down the various state changes across files. Looks to work correctly now.

PR #880, opened 2024-01-15
Use carets for sort indicators

skyepurchase:


skyepurchase:
Would be an idea to make the `boardOrder` into two separate variables, one that tracks the column and one that tracks the order (forward or reverse). This would affect setQuizzes pages which are currently being changed and so that change would need to wait.

src/app/components/pages/MyGameboards.tsx
```diff
@@ -65,6 +65,60 @@ const PhyTable = (props: GameboardsTableProps) => {
...
+
+    function sortItem(props: ComponentProps<"th"> & {title: string, itemOrder: BoardOrder, reverseOrder: BoardOrder}) {
+        const {title, itemOrder, reverseOrder, ...rest} = props;
+        const className = (props.className || "text-center align-middle") + sortClass(itemOrder, reverseOrder);
+        const sortArrows = <button className="sort physics" onClick={() => {toggleSort(itemOrder, reverseOrder);}}>
+                <span className="up">▲</span>
+                <span className="down">▼</span>
+            </button>;
+        return <th key={props.key} {...rest} className={className}>{title}{sortArrows}</th>;
+    }
```
mlt47, line 96: As there's so much duplication between this and the sort item in the AdaTable, I would recommend pulling it out into a separate component. As well as being neater it reduces the chance of bugs when one is updated but not the other.

src/app/components/pages/MyGameboards.tsx
```diff
@@ -65,6 +65,60 @@ const PhyTable = (props: GameboardsTableProps) => {
...
+            return " sorted reverse";
+        } else {
+            return "";
+        }
+    }
+
+    function sortItem(props: ComponentProps<"th"> & {title: string, itemOrder: BoardOrder, reverseOrder: BoardOrder}) {
+        const {title, itemOrder, reverseOrder, ...rest} = props;
+        const className = (props.className || "text-center align-middle") + sortClass(itemOrder, reverseOrder);
+        const sortArrows = <button className="sort physics" onClick={() => {toggleSort(itemOrder, reverseOrder);}}>
```
mlt47, line 91: So that the button is given the space it needs in the table cell when being rendered, I would recommend changing this from being `position: absolute` to `position: relative`. It will then require a bit more CSS tweaking to get the correct positioning.

mlt47 (changes requested):
I much prefer these sort indicators :).

I have made some suggestions.

We might also want to experiment with altering the Physics table to be bordered if it is not clear which column label is associated with which sort button.

src/app/components/pages/MyGameboards.tsx
skyepurchase: This change meant that Ada and Physics are the same. There was weird text-alignment for Ada where the table header had `vertical-align: bottom` but I couldn't find that being set anywhere, I added `vertical-align: middle` to keep the alignment consistent across tables.

src/scss/common/pages.scss
```diff
@@ -106,3 +106,43 @@ main {
     border-left-color: $black;
   }
 }
+
+.my-gameboard-table-header {
+    th {
```
mlt47, line 111: The alignment and positioning of the table header text and sort buttons doesn't look correct in my browser.

mlt47 (changes requested):

src/app/components/pages/SetAssignments.tsx
```diff
@@ -71,6 +71,7 @@ import {StyledSelect} from "../elements/inputs/StyledSelect";
 import {PageFragment} from "../elements/PageFragment";
 import {RenderNothing} from "../elements/RenderNothing";
 import { Spacer } from "../elements/Spacer";
+import { sortItem } from "./MyGameboards";
```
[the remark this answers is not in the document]
jsharkey13: (I dislike the `index.ts` approach unless it is unavoidable since it prevents tree-shaking and bundle splitting for anything which then imports it. Making a smaller new component/element as suggested feels cleaner and avoid this).

src/scss/common/pages.scss
skyepurchase: It appears that Firefox was aligning the two elements (the text and the buttons) differently compared to Chrome and Edge. Wrapping it in a div to align the items within the div works consistently across browsers

src/app/components/pages/MyGameboards.tsx
```diff
@@ -65,6 +65,39 @@ const PhyTable = (props: GameboardsTableProps) => {
         boardCompletion, setBoardCompletion, boardCreator, setBoardCreator,
         boardOrder, setBoardOrder
     } = props;
+
+    const tableHeader = <tr className="my-gameboard-table-header">
+        <SortItemHeader key={"completion"} itemOrder={BoardOrder.completion} reverseOrder={BoardOrder["-completion"]} boardOrder={boardOrder} setBoardOrder={setBoardOrder}>
```
mlt47, line 70: We only need `key`s when we're generating a list of components (i.e. using a map where the order could change between renders). Here we are declaratively setting out the components so it is not necessary.

src/app/components/pages/MyGameboards.tsx
```diff
@@ -195,6 +190,39 @@ const CSTable = (props: GameboardsTableProps) => {
         boardCompletion, setBoardCompletion, boardCreator, setBoardCreator,
         boardOrder, setBoardOrder
     } = props;
+
+    const tableHeader = <tr className="my-gameboard-table-header">
```
mlt47, line 194: The col widths on Ada seem to be incorrect on this table, but also worth checking the others.

skyepurchase: With the additional merge conflict from the changes to Stages and Difficulties on Physics this was dealt with but not very obviously

mlt47 (approved):
Works well and the code looks like an idiomatic React solution to the sharing of code which is great.

PR #881, opened 2024-01-16
Fix date input visual bug when pre-selected year out of range

skyepurchase:
- Display the pre-selected year despite of year range
- Fix ESLint warnings


jacbn (approved):
The prefer-const bit might be solvable with a `useCallback` on `value`? Let's leave it as is for now though, it's unrelated to this fix.

PR #888, opened 2024-01-22
Fix 'Allow another attempt' button visibility

skyepurchase:
Changing the conditional on the due date so that if a due date does not exist the 'Allow another attempt' button is
still shown.


jsharkey13 (approved):
This is [the opposite of the "should the request fail" logic in the API](https://github.com/isaacphysics/isaac-api/blob/v3.12.2/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java#L679-L681), and I'm pleased I remember de Morgan's laws!

PR #889, opened 2024-01-22
Adding expansion animation to Physics

skyepurchase:
- Add expansion animation to Physics reorder questions
- Prevent validation banner overflowing card
- Fix ESLint warnings


src/scss/phy/questions.scss
```diff
@@ -46,8 +46,8 @@
     font-size: 1.25rem;
     line-height: 1.6;
   }
-  // Make sure that parsons questions have a background when they expand on PHY
-  .question-component:not(.isaacParsonsQuestion) {
+  // Make sure that reorder questions have a background when they expand on Phy
+  .question-component:not(.isaacReorderQuestion) {
```
mlt47, line 50: Because isaacReorderQuestions are basically the same as isaacParsonsQuestions, won't these style rules need to be applied when the question component is not either?

mlt47 (changes requested):
Requested some small changes to neaten the codebase. However it's worth considering the quiz/test case where these expanded questions are not in accordions and so the animation is less clear/not visible unless the expanded question is the first question in the quiz section.

src/scss/phy/questions.scss
skyepurchase: Looking at the content base there are no isaacParsonsQuestions on Physics so changed this over. Having both is probably best though, you're right

mlt47 (commented):
Looks good and the functional component looks like a good idiomatic solution.

PR #890, opened 2024-01-22
Feature/reintroduce ada events

mlt47:
Add Ada event routes and tweak the CSS to make them look reasonable with Ada's style sheets.

Adding events to the navigation will be handled as a future task, once there are more events to publicise.


jacbn (approved):
Looks good, would probably change the card image css though as it's stretched and quite small at the moment! Something like
```css
.event-card-image {
    img {
      display: block;
      height: 180px;
      object-fit: cover;
      margin: auto;
    }
  }
```
in `cards.scss` would be consistent with the full-width news pods.

mlt47:
Good point about the images, @jacbn. We used to have specially designed circular illustrations for each the different types of events that we ran on Isaac CS, Ada will not be using them.

PR #891, opened 2024-01-23
Remove remaining references to Google analytics

mlt47:
Removes analytics from the unsupported browser pages and our development Content Security Policy.
I will create a card to add Plausible analytics to the unsupported browser pages, we've not been reporting that since moving away from GA.


jsharkey13 (approved):
If we do add Plausible back, then having deleted the JS files will make it a little more tricky to build - so we should keep a reference to this so we can re-use the code later!
(We need separate JS files for the CSP, and `nginx.conf` expects them to be called a specific thing.)

PR #894, opened 2024-01-24
Throttle dirty test attempt submission

jacbn:
Slows down the rate at which test attempt submissions are automatically saved.


mlt47 (approved):
This solution works and is neat. However I tried swapping throttle with debounce locally as we mentioned in the office and it does the expected thing for me - triggers once after the interval. Will push up that change and merge.

PR #900, opened 2024-01-31
Add meta description tags for static physics pages

jsharkey13:


jacbn (approved):
We should ask Ada whether they want an events description sooner rather than later, but the code looks good.

PR #904, opened 2024-02-01
Add tags for Scotland

jacbn:
Support for the new Scottish tags released for the backend and editor, while enabling users to select new stages in My Account.


src/app/components/pages/ExamSpecifications.tsx
```diff
@@ -1,63 +1,73 @@
...
+        }}
+    >
+        {examBoardTabs}
+    </Tabs>;
+
+    useEffect(() => {
+        // Since examBoardTab is last in the dependency chain, we would ideally only have this useEffect run when examBoardTab changes.
+        // However, since exam boards can be shared among stages, switching from one stage to another can result in the same exam board being selected.
+        // In this case, the useEffect wouldn't run at all if we only had examBoardTab as a dependency.
+        // This unfortunately means we have to have all tabs in the dependency, meaning we can generate an invalid request between when
```
mwtrew, line 57: Makes sense, although I wasn't able to make anything strange happen by switching between stages. What happens in this "invalid request" situation?

src/app/components/pages/ExamSpecifications.tsx
```diff
@@ -1,63 +1,73 @@
...
-import {useHistory, useLocation} from "react-router-dom";
+import {CS_EXAM_BOARDS_BY_STAGE, EXAM_BOARD, STAGE, STAGES_CS, stageLabelMap} from "../../services";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
 import {MetaDescription} from "../elements/MetaDescription";
 
 export const ExamSpecifications = () => {
-    const history = useHistory();
-    const location = useLocation();
+    const STAGES_WITH_EXAM_SPECIFICATIONS = useMemo(() => [STAGE.A_LEVEL, STAGE.GCSE], []);
+    const [stageTab, setStageTab] = useState<typeof STAGES_CS[number]>(STAGES_CS.includes(window.location.hash.split("/")[0].slice(1) as typeof STAGES_CS[number]) ? window.location.hash.split("/")[0].slice(1) as typeof STAGES_CS[number] : STAGE.A_LEVEL);
```
mwtrew, line 11: These lines are getting very long, I'd consider extracting them or parts of them to functions for readability. As an aside, line length is something we should enforce in the linter like we do elsewhere.

mwtrew (approved):
Looks good overall, a couple of minor points.

src/app/components/pages/ExamSpecifications.tsx
mwtrew: Now that you mention it, going from GCSE, OCR to A Level switches me back to AQA, which isn't how it used to work. Is that expected?

jacbn: The invalid state was showing a 'fragment couldn't be found' screen very briefly, but it would resolve itself once the stage/exam board had realigned. After another look I seem to have found a way around it, so even if we get into an invalid state it doesn't update until it's realigned. The exam board tabs do seem to jump around a bit when changing the stage tabs though, think it's a desync between the tab override values and which tabs are being displayed via `CS_EXAM_BOARDS_BY_STAGE`. Had a quick tinker but can't get everything to update at once; any thoughts?

mwtrew: Seems better now, thanks! I'm not too worried about the jumping-around issue.

PR #905, opened 2024-02-05
Deploy script improvements

skyepurchase:
A couple of small quality of life improvements during deployment to prevent incorrectly running commands that reasonably should not be run from within python and excessive skipping if tasks aren't required.


jsharkey13 (approved):
This looks okay; will test it for real this week!

PR #907, opened 2024-02-09
Update dependencies and Node build version

jsharkey13:
This version of Babel requires Node as least 18.12, so this upgrades the Actions runner to the latest LTS.
`@babel/plugin-proposal-class-properties` hasn't been a proposal in 4 years, so use the actual transform package now.
Axios is now 1.0, but has changed the type of the response data and so I have used an any as a quick fix.

There are many changes here, I have tested what I could.


jsharkey13:
I have spent an hour and a half fighting to get the tests to build and pass. Axios and uuid caused chaos, and Jest does not play nicely with the older version of React we are using (although not Jest itself, but the JSDom library).
I am going to upgrade just a few core things instead. It is annoying that a testing framework blocks upgrades that otherwise build fine!

PR #908, opened 2024-02-09
Dependency updates

jsharkey13:


mlt47 (approved):
This looks OK to me and works locally.
I used webpack's `ignoreWarnings` property for a warning which we can't fix without upgrading Bootstrap. The Physics build also has a few bootstrap deprecation warnings which we could add to that list of exceptions but for those warnings I'm hopeful that we can find alternative Sass to achieve the same outcome.

PR #915, opened 2024-02-23
Add validation for user contexts on Ada

mwtrew:
* Disables (Ada) or hides (Physics) the add button unless existing rows are already valid
 * Considers the default value valid for Ada, where the field is optional

Also removes superseded `UserDetails` component.


src/app/components/elements/inputs/UserContextAccountInput.tsx
jacbn, line 1: A couple of points:
- Could we make the exam board stay as is if the user updates the stage and the selected stage is still applicable? Something like replacing:
https://github.com/isaacphysics/isaac-react-app/blob/bcf1e8ea3602db377a60343e12db5d42faefb33a/src/app/components/elements/inputs/UserContextAccountInput.tsx#L56-L58 with:
```tsx
        const possibleExamBoards = getFilteredExamBoardOptions(
            {byStages: [stage || STAGE.ALL], byUserContexts: existingUserContexts, includeNullOptions: onlyOneAtThisStage
        }) || [EXAM_BOARD.ALL];
        const examBoard = possibleExamBoards.map(e => e.value).includes(userContext.examBoard as EXAM_BOARD) && userContext.examBoard || possibleExamBoards[0].value;
```
- Second, is there a reason not to also remove the `required` here as well? 
https://github.com/isaacphysics/isaac-react-app/blob/bcf1e8ea3602db377a60343e12db5d42faefb33a/src/app/components/elements/inputs/UserContextAccountInput.tsx#L79-L79 It makes the red (!) appear when the stage is blank but the exam board isn't.

jacbn (approved):
Looks good, have suggested some minor UI improvements if there's time to check them.

PR #917, opened 2024-02-26
Ada accordion adjustments

jacbn:
- Remove "all stages" from Ada, show comma-separated (but no commas between lines) list of all applicable stages
- Reuse GCSE/A Level colours for N5 / Adv Higher, use dark pink for Higher
- Minor styling changes to fit above changes in accordions, remove accordion-h3 underline.


mlt47 (commented):
I'm OK with this solution, especially as we are likely to revisit the feature in the near future. Worth checking that the correct transformation is applied to the TopicSummaryLink strings too.

mlt47 (approved):
Looks good to me and it does seem to improve the differentiation of the headings on the content linked to from the Trello card.

PR #919, opened 2024-03-01
Add link from 404 page to contact form preset

jsharkey13:
This brings Phy and CS into alignment, except on the breadcrumb title where the CS value of "404" remains.

### DIFF
--- a/src/app/components/pages/Contact.tsx
+++ b/src/app/components/pages/Contact.tsx
@@ -37,24 +37,26 @@ const determineUrlQueryPresets = (user?: Immutable<PotentialUser> | null) => {
     let presetMessage = "";
     let presetPlaceholder = "";
 
-    if (urlQuery?.preset == "teacherRequest" && user?.loggedIn && !isTeacherOrAbove(user)) {
+    if (urlQuery?.preset === "teacherRequest" && user?.loggedIn && !isTeacherOrAbove(user)) {
         presetSubject = "Teacher Account Request";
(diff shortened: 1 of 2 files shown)


src/app/components/pages/Contact.tsx
```diff
@@ -37,24 +37,26 @@ const determineUrlQueryPresets = (user?: Immutable<PotentialUser> | null) => {
...
-        else if (urlQuery?.page) {
+        } else if (urlQuery?.page) {
             presetSubject += ` in "${urlQuery.page}"`
         }
         if (urlQuery?.section != null) {
             presetSubject += `, section "${urlQuery.section}"`
         }
+    } else if (urlQuery?.preset === 'notFound') {
+        presetSubject = "Page not found";
+        presetMessage = `Page: "${urlQuery?.page}"\n\n[Add any details about how you found this missing page here.]`;
```
mwtrew, line 59: It might be better to put the URL in the subject line like we do for content problems and use `presetPlaceholder` for these instructions - I think that makes it less likely that they'll delete the URL (which we need) from the message body by accident.

mwtrew (approved):

src/app/components/pages/Contact.tsx
jsharkey13: I wasn't sure about it, because I don't like the really long subject lines; but I'd not considered the fact they might delete the message. I was sort of expecting them to click send without typing anything else (which they wouldn't be able to do if it was only placeholder text).

mwtrew: I'm happy enough to leave it as-is then, it would be cleaner not to let the user edit the info we need but that's more of a problem with using an email form in general and likely won't be an issue in practice.

PR #922, opened 2024-03-05
URI encoding gameboard titles

skyepurchase:
- Add text showing user's account type
- Circumvent isLoggedIn check during sign up
- Copy for my account pages
- Adjust text for Physics
- URI encoding and decoding gameboard title


jsharkey13 (approved):
Looks good to me, though I did help write it 🙈

PR #923, opened 2024-03-05
Move add more button outside board builder table

jsharkey13:
The button used to be inside the scroll region, which meant on smaller screens it was hidden from view, especially on Safari where scrollbars are not shown to the user.
There are no functionality changes, only layout changes.

This might be easiest to review ignoring whitespace!

### DIFF
--- a/src/app/components/pages/GameboardBuilder.tsx
+++ b/src/app/components/pages/GameboardBuilder.tsx
@@ -279,9 +279,9 @@ const GameboardBuilder = ({user}: {user: RegisteredUserDTO}) => {
                     </Col>
                 </Row>}
 
-                <div className="mt-4 responsive">
+                <div className="my-4 responsive">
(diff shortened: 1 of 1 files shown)


src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -279,9 +279,9 @@ const GameboardBuilder = ({user}: {user: RegisteredUserDTO}) => {
                     </Col>
                 </Row>}
 
-                <div className="mt-4 responsive">
+                <div className="my-4 responsive">
                     <DragDropContext onDragEnd={reorder}>
-                        <Table bordered>
+                        <Table bordered className="m-0">
```
jacbn, line 284: The margin at the bottom of the table can be scrolled out of view if it's full; this moves it to the bounding element

jacbn (approved):

PR #925, opened 2024-03-06
Change Ada isIntendedAudience to behave like isOnSpecification

skyepurchase:
Ada wants a more strict check that enforces a stage and examboard to be present on a question (or item of content) for it to be considered on a user's specification.

This means that the `isIntendedAudience` needs to fail early rather than ignore stage and examboard. This is really a different function entirely and we don't necessarily want all uses of `isIntendedAudience` to behave like this on Ada. This should be changed in future after the release and the two functions used appropriately.


jacbn (approved):

src/app/services/userContext.ts
```diff
@@ -366,6 +366,10 @@ export function findAudienceRecordsMatchingPartial(audience: ContentBaseDTO['aud
 export function isIntendedAudience(intendedAudience: ContentBaseDTO['audience'], userContext: UseUserContextReturnType, user: Immutable<PotentialUser> | null): boolean {
     // If no audience is specified, we default to true
```
mwtrew, line 367: Todos have a habit of hanging around for a lot longer than expected, so I would update or remove this comment since it's now misleading.

PR #926, opened 2024-03-07
Change text for top-most 'no content' warning banner

skyepurchase:


src/app/components/pages/Topic.tsx
```diff
@@ -51,7 +51,9 @@ export const Topic = withRouter(({match: {params: {topicName}}}: {match: {params
                     {!(atLeastOne(relatedConceptsForSpecificViewingContext.length) || atLeastOne(relatedQuestionsForSpecificViewingContext.length)) &&
                         <div className='text-center mt-3'>
                             <div className='alert alert-warning'>
-                                {`There is no material in this topic for ${stageLabelMap[userContext.stage]} ${examBoardLabelMap[userContext.examBoard]}.`}
+                                {`There is no material in this topic for ${stageLabelMap[userContext.stage]} ${examBoardLabelMap[userContext.examBoard]}. ` +
```
mwtrew, line 54: This isn't quite the same as the copy on the card, is that intentional?

mwtrew: Also, I think the intention is for the copy on the Topic and Concept pages to be the same.

skyepurchase: Ah I missed the first sentence changing and focused on the second sentence with the link will adjust that now

PR #929, opened 2024-03-12
Context picker mobile layout improvements

jacbn:
Aligns content in mobile, tries to keep consistency between UserContextPicker in all 3 use cases

### DIFF
--- a/src/app/components/elements/Accordion.tsx
+++ b/src/app/components/elements/Accordion.tsx
@@ -9,7 +9,6 @@ import {
     isAQuestionLikeDoc,
     isPhy,
     NOT_FOUND,
-    notRelevantMessage,
     scrollVerticallyIntoView,
     siteSpecific,
     useUserContext
@@ -183,12 +182,6 @@ export const Accordion = withRouter(({id, trustedTitle, index, children, startOp
                                     {trustedTitle}
                                 </Markup>
                             </h2>
-                        </div>}
-                        {isAda && deEmphasised && <div className="ml-auto mr-3 d-flex align-items-center">
-                            <span id={`audience-help-${componentId}`} className="icon-help mx-1" />
-                            <RS.UncontrolledTooltip placement="bottom" target={`audience-help-${componentId}`}>
-                                {`This content has ${notRelevantMessage(userContext)}.`}
-                            </RS.UncontrolledTooltip>
                         </div>}
                         {typeof disabled === "string" && disabled.length > 0 && <div className={"p-3"}>
                             <span id={`disabled-tooltip-${componentId}`} className="icon-help" />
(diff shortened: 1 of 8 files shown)


src/app/components/elements/Accordion.tsx
```diff
@@ -184,12 +183,6 @@ export const Accordion = withRouter(({id, trustedTitle, index, children, startOp
                                 </Markup>
                             </h2>
                         </div>}
-                        {isAda && deEmphasised && <div className="ml-auto mr-3 d-flex align-items-center">
-                            <span id={`audience-help-${componentId}`} className="icon-help mx-1" />
-                            <RS.UncontrolledTooltip placement="bottom" target={`audience-help-${componentId}`}>
-                                {`This content has ${notRelevantMessage(userContext)}.`}
-                            </RS.UncontrolledTooltip>
-                        </div>}
```
jacbn, line 192: Ada does not have UncontrolledTooltips anymore, this was creating an empty span

src/app/components/elements/inputs/UserContextPicker.tsx
```diff
@@ -42,79 +42,84 @@ export const UserContextPicker = ({className, hideLabels = true}: {className?: s
...
+                {showHideOtherContentSelector && <FormGroup className={`d-flex align-items-center m-0 float-lg-right ${className}`}>
+                    <Label className="d-inline-block m-0" htmlFor="uc-show-other-content-check">Show other content? </Label>
+                    <CustomInput
+                        className="w-auto d-inline-block ml-2 pr-0" type="checkbox" id="uc-show-other-content-check"
+                        checked={userContext.showOtherContent}
+                        onChange={e => dispatch(transientUserContextSlice.actions.setShowOtherContent(e.target.checked))}
+                    />
+                </FormGroup>}
+            </RS.Col>
+            <RS.Col xs={12} lg={7} className={`d-flex m-0 mb-2 p-0 justify-content-lg-end ${className}`}>
```
jacbn, line 58: I am not convinced by this extra ${className}. It shouldn't be there, but it is the easiest way I could find to make topic pages keep the right-justified context picker in `md` display size (the only case where we want this behaviour). This inconsistency is required because we have no 'share' or 'flag' buttons on these pages; is there a reason for this?

src/scss/common/accordions.scss
```diff
@@ -12,6 +12,11 @@
   min-width: 130px;
   max-width: 130px;
   text-align: center;
+
+  @media (max-width: 575px) {
```
jsharkey13, line 16: This can be done with a Bootstrap mixin that doesn't rely on hard-coding the current Bootstrap pixel value:
https://getbootstrap.com/docs/4.6/layout/overview/#responsive-breakpoints

I think:

```suggestion
  @include media-breakpoint-down(xs)
```
is exactly equivalent.

src/scss/common/pages.scss
```diff
@@ -146,3 +146,8 @@ main {
     }
 }
 
+.context-picker-container {
+  @media (min-width: 992px) {
```
jsharkey13, line 150: Ditto as to the above, I think something like:
```suggestion
  @include media-breakpoint-up(lg) {
```
is equivalent?

jsharkey13 (commented):
Seems to work fine for me.

I don't like the hardcoded responsive media queries; there's a Bootstrap way of doing it that we use elsewhere, I have added a comment.

Let's not worry about the Share/Report buttons for Topics for now.

PR #933, opened 2024-03-20
Question Finder UI using question search endpoint

skyepurchase:
This is the first step in creating the new Question Finder as this page uses the desired endpoint which can be changed and behaves closely to how the final page with behave.

This converts the `QuestionSearchModal` (from the `GameboardBuilder`) into a fully fledged page but instead uses the `Search` results panel to display the returned questions. It additionally stores the filter parameters in the URL to autofill when back tracking.

There is a large refactor commit which affects other files not related to the Question Finder.


src/app/components/pages/QuestionFinder.tsx
```diff
@@ -0,0 +1,309 @@
...
+                        </RS.Label>
+                    </RS.Form>
+                </RS.Col>}
+            </RS.CardHeader>
+            <Suspense fallback={<Loading/>}>
+                <RS.CardBody className={classNames({"p-0 m-0": isAda})}>
+                    <ShowLoading until={sortedQuestions}>
+                        {sortedQuestions ?
+                        <LinkToContentSummaryList items={sortedQuestions}/> :
+                        <em>No results found</em>}
```
mlt47, line 303: We should also have an informative message when the user has not selected any filter options.

src/app/components/pages/QuestionFinder.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+            dispatch(searchQuestions({
+                searchString: searchString,
+                tags,
+                stages: stages.join(",") || undefined,
+                difficulties: difficulties.join(",") || undefined,
+                examBoards: examBoardString,
+                fasttrack,
+                startIndex,
+                limit: 30
```
mlt47, line 130: We hardcode the limit to 30 here. This is OK for now but is not the approach we want to take long term.
Do we want to at least indicate that there are more matches for the time being? `totalMatches` is returned with the response.

src/app/components/pages/QuestionFinder.tsx
```diff
@@ -0,0 +1,309 @@
...
+        {value: "phys_book_gcse", label: "GCSE Physics"},
+        {value: "physics_skills_19", label: "A Level Physics (3rd Edition)"},
+        {value: "physics_linking_concepts", label: "Linking Concepts in Pre-Uni Physics"},
+        {value: "maths_book_gcse", label: "GCSE Maths"},
+        {value: "maths_book", label: "Pre-Uni Maths"},
+        {value: "chemistry_16", label: "A-Level Physical Chemistry"}
+    ];
+
+    useEffect(() => {
+        searchDebounce(searchQuery, searchTopics, searchExamBoards, searchBook, searchStages, searchDifficulties, searchFastTrack, 0);
```
mlt47, line 157: We get a URIError: URI malformed if we search for a character followed by `%` as the search string.

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -40,8 +41,11 @@ const GraphSketcherPage = lazy(() => import("../../pages/GraphSketcher"));
 let key = 0;
 export const RoutesPhy = [
     // Registration
-    <TrackedRoute exact path="/register" component={Registration} />,
-    <TrackedRoute exact path="/verifyemail" component={EmailAlterHandler}/>,
+    <TrackedRoute key={key++} exact path="/register" component={Registration} />,
+    <TrackedRoute key={key++} exact path="/verifyemail" component={EmailAlterHandler}/>,
+
+    // Questions
+    <TrackedRoute key={key++} exact path="/questions" component={QuestionFinder} />,
```
mlt47, line 48: I like /questions as the route for this page but do we want it to be located at a path which is as prominent and guessable as /questions while the page is still under active development?

src/app/services/filter.ts
```diff
@@ -0,0 +1,16 @@
+export function arrayFromPossibleCsv(queryParamValue: string | string[] | undefined): string[] {
+    return queryParamValue ?
+        (queryParamValue instanceof Array ?
+            queryParamValue :
+            queryParamValue.split(",")
+        ) :
+        [];
+}
+
+export function toCSV<T extends string>(items: T[]) {
```
mlt47, line 10: Perhaps renaming this to `toSimpleCsv(...)` would be a bit more informative considering we're not doing any escaping of commas and newlines.

mlt47 (changes requested):
This already seems like quite a useful page for students.
I have made some comments to consider or address.

mlt47 (approved):

PR #934, opened 2024-03-20
Show related questions without examboard as off specification

skyepurchase:
If a related question only has a stage (no examboard) it will be
considered "off specification" regardless of what examboard the
user has.

Furthermore, if there are related questions they will always be
displayed regardless of whether any are "on you specification".


mlt47 (approved):
This code change matches what we had agreed upon with the content teams.

PR #935, opened 2024-03-21
Add inline question type

jacbn:
Adds a multi-part inline question "type" to the frontend. Note that this does _not_ introduce a new question type; rather, this exists as a wrapper around existing question types to bundle them into one submission.

When loading a question page from the backend, if an inline region is found, the portals trick used for Cloze questions and tables is used to convert instances of "[inline-question:{id}]" to an InlineXXXEntryZone, based on the inline part type set in the editor. For example, if a String Match question is set with id `inline-question:taxonomy-q1`, a InlineStringEntryZone is rendered in the frontend, which exists as an interface for answering the IsaacStringMatchQuestion with id `{inline-region-id}|taxonomy-q1`.

A context wrapper stores properties of the inline region, most notably a map between element ID and question ID, used frequently to obtain the DTO related to the inline entry zone (e.g. for retrieving previous answers). 

```
InlineContext
  -> InlineRegion
    -> InlineEntryZone
      -> InlineStringEntryZone
      -> InlineNumericEntryZone
```

A particularly useful test page can be found at `/questions/f61fc19e-043e-464d-a48f-6af561a00f5f`.


mlt47 (commented):
There's a lot of good work here. It's a complicated bit of code but an useful feature.
I don't think the changes suggested needs to stop it from being merged but as we discussed at the meeting the changes can be made over the next couple of weeks.

PR #936, opened 2024-03-21
Restore prominent upgrade role link to My Account

jsharkey13:
Annoyingly the Ada designers want bolded links, which means a bunch of siteSpecifics. Ideas welcome on how to fix this/make it nicer!
Also remove the existing text next to the context picker since it was only shown to students but only helpful to teachers. 
Role names are now lowercase, but these are only used by the AccountTypeMessage component anyhow.

### DIFF
--- a/src/app/components/elements/AccountTypeMessage.tsx
+++ b/src/app/components/elements/AccountTypeMessage.tsx
@@ -4,25 +4,17 @@ import React from "react";
 import {UserRole} from "../../../IsaacApiTypes";
 
 
-export const AccountTypeMessage = ({role}: { role?: UserRole }) => {
+export const AccountTypeMessage = ({role, hideUpgradeMessage}: { role?: UserRole, hideUpgradeMessage?: boolean }) => {
     return <div>
-        Account type: <b>{role && UserFacingRole[role]}</b> {role == "STUDENT" &&
-        siteSpecific(
-            <span>
-                    <small>(Are you a teacher or tutor? {" "}
-                        <Link to={TEACHER_REQUEST_ROUTE} target="_blank">
-                            Upgrade your account
-                        </Link>{".)"}
-                    </small>
-            </span>,
-            <span>
-                    <p className="font-style-italic">Are you a teacher or tutor? {" "}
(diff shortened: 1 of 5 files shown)


src/app/components/elements/AccountTypeMessage.tsx
```diff
@@ -4,25 +4,17 @@ import React from "react";
...
-                    </p>
+        You have a <strong>{role ? UserFacingRole[role] : ""}</strong> account.
+        {role === "STUDENT" && !hideUpgradeMessage &&
+            <span> If you are a teacher{siteSpecific(" or tutor", "")}, you can {" "}
+                {siteSpecific(
+                    <Link to={TEACHER_REQUEST_ROUTE} target="_blank">upgrade your account</Link>,
+                    <strong><Link to={TEACHER_REQUEST_ROUTE} target="_blank">upgrade your account</Link></strong>
+                )}
+                {"."}
             </span>
```
jacbn, line 17: I doubt we'll ever see this currently as role should always be set but if, somehow, it isn't, this reads `You have a  account`. It's probably better to hide the whole text in this case, e.g.:

```suggestion
    {role && <>
        You have a <strong>{UserFacingRole[role]}</strong> account.
        {role === "STUDENT" && !hideUpgradeMessage &&
            <span> If you are a teacher{siteSpecific(" or tutor", "")}, you can {" "}
                {siteSpecific(
                    <Link to={TEACHER_REQUEST_ROUTE} target="_blank">upgrade your account</Link>,
                    <strong><Link to={TEACHER_REQUEST_ROUTE} target="_blank">upgrade your account</Link></strong>
                )}
                {"."}
            </span>
        }
    </>
```

jacbn (commented):
One comment on if role is undefined; otherwise looks good, agree that there's not a lot we can do about the siteSpecific <strong>.

src/app/components/elements/AccountTypeMessage.tsx
jsharkey13: Yes, I saw that that was what it did (and has always done!). I am sure it is not possible, but one extra check does no harm!

PR #938, opened 2024-03-22
Safari Cloze DND fix

jacbn:
`window.scrollBy` seems to move elements in the DOM around incorrectly on Safari -- they appear visually in the correct place, but inspecting them shows the layout boxes do not line up with these visuals. This solves itself if you scroll even a tiny bit, but if you don't, all the dnd fields are offset. 

Seemingly `window.scrollTo` is not subject to the same curse, so this replaces scrollBy with its equivalent scrollTo.


skyepurchase (approved):
LGTM. Working on Safari, Chrome, Firefox, as well as Firefox on mobile

PR #942, opened 2024-04-02
Add ability for student self-removal from groups

jacbn:
Allows self-removal from groups for which this is enabled. This can be set via a new checkbox on the Manage Groups page.

### DIFF
--- a/src/app/components/elements/panels/TeacherConnections.tsx
+++ b/src/app/components/elements/panels/TeacherConnections.tsx
@@ -318,7 +318,7 @@ export const TeacherConnections = ({user, authToken, editingOtherUser, userToEdi
                                                     :
                                                     <span><b>{(membership.group.groupName ?? "Group " + membership.group.id)}</b></span>
                                                 }
-                                                {membership.group.selfRemoval && <img className="self-removal-group ml-1" src="/assets/phy/teacher_features_sprite.svg#groups" alt=""/>}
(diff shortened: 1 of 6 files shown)


src/app/state/slices/api/groupsApi.ts
```diff
@@ -162,21 +162,28 @@ export const groupsApi = assignmentsApi.injectEndpoints({
...
+                                    .map(g => g.id === groupId
+                                        ? {...g, members: g.members?.filter(m => m.id !== userId)}
+                                        : g
+                                    )
                         ));
                     });
                 },
-                errorTitle: "Failed to delete member"
+                successTitle: "Removal successful",
+                successMessage: "You have been removed from the group.",
```
mwtrew, line 184: This logic is still used by teachers to remove students through Manage Groups, so this message would be misleading.

src/app/components/elements/panels/TeacherConnections.tsx
```diff
@@ -316,6 +318,7 @@ export const TeacherConnections = ({user, authToken, editingOtherUser, userToEdi
                                                     :
                                                     <span><b>{(membership.group.groupName ?? "Group " + membership.group.id)}</b></span>
                                                 }
+                                                {membership.group.selfRemoval && <img className="self-removal-group ml-1" src="/assets/phy/teacher_features_sprite.svg#groups" alt=""/>}
```
mwtrew, line 321: I think adding the icon is nice, but this one doesn't work for CS.

mwtrew (changes requested):
Works well, a couple of minor points relating to presentation.

PR #943, opened 2024-04-03
Add support for titled hints

skyepurchase:


jacbn:
If there are a mix of titled/untitled hints (e.g. `/questions/621bbdf4-21dc-40e9-a276-2fc928319da9`) the untitled ones should probably start from Hint 1 and increment only after each untitled hint.

src/app/components/content/IsaacHints.tsx
```diff
@@ -61,14 +61,30 @@ export const IsaacTabbedHints = ({hints, questionPartId}: HintsProps) => {
...
+            }
+            setTitles(newTitles);
+        }
+    }, [hints]);
+
     return <div className="tabbed-hints">
         {hints && <Tabs onActiveTabChange={logHintView} className="no-print" tabTitleClass="hint-tab-title" tabContentClass="mt-1" deselectable activeTabOverride={-1}>
             {Object.assign({}, ...hints.map((hint, index) => ({
-                [`Hint\u00A0${index + 1}`]: <div className="mt-3 mt-lg-4 pt-2">
+                [titles[index]]: <div className="mt-3 mt-lg-4 pt-2">
```
jacbn, line 83: This looks slightly dubious as there's no check to ensure `titles` is the same length as `hints`, however since `hints` is in the `useEffect` above, if `hints` ever changes then `titles` must do so too, to something that must be the same length as `hints`. So all good as far as I can tell.

jacbn (approved):

PR #944, opened 2024-04-03
Context Picker on Physics Search, layout improvements

jacbn:
Adds a context picker for physics search, allowing a workaround for issues with only the first context being displayed. Also restructures the code to hugely simplify both how the UserContextPicker is used and the produced DOM by preferring the typical Bootstrap-style format.

Note for reviewer, the context picker is (now) used in 3 places on both sites:
- search
- concept pages
- topic summary pages

### DIFF
--- a/src/app/components/elements/inputs/UserContextPicker.tsx
+++ b/src/app/components/elements/inputs/UserContextPicker.tsx
@@ -36,7 +36,6 @@ export const UserContextPicker = ({className, hideLabels = true}: {className?: s
         examBoard: isAda && !filteredExamBoardOptions.map(s => s.value).includes(userContext.examBoard),
     };
     const showUnusualContextMessage = unusual.stage || unusual.examBoard;
-    const showHideOtherContentSelector = isAda;
     const showStageSelector = getFilteredStageOptions({byUser: user}).length > 1 || showUnusualContextMessage;
     const showExamBoardSelector = isAda && (getFilteredExamBoardOptions({byUser: user}).length > 1 || showUnusualContextMessage);
 
@@ -116,8 +115,8 @@ export const UserContextPicker = ({className, hideLabels = true}: {className?: s
         </RS.Row>
         
 
-        {/* Other content Selector */}
-        {showHideOtherContentSelector && <RS.Row className="w-100 px-0 m-0 pb-2 justify-content-end">
+        {/* "Show other content" selector */}
+        {isAda && <RS.Row className="w-100 px-0 m-0 pb-2 justify-content-end">
             <FormGroup className="w-auto m-0">
                 <Label className="d-inline-block m-0" htmlFor="uc-show-other-content-check">Show other content? </Label>
                 <CustomInput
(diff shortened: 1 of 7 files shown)


src/app/components/elements/inputs/UserContextPicker.tsx
```diff
@@ -37,33 +36,22 @@ export const UserContextPicker = ({className, hideLabels = true}: {className?: s
         examBoard: isAda && !filteredExamBoardOptions.map(s => s.value).includes(userContext.examBoard),
     };
     const showUnusualContextMessage = unusual.stage || unusual.examBoard;
-    const showHideOtherContentSelector = isAda && segueEnvironment === "DEV";
+    const showHideOtherContentSelector = isAda;
```
mwtrew, line 39: `showOtherContentSelector` would be a better name - as-is it's not immediately obvious what a value of `true` means.

src/app/components/pages/Concept.tsx
```diff
@@ -52,19 +66,12 @@ export const Concept = withRouter(({match: {params}, location: {search}, concept
...
-                    <div className="question-actions not-mobile">
-                        <PrintButton />
-                    </div>
-                    <div className="question-actions">
-                        <ReportButton pageId={conceptId}/>
-                    </div>
+
+                {below["sm"](deviceSize) && <ManageButtons />}
+
+                <div className="d-flex mr-sm-1 flex-grow-1">
```
mwtrew, line 72: Can we also center-align the picker and manage buttons? I think that would look a bit neater. 
image

src/app/components/pages/Search.tsx
```diff
@@ -128,26 +128,30 @@ export const Search = withRouter((props: RouteComponentProps) => {
...
-                                        styles={selectStyle}
-                                    />
-                                    {isAda && <RS.Label className="mt-2 mb-2 mb-md-0 mx-0">
-                                        <UserContextPicker className="text-right" />
-                                    </RS.Label>}
+                                    <RS.Row className="w-100 align-items-center justify-content-end m-0">
+                                        <RS.Label htmlFor="document-filter" className="d-none d-lg-inline-block mr-1">
+                                            {`Filter${siteSpecific("","s")}:`}
+                                        </RS.Label>
+                                        <div className="search-filters-select-containter">
```
mwtrew, line 135: This is working, but worth fixing the typo in both places.

mwtrew (approved):
Looks good, quite a bit neater on small devices too.

src/app/components/elements/inputs/UserContextPicker.tsx
jacbn: Sure, to be honest it's only used in one place and _is_ just `isAda` so I've removed it. The main thing I wanted to flag here is that this is now going to show on live, so you can change whether it's using your user setting or not from this context picker; just checking that's okay.

mwtrew: Yep, I think that's fine.

PR #947, opened 2024-04-08
Question finder URL not updating when field removed

jacbn:
Provides a temporary solution for the URL parameters not updating as stage/topic/fields in the Question Finder page are removed from the search.

Seemingly, `history.replace` does not update a search query if the new parameters are a strict subset of the ones provided previously; this solution therefore resets them before replacing.

A much better solution exists with `react-router-dom` v6, which we should consider upgrading to alongside the planned React upgrade.

### DIFF
--- a/src/app/services/userContext.ts
+++ b/src/app/services/userContext.ts
@@ -153,7 +153,6 @@ export function useUserContext(): UseUserContextReturnType {
                 history.replace({
                     ...window.location,
                     search: queryString.stringify({
-                        ...queryParams,
                         ...actualParams,
                         stage,
                         examBoard: isAda ? examBoard : undefined,
@@ -165,7 +164,7 @@ export function useUserContext(): UseUserContextReturnType {
                 // trying to render, causing a loop and a spike in client-side errors.
             }
         }
-    }, [stage, examBoard, queryParams.stage, queryParams.examBoard]);
+    }, [stage, examBoard]);
 
     return {
         stage, setStage, examBoard, setExamBoard, explanation,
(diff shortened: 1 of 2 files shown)


src/app/services/userContext.ts
```diff
@@ -153,7 +153,6 @@ export function useUserContext(): UseUserContextReturnType {
                 history.replace({
                     ...window.location,
                     search: queryString.stringify({
-                        ...queryParams,
```
mlt47, line 156: I believe that this `useEffect` update was the actual cause of this problem.
`useUserContext` is called all over the place as it is used when we render any content-developer-written content so it is quite hard to trace exactly what happens here. But, when the gameboard is returned and updates the query params correctly, this `useEffect` is later queued up to run after the render as `queryParams.stage` has a different reference. When we do the history replace here `queryParams` holds the old query parameters which then get overwritten by the more recent `actualParams` but this fails when a query param has been removed as is the case in this reported bug.
I'm 99% sure it is safe to remove these dependencies from the `useEffect`. I've tested changing the user context selector and tested the global search functionality, which also reflects filter choices with params, and they all behaved correctly. I don't think we soft-change the stage and exam board query params or that it is even possible without retriggering `useQueryParams`. I think this just reduces the number of times we try to replace the history, which is an improvement.

skyepurchase (approved):
LGTM. I have done some local testing on pages that use `useUserContext` and have found no issues. I am curious as to why queryParams was being passed to the `useEffect` and the `replace` function in the first place.

PR #949, opened 2024-04-15
Add hash anchors and name filtering to My Tests

jacbn:
- Adds `#in-progress`, `#completed` and `#practice` anchors to My Tests
- Adds a filter-by-name box to #practice, with a sharable link


mwtrew (approved):
UX-wise I'm not sure if it's particularly clear to the user that they're copying a link to the filtered results - I'd have recommended a tooltip but it sounds like [we're phasing those out](https://trello.com/c/eKsubwRw/4293-phase-out-hoverable-info-tooltips). Maybe worth re-visiting later.

PR #950, opened 2024-04-17
Events stage filter bug fixes

jacbn:
- Adds All Stages option, defaults filter to this instead of GCSE
- Adds the remaining stages (Year 7&8, Year 9)


mlt47 (approved):
This looks good. Good code reuse.

PR #952, opened 2024-04-17
Correctly update group when removing user

jacbn:
Invalidates the Groups tag in RTK as to pull the updated group data when removing a user. This prevents the need for filtering groups after making the request, which was previously buggy.


mlt47 (approved):
This fixes the problem noted in the card.

PR #953, opened 2024-04-18
Cleanup inline questions code

jacbn:
Fixes some issues with the recent addition of inline questions:
- Moves various function definitions or exports around to reduce the change of circular dependencies
- Lightens the fully-incorrect feedback panel
- Prevents input fields being uncontrolled by default
- Fixes a bug with loading incorrect question attempts on page reload


mlt47 (approved):
This is neater and fixes the bug.

PR #958, opened 2024-04-23
Asset cleanup

jacbn:
Cleans up the `/assets/` folder.

Main changes:
- There now exist precisely 3 folders at the top level, `common/`, `cs/` and `phy/`.
- Each of these contain a subset of {`fonts/`, `icons/`, `logos/` and `favicon/`}, depending on use case, alongside various site-specific folders or assets that do not fit in one of these categories. These folders contain the bulk of the assets.
- **All\* unused assets have been deleted**. Note that while some files -- e.g. `greentick.png` -- are unused in the frontend repository, they _are_ used in the content; such files are untouched. \*(Assets that we enable/disable for e.g. maintenance periods have been retained)
- Some (likely not all) unused .tsx and .scss files which referenced otherwise-unused assets have been deleted. A full list of such files is below:

```
src/app/components/elements/FeaturedContentTabs.tsx
src/app/components/elements/cards/ActionCard.tsx
src/app/components/elements/cards/LinkCard.tsx
src/app/components/elements/svg/Diamond.tsx
src/app/components/elements/svg/LevelsFilter.tsx
src/app/components/elements/svg/Octagon.tsx
src/app/components/pages/Students.tsx
src/app/components/pages/TeacherTools.tsx
src/app/components/pages/books/Workbook20AQA.tsx
src/app/components/pages/books/Workbook20OCR.tsx
src/scss/common/tables.scss
src/scss/cs/preload.scss
src/scss/phy/preload.scss
```


src/scss/common/inequality-modal.scss
```diff
@@ -137,7 +136,6 @@
                 bottom: max(105px, 45%);
                 right: 5px;
                 background-color: transparent;
-                background-image: url('/assets/inequality-trash-phy.svg');
```
jsharkey13, line 140: All three lines of deletion in this file look wrong to me, but this one definitely removes the only visual indication that there is a bin on the editor. The functionality is still there, it's just invisible now!

src/app/state/store.tsx
```diff
@@ -18,10 +18,10 @@ export const middleware: Middleware[] = [
     notificationCheckerMiddleware
 ];
 const defaultMiddlewareOptions = {
-    serializableCheck: process.env.NODE_ENV !== 'test',
+    // serializableCheck: process.env.NODE_ENV !== 'test',
     // Make local dev actually responsive?
-    // serializableCheck: false,
-    // immutableCheck: false,
+    serializableCheck: false,
+    immutableCheck: false,
```
jsharkey13, line 24: However nice this makes local dev, still probably shouldn't be here 😅

jsharkey13 (changes requested):

jsharkey13 (approved):

PR #969, opened 2024-05-02
Add visual filter support to Physics QuestionFinder

skyepurchase:
On Physics the visual hexagonal filter is added.

Questions are filtered on subject, field, and topic separately rather than a as a set of tags. This allows for subject, field and topic to be required tags in the question if they are specified. If multiple tags are specified for each tier only one of the the tags must match, that is it acts like an OR.


jacbn (approved):
Looks good, the stage/exam board filters work much better now.

PR #971, opened 2024-05-07
Add unanswered styling to inline questions

jacbn:
If a question is unanswered, it should display with a yellow outline.
- However, any boxes that are empty on load should remain blank, even if a (empty) bestAttempt exists

Additionally, resubmitting an answer will no longer generate another API request, instead reusing the previous response
- An attempt that has not been modified since load still needs to be submitted once (even if empty), as there may be multiple parts to the question (e.g. numerics have `value` and `units` field) and the only one that is guaranteed is `value`; for these multiple part questions, however, `value` being blank does not been it should not be marked as e.g. `units` may have changed.


src/app/components/elements/inputs/InlineNumericEntryZone.tsx
skyepurchase, line 1: The unanswered feedback looks good for the input boxes. The only problem is [the unit boxes](https://github.com/isaacphysics/isaac-react-app/pull/971/files#diff-69d031734be510424be7815f2bdc0f86d29fbb4002a5b34c912e8bb8f2df6b23L72) which can only be correct or incorrect. This results in odd behaviour where the units have the red crosses but the questions are yellow even if unanswered.

skyepurchase (changes requested):
LGTM. Would just request that the units box has a the unanswered feedback as well

skyepurchase (approved):

PR #976, opened 2024-05-13
Prevent screenreader page jump on question submission

skyepurchase:
After submission the focus on the page resets to `body` which means
using a screenreader (or tabbing) takes the user to the top of the page.

As the submit button disappears after a correct submission the response
banner is focused instead. This way the user can continue from the
question answer. This does require making the `div` tabbable,
which should ideally mean that element is responsive (which it is not),
as otherwise it cannot be focused on.

### DIFF
--- a/src/app/components/content/IsaacQuestion.tsx
+++ b/src/app/components/content/IsaacQuestion.tsx
@@ -1,4 +1,4 @@
-import React, {Suspense, useContext, useEffect} from "react";
+import React, {Suspense, useContext, useEffect, useRef} from "react";
 import {
     deregisterQuestions,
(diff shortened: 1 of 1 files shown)


jsharkey13 (changes requested):
This is looking pretty good now, but there is one major problem; on pages with a previously-answered part, one of the previously answered parts (they race!) will steal the focus from the title element on pageload.

The problem is that the useEffect runs even before the first attempt, because a bestAttempt leads to a preexisting validationResponse. We need to stop it from running until at least one submission has occurred too.

jsharkey13 (approved):

PR #978, opened 2024-05-15
Respect 'Show other content' user preference

skyepurchase:


skyepurchase:
Note that `ContentListGroupItem` already checks `isIntendedAudience` to grey out the search results. This means the only change required was not filtering results if `showOtherContent` is set.

mlt47 (approved):

PR #993, opened 2024-06-10
Show Question Finder loading interstitial

jacbn:
Show "Searching..." interstitial when searching in the new question finder.

### DIFF
--- a/src/app/components/handlers/ShowLoading.tsx
+++ b/src/app/components/handlers/ShowLoading.tsx
@@ -32,7 +32,7 @@ export const ShowLoading = <T extends {}>({until, children, thenRender, placehol
                 clearTimeout(timeout);
             }
         };
-    }, [until, children, thenRender]);
+    }, [until, thenRender]);
 
     switch(until) {
         case null:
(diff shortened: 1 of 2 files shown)


src/app/components/handlers/ShowLoading.tsx
```diff
@@ -32,7 +32,7 @@ export const ShowLoading = <T extends {}>({until, children, thenRender, placehol
                 clearTimeout(timeout);
             }
         };
-    }, [until, children, thenRender]);
+    }, [until]);
```
jacbn, line 35: This is much better (previously the placeholder appeared, disappeared for 200ms, reappeared, then finally once  done, showed the content), but there is still one render cycle at the start where "Searching..." can appear before `duringLoad` is set to true and it switches to the empty space for 200ms.

I think this is a side-effect of the order of the render cycle; ideally (and as is the case most of the time) `until` would always set to undefined after the render but before the `useEffect`, but if this is not the case for one render it is out of sync.

skyepurchase (approved):

PR #997, opened 2024-06-12
Fix jest memory and timeout issues on windows

skyepurchase:
This prevents idle workers from consuming memory that they do not need preventing the tests from successfully completing. This slows down the tests causing some to go over the default time limit of 10000ms (10s).


jsharkey13 (commented):
There is a `jest.config.common.js` file with common config; given these are the same, they can go in there?

jsharkey13:
It does work for me, though, and CI still runs!

jsharkey13 (approved):

PR #1001, opened 2024-06-13
Allow deploy script on the standby server

skyepurchase:
The deploy script currently expects previously running containers to exist. It both queries these containers and brings them down.

The standby server has no containers running and therefore the script fails when attempted on the standby server. This PR skips commands that use the old containers when it detects they are not running and adjust commands as necessary.

Support for staging, test, and dev have not been added as these are unlikely to be required on the standby server.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -121,35 +121,44 @@ def build_docker_image_for_version(ctx):
 
 
 def get_old_versions(ctx):
+    if not ctx['previous_servers_exist']:
+        return
+
     print("# Finding old versions")
     app_name_prefix = f"{ctx['site']}-app-live-"
     previous_app_version = ""
 
     if 'old_app' not in ctx or ctx['old_api'] is None:
-        print("Find the previous app version")
-        automatic = 'y' == input("If docker containers are running this can be done automatically! Run automatically? [y/n]").lower()
-        if automatic:
-            while previous_app_version == "":
-                previous_app_version = ask_to_run_command(
-                    "docker ps --format '{{.Names}}' | " + f"grep {app_name_prefix} | cut -c{len(app_name_prefix) + 1}-"
-                ).rstrip()
-            ctx['old_app'] = previous_app_version
-        else:
-            ctx['old_app'] = input(f"Enter old APP version (e.g. v1.2.3) for {ctx['site']} {ctx['env']}: ")
-            previous_app_version = ctx['old_app']
+        print("# Find the previous app version")
+        while previous_app_version == "":
+            previous_app_version = ask_to_run_command(
+                "docker ps --format '{{.Names}}' | " + f"grep {app_name_prefix} | cut -c{len(app_name_prefix) + 1}-"
+            )
+
+            if previous_app_version is not None:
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -282,6 +305,25 @@ def deploy_etl(ctx):
 
     build_docker_image_for_version(context)
 
+    print("\n# DETERMINING WHETHER OLD SERVICES RUNNING\nMay return exit code 1.")
```
mwtrew, line 308: I'm somewhat intimidated by the random **ALL CAPS**.

deploy.py
```diff
@@ -290,11 +332,14 @@ def deploy_etl(ctx):
...
-            context['env'] = 'staging'
-            deploy_staging_or_dev(context)
+            if context['previous_servers_exist']:
+                print("# Bring down test containers")
+                context['env'] = 'test'
+                bring_down_any_existing_containers(context)
+                context['env'] = 'staging'
+                deploy_staging_or_dev(context)
+            else:
+                print("No support for deploying staging when no containers are running!")
```
mwtrew, line 342: This is kind of confusing, it seems like you *can* deploy `staging` by itself but not along with `live`. Why is that?

mwtrew (approved):
Looks fine overall - I take it you've tried this out on standby?

deploy.py
mwtrew: > Support for staging, test, and dev have not been added as these are unlikely to be required on the standby server.
One further comment, `test` ideally would be supported. If we're doing a full deployment cycle on standby for whatever reason, we should still be running the E2E tests and doing a quick manual check on test before deploying live.

skyepurchase: It's unlikely that `staging` would be wanted on the standby server but if we are building from scratch this could be useful. I have tweaked the code to allow for all environments to be deployed, the changes were simpler than I anticipated.

deploy.py
```diff
@@ -188,8 +202,49 @@ def bring_up_the_new_containers(ctx):
...
+        "docker ps --format '{{.Names}}' | grep app-live",
+        expected_nonzero_exit_codes=[1],
+        run_anyway=True
+    )
+    previous_servers_exist = api_running != "" and app_running != ""
+    ctx['previous_servers_exist'] = previous_servers_exist
+
+    if not previous_servers_exist:
+        print("# OLD CONTAINERS NOT FOUND.")
+        print("\n# ! THIS SCRIPT WILL NOT TAKE DOWN ALL RUNNING CONTAINERS SO CLEAN UP OLD CONTAINERS AFTER !\n")
```
mwtrew, line 223: > ! THIS SCRIPT WILL NOT TAKE DOWN ALL RUNNING CONTAINERS SO CLEAN UP OLD CONTAINERS AFTER !

I'm not sure what the implication of this is - what exactly are you telling the user to do?

skyepurchase: Usually the script asks to take down all containers and bring up new ones (mainly during deploying live) if the script only checks for live containers (as these are the ones that cause failures) so these other containers would remain up. It's not the end of the world but it's best to check that unnecessary containers are not still up

PR #1006, opened 2024-06-20
Use dropdown menus for cloze questions on mobile

skyepurchase:
When the screen is small (e.g. a mobile device) the drop zones are converted to dropdown menus with the available items. The item tray is removed.

If there is no replacement then selecting an item removes it from other dropdowns. A blank option is added to allow for resetting individual zones.

This requires a new handler function in the IsaacClozeQuestion component that is passed to the InlineDropZone rendering component through the DropRegionContext. This is an altered version of the onDragEnd handler.

### DIFF
--- a/src/IsaacAppTypes.tsx
+++ b/src/IsaacAppTypes.tsx
@@ -31,6 +31,7 @@ import {
     TAG_LEVEL
 } from "./app/services";
 import {Immutable} from "immer";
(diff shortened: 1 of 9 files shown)


jacbn (approved):
We should keep an eye on whether we also get requests for this on tablets (up to `lg`), but lgtm. It might be preferable to switch the colours on Ada to similar to that of phy, but this is non-blocking.

PR #1009, opened 2024-06-25
New academic year modal redesign

jacbn:
Fixes a number of issues with components lifted from My Account, then implements them into a redesigned new academic year modal.

You can test this by changing the `else if (...)` [here](https://github.com/isaacphysics/isaac-react-app/blob/master/src/app/state/middleware/notificationManager.ts#L40) to just `else if (true)`. 

Confirmed Ada & physics are happy with the changes.

### DIFF
--- a/src/app/components/elements/inputs/UserContextAccountInput.tsx
+++ b/src/app/components/elements/inputs/UserContextAccountInput.tsx
@@ -12,16 +12,15 @@ import {
     isTutorOrAbove,
     siteSpecific,
     STAGE,
-    TEACHER_REQUEST_ROUTE,
     validateUserContexts
 } from "../../../services";
-import {Button, Col, CustomInput, FormGroup, Label, UncontrolledTooltip} from "reactstrap";
+import {Button, Col, FormGroup, Label, UncontrolledTooltip} from "reactstrap";
 import {UserContext} from "../../../../IsaacApiTypes";
 import {v4 as uuid_v4} from "uuid";
-import {Link} from "react-router-dom";
 import classNames from "classnames";
 import {Immutable} from "immer";
 import {StyledDropdown} from "./DropdownInput";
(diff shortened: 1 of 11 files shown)


src/app/components/elements/inputs/UserContextAccountInput.tsx
```diff
@@ -43,7 +42,7 @@ function UserContextRow({
     userContext, setUserContext, showNullStageOption, submissionAttempted, existingUserContexts, setBooleanNotation, setDisplaySettings,
     tutorOrAbove, showPlusOption, userContexts, setUserContexts, index, required
 }: UserContextRowProps) {
-    const onlyUCWithThisStage = existingUserContexts.filter(uc => uc.stage === userContext.stage).length === 1;
+    const onlyUCWithThisStage = existingUserContexts.length === 0 || existingUserContexts.filter(uc => uc.stage === userContext.stage).length === 1;
```
jacbn, line 45: This fixes an issue for users who have not previously set any context; `existingUserContexts` is empty but we want All Stages, All EBs to show. This necessitates All EBs being a valid option, so this bool needs to be true.

src/app/components/elements/modals/UserContextReconfirmationModal.tsx
```diff
@@ -95,25 +124,47 @@ const UserContextReconfimationModalBody = () => {
     return <Form onSubmit={formSubmission} className={"mb-2"}>
         <p>{modalText.intro}</p>
         <p>{modalText.connections}</p>
-        <div className="text-right text-muted required-before">
+        {isPhy && <div className="text-right text-muted required-before">
```
jacbn, line 127: Missed this in Ada My Account redesign; Ada prefer showing `(optional)` around optional fields, physics want `*` around required fields

skyepurchase (approved):

PR #1010, opened 2024-06-26
Allow for inline q details to be undefined on 1st render after soft-link-follow

mlt47:
On following a soft-link to an inline question, the new page's IsaacInlineRegion is rendered before the questions have been registered and elementToQuestionMap has been built. We have updated the type to allow for this and, importantly, added the value to the dependencies of the useEffect that registers the question parts so that the inline regions render when the data is available.


jsharkey13:
We're not 100% sure this fixes it, but given this is time-critical I am going to merge and continue to monitor errors after releasing it.

PR #1014, opened 2024-07-02
Ada core and advanced tagging

skyepurchase:
If a topic or concept is not on a users specification display the tag as "core" or "advanced" based on prior tagging. This prefers "core" over "advanced" and only shows "advanced" if "core" is not present.

If a user has not set their preferences every tag is displayed as "core" or "advanced". To accomodate this, if a user has not set their preferences , on Ada, they have a new "none" stage and examboard to distinguish from deliberately selected "all".


src/app/components/elements/list-groups/TopicSummaryLinks.tsx
```diff
@@ -46,17 +48,34 @@ export function TopicSummaryLinks({items, search}: {items: ContentSummaryDTO[];
             // Render remaining items
             .map((item, index) => {
```
mwtrew, line 49: There seems to be a bit of a mismatch between what's happening here and in Accordions - if I set my stage to `Ada` `Core`, Core topic summaries have an empty but still coloured stage label (see attached) and Core accordions have no label at all. I'm not sure which if either is desired.
Screenshot 2024-07-16 at 17 21 03

It might be nice to factor out the common code to reduce the likelihood of differences, but perhaps not now.

mwtrew (changes requested):

mwtrew (approved):

src/app/components/elements/inputs/UserContextAccountInput.tsx
```diff
@@ -53,7 +54,7 @@ function UserContextRow({
         const onlyOneAtThisStage = existingUserContexts.length === 0 || existingUserContexts.filter(uc => uc.stage === e.target.value).length === 1;
         const possibleExamBoards = getFilteredExamBoardOptions(
             {byStages: [stage || STAGE.ALL], byUserContexts: existingUserContexts, includeNullOptions: onlyOneAtThisStage
-            }) || [EXAM_BOARD.ALL];
+            }) || [EXAM_BOARD.ADA];
```
jsharkey13, line 57: The method `getFilteredExamBoardOptions` returns an array of `{label: string, value: EXAM_BOARD}` objects, but then we have this array that just contains `EXAM_BOARD.ADA` and is thus a completely different type. This isn't a new problem, but I do think we should fix it.

jsharkey13:
I merged the Bootstrap changes into this, and I think I resolved the messy conflict correctly. However, there are type warnings in my IDE, which I think are valid and I have commented about above.

PR #1021, opened 2024-07-12
Fix assignment progress types

jacbn:
Makes a distinction in the type of studentProgress depending on whether the user has authorised access (`AuthorisedAssignmentProgress`) or not (base class, `AssignmentProgressDTO`).


mlt47 (approved):
I'm happy with these changes. I've altered it to use a type guard which is a bit safer and cleaner than having to assert that the type is different for users where we have authorisation to see their data. However I've also changed it to assume anything returned from the API could be null (undefined), which is less neat but more correct.

PR #1023, opened 2024-07-25
Ada school year refresh - homepage updates

mwtrew:


src/app/components/site/cs/HomepageCS.tsx
```diff
@@ -120,16 +95,14 @@ export const HomepageCS = () => {
...
-                                Use the question finder to explore more than 1,000 self-marking questions, filtering by
-                                topic, concept and qualification. Written by experts and updated frequently, they
-                                address common misconceptions with tailored feedback.
+                                Explore our bank of over 1000 self-marking questions. Filter by topic, concept, and qualification.
                             </p>
-                            <p><b>Students</b>: review key topics and get instant feedback</p>
-                            <p><b>Teachers</b>: save time by creating self-marking quizzes</p>
+                            <p><b>For students</b>: Learn or revise a topic and receive instant feedback.</p>
+                            <p><b>For teachers</b>: Save time by creating self-marking quizzes for your class.</p>
                             <Button className={"mt-4"} tag={Link} to="/quizzes/new" color='primary'>
```
mlt47, line 104: I'm guessing this should go to the new question finder rather than the old one?
The copy maybe also indicates that for teachers we should instead go to the gameboard/quiz builder page.

src/app/components/site/cs/HomepageCS.tsx
```diff
@@ -68,43 +65,21 @@ export const HomepageCS = () => {
...
                         <Col xs={12} lg={6}>
-                            <h2 className={"font-size-1-75 mb-4"}>Why use Ada Computer Science?</h2>
+                            <h2 className={"font-size-1-75 mb-4"}>What we offer</h2>
                             <ul className={"font-size-1 font-size-md-1-25"}>
-                                <li>Made by the University of Cambridge and the Raspberry Pi Foundation</li>
-                                <li>Mapped to <strong><Link to={"/exam_specifications"}>computer science exam specifications</Link></strong>, including GCSE and A level</li>
-                                <li>A great way to save time when planning lessons and homework</li>
-                                <li>Perfect for learning a topic after class or preparing for exams</li>
-                                <li>Free forever — no hidden costs</li>
+                                <li><b>Free computer science resources:</b> Tailored for students aged 14 to 19</li>
```
mlt47, line 74: Do these items maybe need to end with a full stop and perhaps a page break after the colon?

mlt47 (approved):
I didn't have access to the RPF G Doc to compare but looks reasonable.
Made a couple of suggestions to consider.

src/app/components/site/cs/HomepageCS.tsx
mwtrew: > I'm guessing this should go to the new question finder rather than the old one?
The navbar link still goes to the old one, I suppose we should shift all links over at the same time. I can create a new card for that?

PR #1026, opened 2024-08-01
Question finder redesign

skyepurchase:


src/app/components/pages/QuestionFinder.tsx
```diff
@@ -269,23 +273,37 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
...
+        setFiltersApplied(false);
+        return searchDifficulties.length > 0
+            || searchTopics.length > 0
+            || searchExamBoards.length > 0
+            || searchStages.length > 0
+            || searchBooks.length > 0
+            || selections.some(tier => tier.length > 0)
+            || Object.entries(searchStatuses)
+                .filter(e => e[0] !== "revisionMode"
+                        && e[0] !== "hideCompleted") // Ignore revision mode as it isn't really a filter
```
mwtrew, line 286: Is there a reason not to take `hideCompleted` into account? I note that changing it doesn't re-enable the "Apply filters" button, which is what I would expect.

src/app/components/elements/panels/QuestionFinderFilterPanel.tsx
```diff
@@ -0,0 +1,438 @@
...
+                    <StyledCheckbox
+                        color="primary"
+                        checked={searchStatuses.hideCompleted}
+                        onChange={() => setSearchStatuses(s => {return {...s, hideCompleted: !s.hideCompleted};})}
+                        label={<div>
+                            <span>Hide complete</span>
+                        </div>}
+                    />
+                </div>
+                {/* TODO: implement new completeness filters
```
mwtrew, line 357: As a general rule it's best not to merge commented-out code, it has a habit of hanging around and confusing things if priorities change!

mwtrew (approved):
Looking great! 

One other thing I noticed is that the `CollapsibleList`s towards the bottom of Topics filters don't seem to open correctly and/or some of the items are cut off.

PR #1037, opened 2024-08-09
Warn users about inverse trig notation

axlewin:
If users enter e.g. sin^-1, provide instant feedback to correct to arcsin.


src/app/components/content/IsaacSymbolicQuestion.tsx
```diff
@@ -76,6 +76,19 @@ export const symbolicInputValidator = (input: string) => {
     if (/\.[0-9]/.test(input)) {
         errors.push('Please convert decimal numbers to fractions.');
     }
+    const invRegex = /((sin|cos|tan|sec|cosec|cot)h?(\^|\*\*)[({]?-1[)}]?)/;
+    if (invRegex.test(input)) {
+        const match = input.match(invRegex);
+        if (match != null) {
```
jsharkey13, line 82: We're doing two lots of the same work here if the regex matches. Rather than two nested if statements, we could instead just remove the outer if entirely. The code will still work fine!

src/app/components/content/IsaacSymbolicQuestion.tsx
```diff
@@ -76,6 +76,19 @@ export const symbolicInputValidator = (input: string) => {
...
+            const trigFunction = match[0].toString().replace(/[^a-zA-Z]+/g, '');
```
jsharkey13, line 83: The main regex could already do all this work. If we modify it so that the trig function name and the optional "h" are groups, we can make this much simpler!
```typescript
const invRegex = /(((sin|cos|tan|sec|cosec|cot)(h?))(\^|\*\*)[({]?-1[)}]?)/;
```
Then `match[2]` is the full name of the trig function with no extra stuff, and if `match[4] === 'h'` it's a hyperbolic one. Note we usually use `===` when comparing values to ensure no casting, too.

jsharkey13 (commented):
This code could probably still be simpler, though there's nothing actually wrong with it.

jsharkey13 (approved):

PR #1038, opened 2024-08-09
Show indicator by 'Work to do'

axlewin:
On Ada CS, show teachers a numerical indicator if they have work (assignments) to do.


axlewin:
Need to fix side effects caused on Isaac Physics.

PR #1044, opened 2024-08-12
Align checkbox text and login buttons

sjd210:
"Remember me" checkbox vertically aligned to be inline with text.
"Forgotten your password?" horizontally aligned to right when wrapping on small screen.

---
edit: Vertically aligns all checkboxes with their texts across both sites.


mwtrew (changes requested):
Sorry for taking a while to get around to reviewing this! The "Forgotten your password" change looks good. 

For the "Remember me" checkbox alignment, this seems like an issue with `form-check-input`s across the site (see also the [age radio buttons](https://adacomputerscience.org/register/student/age) for the Student signup flow). I think it's better to fix it in general rather than in this specific place. 

Removing `margin-top: 0` [here](https://github.com/isaacphysics/isaac-react-app/blob/3a5a54339c09c2abb384dbad2dea2bdfb2912ee8/src/scss/common/forms.scss#L241) seems to fix it generally and I'm not sure why we're doing that, perhaps you can ask Jacob if that's likely to cause any problems.

sjd210:
@mwtrew Looks like that works well for Ada, but significantly offsets the alignment on Isaac. I think probably the better solution is replacing each of these old `Input` checkboxes with new `StyledCheckbox`s which we've been a lot more careful with the consistent behavior of. 

I'll see what that looks like now.

sjd210:
I've set the "Remember me" checkbox as a `StyledCheckbox` as mentioned above, however radio-style inputs don't function well as a `StyledCheckbox` so I've left those as they are.

Otherwise, I've been investigating everywhere that we use checkboxes and it turns out that there's a lot of inconsistent behavior across checkboxes both between sites and even within just Isaac. 

Due to this, there's not much to be done in css to fix this, so I've just made various (mostly `siteSpecific`) fixes for all cases where text was misaligned from its box, and this seems to work perfectly well.

PR #1046, opened 2024-08-12
Physics student sign-up flow

jacbn:
Adds Physics changes alongside the Ada components for the new sign-up flow (no noticeable changes should be visible on Ada).


src/app/components/pages/RegistrationAgeCheck.tsx
```diff
@@ -12,23 +12,30 @@ import {
...
-import {history, SITE_TITLE} from "../../services";
+import {history, isPhy, SITE_TITLE, siteSpecific} from "../../services";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
 
+type AgePermission = "denied" | "additional_info" | "allowed";
+
 export const RegistrationAgeCheck = () => {
 
-    const [over13, setOver13] = useState<boolean | undefined>(undefined);
+    const [over13, setOver13] = useState<AgePermission | undefined>(undefined);
```
skyepurchase, line 22: Would be good to rename `over13` to something like `agePermission` or just `permission` to avoid potential confusion. This name change will obviously need to propagate.

skyepurchase (changes requested):

src/app/components/elements/inputs/SetPasswordInput.tsx
```diff
@@ -10,37 +10,56 @@ interface SetPasswordInputProps {
...
-                <strong>Password strength: </strong>
-                <span id="password-strength-feedback">
-                    {passwordFeedback.feedbackText}
+    return <div className={className}>
+        <FormGroup className="form-group">
+            <Label htmlFor={`${idPrefix}-password-set`} className={"fw-bold form-required"}>Password</Label>
+            <p className="d-block input-description">Your password must be at least {MINIMUM_PASSWORD_LENGTH} characters long.</p>
+            <TogglablePasswordInput
+                id={`${idPrefix}-password-set`} name="password" type="password"
+                required aria-describedby="invalidPassword"
```
mwtrew, line 31: Instead of the `required` prop we instead use our own `FormFeedback` everywhere for consistency, and including it breaks that - try submitting the form without filling in any fields and you'll see that `FormFeedback` no longer works.

mwtrew: (Sorry to jump in @skyepurchase!)

mwtrew:
Had a play around with this again and spotted a couple of minor issues:

1. You have to press tab twice to shift focus from Password to Confirm password, it must be focusing on something in between.
2. Invalid DoB (`You must be over 10 years old to create an account.`) doesn't prevent form submission as it's not a required field - maybe this is fine since the user has already confirmed they're old enough, but it's a bit weird for the message to show up.

mlt47 (approved):
I approve the changes since Matthew's last comment.

skyepurchase (approved):

PR #1047, opened 2024-08-13
Display assignments to do as default

mlt47:
This is a temporary improvement while we discuss designs to correctly emphasise the assignments referred to by the notification counters.


jacbn:
It's probably worth noting that you can't ever see the help text on Phy because a modal with that ID exists, but it's useful to have around in case that modal breaks or otherwise doesn't exist.

PR #1050, opened 2024-08-13
Cloze dropdown scrolling and greying out

sjd210:
Limits the size of cloze dropdown menus and adds scrollbar on vertical overflow.

Greys out previously selected options on non-replaceable cloze questions rather than removing them from the dropdown.

[Example for scrolling](https://isaacphysics.org/questions/hypothesis_meeples?stage=all)
[Example for non-replaceable questions](https://isaacphysics.org/questions/itsp_density_hw_q2?stage=all)

### DIFF
--- a/src/app/components/elements/markup/portals/InlineDropZones.tsx
+++ b/src/app/components/elements/markup/portals/InlineDropZones.tsx
@@ -59,7 +59,7 @@ function InlineDropRegion({id, index, emptyWidth, emptyHeight, rootElement}: {id
     const droppableId = CLOZE_DROP_ZONE_ID_PREFIX + `${index + 1}`;
     const dropdownItems = dropRegionContext?.allItems ?? [];
     const nonSelectedItems = dropRegionContext?.nonSelectedItems ?? [];
-    const nonSelectedItemIds = nonSelectedItems.map(item => item.id)
+    const nonSelectedItemIds = nonSelectedItems.map(item => item.id);
 
     useEffect(() => {
         // Register with the current cloze question on first render
--- a/src/app/components/elements/markup/portals/InlineDropZones.tsx
+++ b/src/app/components/elements/markup/portals/InlineDropZones.tsx
@@ -57,7 +57,9 @@ function InlineDropRegion({id, index, emptyWidth, emptyHeight, rootElement}: {id
     const deviceSize = useDeviceSize();
     const [isOpen, setIsOpen] = useState<boolean>(false);
     const droppableId = CLOZE_DROP_ZONE_ID_PREFIX + `${index + 1}`;
-    const dropdownItems = dropRegionContext?.nonSelectedItems ?? [];
+    const dropdownItems = dropRegionContext?.allItems ?? [];
+    const nonSelectedItems = dropRegionContext?.nonSelectedItems ?? [];
+    const nonSelectedItemIds = nonSelectedItems.map(item => item.id)
 
     useEffect(() => {
         // Register with the current cloze question on first render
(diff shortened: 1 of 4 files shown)


src/app/components/elements/markup/portals/InlineDropZones.tsx
```diff
@@ -122,14 +124,26 @@ function InlineDropRegion({id, index, emptyWidth, emptyHeight, rootElement}: {id
...
-                </DropdownItem>;
+                    >
+                        <Markup trusted-markup-encoding={"html"}>
+                            {item.value ?? ""}
+                        </Markup>
+                    </DropdownItem>;
+                }
+                else 
+                {
+                    return <DropdownItem key={i} style={{color:"darkgray", backgroundColor:"lightgray", pointerEvents:"none"}}>
```
skyepurchase, line 140: Unfortunately `'pointerEvents': "none"` means that the item acts invisible to clicks and so the user would click through the item onto the page behind potentially interacting with a component they did not intend.

`DropdownItem`s have a `disabled` property that should do what you want here. It would also be worth checking the `active` property as it may do the graying out.

src/app/components/elements/markup/portals/InlineDropZones.tsx
```diff
@@ -122,14 +124,26 @@ function InlineDropRegion({id, index, emptyWidth, emptyHeight, rootElement}: {id
                 <span className="d-inline-block"></span>
             </DropdownItem>
             {dropdownItems.map((item, i) => {
-                return <DropdownItem key={i}
+                if (nonSelectedItemIds.includes(item.id))
```
skyepurchase, line 127: After the changes I suggested for invalid items the two different `DropdownItem`s will be very similar. It would be better to conditionally set the `disabled` property based on `nonSelectedItemIds.includes(item.id)` similarly for adding the `invalid` classname using the `classNames` function.

src/app/components/elements/markup/portals/InlineDropZones.tsx
```diff
@@ -57,7 +57,9 @@ function InlineDropRegion({id, index, emptyWidth, emptyHeight, rootElement}: {id
     const deviceSize = useDeviceSize();
     const [isOpen, setIsOpen] = useState<boolean>(false);
     const droppableId = CLOZE_DROP_ZONE_ID_PREFIX + `${index + 1}`;
-    const dropdownItems = dropRegionContext?.nonSelectedItems ?? [];
+    const dropdownItems = dropRegionContext?.allItems ?? [];
+    const nonSelectedItems = dropRegionContext?.nonSelectedItems ?? [];
+    const nonSelectedItemIds = nonSelectedItems.map(item => item.id);
```
skyepurchase, line 62: This works perfectly fine and doesn't need to however with `line 127` you could use `dropRegionContext?.nonSelectedItems.every(nsi => nsi.id !== item.id)` which would be a bit neater with less variables

skyepurchase (changes requested):
Good solution! This is a not the simplest section of the code base and I hoped the task would be more straight forward than it was.

I have a couple of suggestions for improved code style and removing unintended behaviour.

src/app/components/elements/markup/portals/InlineDropZones.tsx
sjd210: Good note. I've decided that probably the easiest-to-read option here is to keep `nonSelectedItemIds` as its own variable while removing `nonSelecteditems` as it can already be directly accessed through `dropRegionContext`.

src/app/components/elements/markup/portals/InlineDropZones.tsx
sjd210: It looks like `disabled` handles both disabling clicks and graying out the items (`active` seems to be unrelated to this kind of behavior).

That said, on closer inspection it also appears that both `disabled` AND `'pointerEvents': "none"` interact poorly with the new dropdown scrolling functionality on touchscreen, so potentially the best solution is just to disable `onClick` functionality for selected options. I'm looking into this now.

sjd210: I've implemented this in my latest push. This does mean that the pointer on desktop will still highlight the invalid dropdown items as clickable, but this fixes touchscreen usability. 

Realistically, I think this makes no real difference. The dropdowns only exist for `xs` screens which should all be touchscreen.

skyepurchase (approved):
Annoying that bootstrap puts `pointer-events: none` on it's disabled class!!

PR #1051, opened 2024-08-13
Fix dropdown cloze questions embedded in katex

skyepurchase:
Overflow clipping and hiding has been removed to prevent the dropdown menu being clipped. I have checked a few cases but this may cause unintended behaviour that should be flagged.


src/scss/phy/isaac.scss
```diff
@@ -257,8 +257,6 @@ p small {
 .katex.katex {
     font-size: 1.1em;
     white-space: nowrap !important;
-    overflow-x: clip;
-    overflow-y: hidden;
```
sjd210, line 261: I've noticed an issue here with viewing long LaTeX examples on small screens i.e. mobile. If the block is larger than the width of the screen, the live site allows it to be scrolled - but that is not possible here.

e.g. [Matrices - Definition](https://isaacphysics.org/concepts/cm_matrices?stage=all). (The **m x n** matrix near the top)

This was already true with my `overflow-x: clip` change (oops). This edit reverts that, however removing `overflow-y: hidden` has the exact same issue. It looks like adding `overflow-x: scroll` to `katex-display` solves the issue.

sjd210 (changes requested):

sjd210 (approved):

sjd210:
Tested across a range of screen types/sizes in various LaTeX display contexts and I'm convinced this works thoroughly.

The only (very edge) case that I could find with a remotely related problem is for sufficiently long accordion section headings written in LaTeX on particularly small phone screens (e.g. iPhone 5/SE). - [Manipulating Algebraic Fractions](https://isaacphysics.org/questions/bio_maths_q8?stage=all). 

Given that this was the only example I could find with it, this was a pre-exisiting problem, is only on old phones, and the expression is written out in full directly below (AND that we almost certainly don't want scrolling on headers anyway), I think we can live with that and it's ready to merge.

PR #1054, opened 2024-08-14
Use single difficulties across the Ada site

skyepurchase:
Includes tidying up existing QF design code removing the check on whether all difficulties are the same. The content team for Ada do not want to set multiple difficulties that could be inconsistent.

### DIFF
--- a/src/app/components/elements/PageTitle.tsx
+++ b/src/app/components/elements/PageTitle.tsx
@@ -5,7 +5,9 @@ import {
     filterAudienceViewsByProperties,
     isAda,
     isPhy,
+    simpleDifficultyLabelMap,
     SITE_TITLE,
+    siteSpecific,
     STAGE,
     stageLabelMap,
     useUserViewingContext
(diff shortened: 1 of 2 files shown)


mwtrew (changes requested):
Looks good, could we also use this in the Quiz Builder and Manage assignments if straightforward enough?

skyepurchase:
Have adjusted the quiz builder to use a single difficulty. I added an additional change to prevent the exam-board tags from being repeated for each stage and rather have the exam-boards specific to each stage shown.

The manage assignments should still have multiple difficulties as the difficulties correspond to the different questions in the assignment. And assignment may have a question for GCSE and A level P1 and a question for A level P2 this should show:
GCSE (P1)
A level (P1, P2)
So it doesn't make sense trying to collapse these into a single difficulty. Maybe a discussion to just have Core and Advanced is needed?

mwtrew (approved):

PR #1056, opened 2024-08-15
Show warning for double inequalities

axlewin:
If a symbolic input contains more than one equals/inequality sign, show instant feedback that this is not allowed.

### DIFF
--- a/src/app/components/content/IsaacSymbolicQuestion.tsx
+++ b/src/app/components/content/IsaacSymbolicQuestion.tsx
@@ -76,7 +76,7 @@ export const symbolicInputValidator = (input: string) => {
     if (/\.[0-9]/.test(input)) {
         errors.push('Please convert decimal numbers to fractions.');
     }
-    if (/[<>=].*[<>=]/.test(input)) {
+    if (/[<>=].+[<>=]/.test(input)) {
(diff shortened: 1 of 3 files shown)


mwtrew (changes requested):
This will also erroneously match e.g. `a >= b`, since `.*` will match any character any number of times including none. Perhaps you could also add a couple of test cases for this to `IsaacNumericQuestion.test.tsx?`

src/app/components/content/IsaacSymbolicQuestion.tsx
```diff
@@ -76,6 +76,9 @@ export const symbolicInputValidator = (input: string) => {
     if (/\.[0-9]/.test(input)) {
         errors.push('Please convert decimal numbers to fractions.');
     }
+    if (/[<>=].+[<>=]/.test(input)) {
+        errors.push('We are not able to accept double equations or inequalities.');
```
jsharkey13, line 80: I think this message, which is at least shorter than the proposal could still be clearer:
> We are not able to accept double inequalities, and answers will never require them.

I doubt anyone is typing `a = sqrt(4) = 2` and I don't think "double equation" is a thing anyone would know as a name (we can revisit if anyone notices).

jsharkey13 (approved):

PR #1057, opened 2024-08-15
Remove ampersands from report URLs

axlewin:
When the report flag is clicked on a specific question part, ensure the section title passed to the report page's
URL does not contain ampersands.

### DIFF
--- a/src/app/components/elements/ReportAccordionButton.tsx
+++ b/src/app/components/elements/ReportAccordionButton.tsx
@@ -22,7 +22,7 @@ export const ReportAccordionButton = ({pageId, sectionId, sectionTitle, sectionI
             params +=`&page=${pageId}`
             // In absence of an ID, the section titles are likely most useful in locating the problem
             if (sectionTitle) {
-                params +=`&section=${sectionTitle}`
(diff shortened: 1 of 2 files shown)


src/app/components/elements/ReportAccordionButton.tsx
```diff
@@ -22,7 +22,7 @@ export const ReportAccordionButton = ({pageId, sectionId, sectionTitle, sectionI
             params +=`&page=${pageId}`
             // In absence of an ID, the section titles are likely most useful in locating the problem
             if (sectionTitle) {
-                params +=`&section=${sectionTitle}`
+                params +=`&section=${sectionTitle.replaceAll("&", "_")}`
```
jsharkey13, line 25: This just removes the ampersands, but it would be better to keep them but encode them correctly with [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) to address the ampersand issue and other related issues at the same time.

jsharkey13 (changes requested):

jsharkey13 (approved):

PR #1060, opened 2024-08-16
Question finder completeness filters

skyepurchase:
- Improve email regex
- Implement question filtering by completion status
- Add styling based on question completion status
- Fix difficulty icon width


src/app/services/reactRouterExtension.ts
```diff
@@ -18,8 +18,10 @@ function removeQuestionMark(str: string) {
     }
 }
 
+export type listParams = {[key: string]: string | string[] | undefined};
+export type singleParams = {[key: string]: string | undefined};
```
mlt47, line 22: Types are PascalCase by convention.

src/app/components/elements/svg/DifficultyIcons.tsx
```diff
@@ -46,7 +46,7 @@ export function DifficultyIcons({difficulty, blank, classnames} : {difficulty: D
     return <div className={classnames}>
         <svg
             role={"img"}
-            width={`${difficultyCategoryLevels.length * (difficultyIconWidth + 2 * difficultyIconXPadding) - difficultyIconXPadding}px`}
```
mlt47, line 49: This doesn't seem correct to me. I would have thought there should be 1 unit of padding to the left of each icon + 1 on the right? This doesn't seem like the correct formula.

src/app/components/pages/QuestionFinder.tsx
```diff
@@ -371,16 +420,24 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
...
+                    <CardBody className={classNames({"border-0": isPhy, "p-0": displayQuestions?.length, "m-0": isAda && displayQuestions?.length})}>
                         <ShowLoading until={displayQuestions} placeholder={loadingPlaceholder}>
                             {displayQuestions?.length
                                 ? <LinkToContentSummaryList items={displayQuestions} noCaret className="m-0" />
                                 : (!applyFiltersClicked && searchQuery === ""
                                     ? <em>Please select and apply filters</em>
-                                    : <em>No results match your criteria</em>)
+                                    : filteringByStatus
+                                        ? <em>Expecting results? Try narrowing down your filters</em>
+                                        : <em>No results match your criteria</em>)
```
mlt47, line 440: I get this response when searching with just the filter statuses on. It is not made clear to the user why we don't return anything in this case. I actually think we should also be returning results in this case.

src/app/components/pages/QuestionFinder.tsx
```diff
@@ -68,10 +80,30 @@ function processTagHierarchy(subjects: string[], fields: string[], topics: strin
...
 
+function getInitialQuestionStatuses(params: listParams): QuestionStatus {
+    const statuses = arrayFromPossibleCsv(params.statuses);
+    if (statuses.length < 1) {
+        // If no statuses set use default
+        return {
+            notAttempted: true,
+            complete: true,
+            incorrect: true,
+        };
```
mlt47, line 91: It might look neater for these to all be false, initially.

src/app/components/pages/QuestionFinder.tsx
```diff
@@ -227,14 +263,28 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
...
+    useEffect(() => {
+        if (applyFiltersClicked) {
+            setFilteringByStatus(
+                !(Object.values(searchStatuses).every(v => v)
+                || Object.values(searchStatuses).every(v => !v))
+            );
+        };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [applyFiltersClicked]);
+
```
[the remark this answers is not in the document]
mlt47: We can get rid of the useEffect and the useState if we just have `const filteringByStatus = !(Object.values(searchStatuses).every(v => v) || Object.values(searchStatuses).every(v => !v));`

mlt47 (changes requested):
The feature looks neat. There are a few comments to look at before merging.

src/app/components/pages/QuestionFinder.tsx
```diff
@@ -215,7 +245,13 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
             params.book = toSimpleCSV(searchBooks);
         }
         if (isPhy && excludeBooks) params.excludeBooks = "set";
-        if (searchStatuses.hideCompleted) params.hideCompleted = "set";
+        if (!(searchStatuses.notAttempted
+            && searchStatuses.complete
+            && searchStatuses.incorrect)
+        ) {
+            params.statuses = Object.entries(searchStatuses).filter(e => e[1])
+                .map(e => e[0] === "incorrect" ? "tryAgain" : e[0]).join(",");
```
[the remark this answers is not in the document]
skyepurchase: Ada wants a friendly presentation. It might be worth changing the status entry to "tryAgain".

src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
```diff
@@ -57,15 +57,30 @@ export const ContentSummaryListGroupItem = ({item, search, displayTopicTitle, no
...
-    const questionIconLabel = item.correct ? "Completed question icon" : "Question icon";
-    const questionIcon = siteSpecific(
-        item.correct ?
-            <svg className={iconClasses} aria-label={questionIconLabel}><use href={`/assets/phy/icons/tick-rp-hex.svg#icon`} xlinkHref={`/assets/phy/icons/tick-rp-hex.svg#icon`}/></svg> :
-            <svg className={iconClasses} aria-label={questionIconLabel}><use href={`/assets/phy/icons/question-hex.svg#icon`} xlinkHref={`/assets/phy/icons/question-hex.svg#icon`}/></svg>,
-        item.correct ?
-            <img src="/assets/common/icons/completed.svg" alt={questionIconLabel}/> :
-            <img src="/assets/common/icons/not-started.svg" alt={questionIconLabel}/>
-    );
+    let questionIconLabel, questionIcon;
```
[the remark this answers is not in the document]
skyepurchase: Need to be running the new backend as it has breaking changes that this requires

src/app/components/elements/svg/DifficultyIcons.tsx
skyepurchase: Will look into this in future. For now minor change to block the PR on.

mlt47: Agree. (It might be that there are two paddings between each icon in which case the original formula is correct but we might need some small amount of additional padding to account for borders or similar).

mlt47 (approved):
Looks good to me. I think that students will find the status search, combined with the rest of the page, very useful!

PR #1062, opened 2024-08-20
Close modals on page change

axlewin:
When the page is changed (e.g. using the browser back button), close the active modal. This fixes cases where a modal remains open and gets stuck on 'loading' after a page change.

This assumes there are no cases where a modal should stay open after a page change.

### DIFF
--- a/src/app/state/actions/index.tsx
+++ b/src/app/state/actions/index.tsx
@@ -745,6 +745,7 @@ export const redirectTo = (path: string) => {
 
 export const registerPageChange = (path: string) => {
     store.dispatch(routerPageChange(path));
+    store.dispatch(closeActiveModal());
 };
 
 export const handleServerError = () => {


mlt47 (changes requested):
Well done in finding that part of the code!

I looked through the codebase for where we open modals and there is unfortunately the required account information modal on Physics, which is displayed after initially registering with a Google account, shows a modal and is then immediately followed by a redirect (and page change) by the component. So that unfortunately breaks with this change.

I think it might be possible to solve this a different way, but it is maybe not quite as neat.
In the ActiveModal component we can use a [useEffect()](https://react.dev/reference/react/useEffect) (a place to perform actions outside of the react render loop) to register an event listener that closes the modal on a ["popstate"](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event ) event. We should remember to remove the event listener using useEffect's "cleanup" function. I'm not 100% convinced it will work but it is worth giving it a try. Let me know if you'd like help with it, we've not covered the useEffect hook yet.

mwtrew (approved):
That seems to work, nice!

PR #1063, opened 2024-08-20
Improve gameboard builder table

skyepurchase:
Use multiple rows for each question specification and merge the rows that do not change across specification (such as the title).

This reverts back to `findAudienceRecordsMatchingPartial` to convert the single (partial) audience view examboard into all the relevant examboards.


src/app/services/userViewingContext.ts
```diff
@@ -343,10 +343,21 @@ export function filterAudienceViewsByProperties(views: ViewingContext[], propert
...
+ */
+export function findAudienceRecordsMatchingPartial(
+    audience: AudienceContext[] | undefined,
+    partialViewingContext: Partial<ViewingContext>
+) {
     return audience?.filter((audienceRecord) => {
-        return Object.entries(partialViewingContext).every(([key, value]) => audienceRecord[key as keyof AudienceContext]?.[0] === value);
+        return Object.entries(partialViewingContext)
+            .every(([key, value]) =>
+                audienceRecord[key as keyof AudienceContext]?.[0] === value);
```
sjd210, line 360: Only checking position `0` leads to incorrectly not displaying exam boards when a valid exam board exists in a later position in the `audienceRecord` array (see picture):

image
(The non-displayed exam boards have AQA as their first element, but it is not a filtered exam board)

---

Changing the line to `audienceRecord[key as keyof AudienceContext]?.includes(value as never)); ` seems to fix this, although you may want to look for a better `TypeScript` solution than casting as never.

sjd210 (changes requested):

src/app/services/userViewingContext.ts
skyepurchase: Ah I thought this might happen. Rather than fix this problem I have updated the tagging so that it only shows the exam-boards in the viewing contexts based on filtering. This means no weird expansion from a single exam-board (I have removed the function as it is not used anywhere and its functionality is not required).

src/app/components/elements/GameboardBuilderRow.tsx
```diff
@@ -42,99 +43,89 @@ const GameboardBuilderRow = (
...
-                    {difficulty && <DifficultyIcons difficulty={difficulty} />}
-                </div>),
-                <div>
-                    {filteredAudienceViews[0]?.difficulty && <DifficultyIcons difficulty={filteredAudienceViews[0].difficulty} />}
-                </div>
-            )}
+            {view.stage && <span>{stageLabelMap[view.stage]}</span>}
         </td>
+        {(isPhy || i === 0) && <td rowSpan={siteSpecific(1, arr.length)} className={classNames(cellClasses, "w-15")}>
+            {/* Show each difficulty icon for Physics or just the first one for Ada */}
```
sjd210, line 120: This *would* be fine if we assumed that all Ada questions have a difficulty set on their first filtered option. Unfortunately, I have found at least a few questions where this isn't the case:

image

_See [here](https://github.com/isaacphysics/ada-content/blob/710bd180ecf12dc6402bbeb54e8ed1d737eb6409/content/computer_science/data_structures_and_algorithms/data_structures/questions/struct_64.json#L133) in the `ada-content` repo - the GCSE audience doesn't have a difficulty marked (which it definitely should, and a [recent change](https://github.com/isaacphysics/isaac-content-editor/commit/94a547e662447d776a38bef645beff60557738cd) to the content editor should fix this for all future questions - perhaps we should have a script to fix this for existing questions?)._

Regardless, I advise changing to the first non-null `difficulty` from the entire question rather than just the first `difficulty` from the filtered selection.

sjd210 (changes requested):

sjd210 (approved):

mlt47 (approved):

PR #1065, opened 2024-08-20
Fix Ada search formatting

axlewin:


src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
```diff
@@ -64,7 +64,7 @@ export const ContentSummaryListGroupItem = ({item, search, displayTopicTitle, no
             <svg className={iconClasses} aria-label={questionIconLabel}><use href={`/assets/phy/icons/question-hex.svg#icon`} xlinkHref={`/assets/phy/icons/question-hex.svg#icon`}/></svg>,
         item.correct ?
             <img src="/assets/common/icons/completed.svg" alt={questionIconLabel}/> :
-            <img src="/assets/common/icons/not-started.svg" alt={questionIconLabel}/>
+            <img src="/assets/common/icons/not-started.svg" height="34px" alt={questionIconLabel}/>
```
jacbn, line 67: `ContentSummaryListGroupItem`s are unfortunately used elsewhere (c.f. new question finder in particular) with other images that need this to be the height it is, so we can't change this here. I don't think it looks _too_ bad as is, but if you would rather have it you'd need to add this in the css somewhere under `#search-page {...} img`.

src/scss/cs/boards.scss
```diff
@@ -44,6 +44,7 @@
 
 .question-progress-icon {
   height: 88px;
+  min-width: 75px !important;
```
jacbn, line 47: Two things:
1. I'm not really sure why the height is fixed here. The rows change height (particularly on mobile) and fixing it will only break it when it's not its standard height. Indeed, it is broken on [the live site](https://adacomputerscience.org/search?examBoard=all&query=arch&stage=all&types=isaacConceptPage,isaacTopicSummaryPage,page,isaacQuestionPage) too... (scroll down to Von Neumann at `md` width -- note the border-right on this component stopping after 88px). Given that it's `display: inline-block` you can actually just remove the line setting the height from the css and it should work.

2. Setting the min-width is the right thing to do, but if we do it one layer down (inside `.inner-progress-icon` we can both keep all relevant props on one element (particularly now that we don't need the height in `.question-progress-icon`), and no longer need the `!important`.

src/app/components/elements/list-groups/ContentSummaryListGroupItem.tsx
jacbn: Also, just on a more general note -- if you're having to modify one component of a given type and not the others (as in this case, where you're adding a height to this one svg but not the ones nearby which do similar things) there's probably a cleaner way of doing it, e.g. through css via the parent component.

jacbn (changes requested):
Nice stuff, the search page is definitely a lot cleaner now! A couple of points before it can go in though.

mwtrew (changes requested):
Screenshot 2024-08-26 at 10 58 59

Looks like [Jacob's first point here](https://github.com/isaacphysics/isaac-react-app/pull/1065#discussion_r1726889738) is still an issue - getting rid of [this CSS](https://github.com/isaacphysics/isaac-react-app/blob/master/src/scss/cs/list-groups.scss#L4) (I'm not sure why that's there either) should do it.

jacbn (approved):

jsharkey13 (approved):

PR #1066, opened 2024-08-20
Fix several "My Account" inconsistencies

sjd210:
A few small bootstrap/css changes to make all inputs on "My Account" consistent across their respective sites.

The changes are pictured below for clarity


src/app/components/elements/inputs/SchoolInput.tsx
```diff
@@ -137,8 +144,8 @@ export const SchoolInput = ({userToUpdate, setUserToUpdate, submissionAttempted,
                 <FormFeedback>
```
jacbn, line 144: This `FormFeedback` _wasn't_ ever pinging when you made this -- it needs to be a sibling of an `<Input invalid/>` to appear. [One of the commits in the last release](https://github.com/isaacphysics/isaac-react-app/pull/1046/commits/60c0be0bba8467c19acd0b3eb311ad466560bb68#diff-9fef467f53bf3a4c015015ce4bc1f3f855cb68e059b1069c5188f94f5d8e8bf8R23) modified `StyledCheckbox` to cheat this somewhat, so the checkbox component acts as if it were a `<Input type="checkbox"/>` itself, including activating `FormFeedback`s if it's invalid. 

In this case, annoyingly, that means that after updating this branch the feedback _is_ now showing, but it's styled inside the (flex-row) parent div, so it looks a bit weird. I'd (1) make the parent flex-column, so the feedback will appear underneath the checkbox, and (2) replace the <Label/> with the `StyledCheckbox label={...}` prop, such that _that_ appears horizontally next to the checkbox. Something like

```tsx
label={<span>Not associated with a {siteSpecific("","UK ")}school</span>}
```

should work fine -- the styling is all done inside `StyledCheckbox` for consistency with others.

src/app/components/elements/panels/UserBetaFeatures.tsx
```diff
@@ -19,7 +19,7 @@ export const RevisionModeInput = ({displaySettings, setDisplaySettings}: Revisio
         onChange={e => {
             setDisplaySettings((oldDs) => ({...oldDs, HIDE_QUESTION_ATTEMPTS: e.target.checked}));
         }} 
-        label={<p>Hide previous question attempts</p>}
+        label={<p className="pt-0">Hide previous question attempts</p>}
```
jacbn, line 22: I'd rather all `<StyledCheckbox/>` labels had the same styling (none) unless there's a good reason -- this `pt-1` we're overriding here is honestly a curse at this point but I think its removal now applies to all physics checkboxes, so you can modify `StyledCheckbox.tsx` to remove it directly!

jacbn (commented):
This page looks _so_ much cleaner, nice stuff. There is one important point about `StyledCheckbox` that has changed since you branched off master to make this and so will need updating, and a couple of other things where I was getting a bit pedantic :P

src/app/components/elements/inputs/SchoolInput.tsx
```diff
@@ -108,6 +110,11 @@ export const SchoolInput = ({userToUpdate, setUserToUpdate, submissionAttempted,
                 inputId={`school-input-${randomNumber}`}
                 placeholder={"Type your school name"}
                 value={schoolValue}
+                components={{ 
+                    IndicatorSeparator: () => null, 
+                    DropdownIndicator: () => null, 
+                    ValueContainer: ((props: ValueContainerProps) => <components.ValueContainer {...props} className="form-select border-0" />) as () => React.JSX.Element,
+                    Control: ((props: ControlProps) => <components.Control {...props} className="form-control p-0" />) as () => React.JSX.Element}}
```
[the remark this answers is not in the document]
sjd210: Huh, interesting - I was aligning the input text to the `inputs` above on Isaac. It seems they're out of line with the `selects` at the bottom of the page (but not the DOB `select`, annoyingly). 

If we're being extra pedantic, I think all of these should be in line with each other - which seems to be any easy fix by removing `ps-3` from the offending `classNames`. I'll make sure this works well on both sites now.

src/app/components/elements/inputs/SchoolInput.tsx
```diff
@@ -119,8 +130,8 @@ export const SchoolInput = ({userToUpdate, setUserToUpdate, submissionAttempted,
             />
         </React.Fragment>}
 
-        {((userToUpdate.schoolOther == undefined && !(selectedSchoolObject && selectedSchoolObject.name)) || userToUpdate.schoolOther == NOT_APPLICABLE) && <div className="d-flex mt-2">
-            <RS.Input
+        {((userToUpdate.schoolOther == undefined && !(selectedSchoolObject && selectedSchoolObject.name)) || userToUpdate.schoolOther == NOT_APPLICABLE) && <div className="d-flex-column mt-2 align-content-center">
```
jacbn, line 133:
```suggestion
        {((userToUpdate.schoolOther == undefined && !(selectedSchoolObject && selectedSchoolObject.name)) || userToUpdate.schoolOther == NOT_APPLICABLE) && <div className="d-flex flex-column mt-2 align-content-center">
```
Amusingly, this works regardless because the default `d-block` is vertically aligned. Best not to commit valueless styles though!

jacbn (approved):
LGTM! One tiny edit for cleanup but non-blocking.

PR #1068, opened 2024-08-21
Inequality mobile editor buttons double pressing

jacbn:
Removes the `onTouchEnd` handler from inequality modal buttons. This was causing the `update()` method to fire twice on mobile.

Needs testing on iOS devices. A quick search suggests `onClick` will still fire, but worth checking.

### DIFF
--- a/src/app/components/elements/modals/inequality/InequalityModal.tsx
+++ b/src/app/components/elements/modals/inequality/InequalityModal.tsx
@@ -82,7 +82,6 @@ const InequalityMenuNumber = ({n, update}: {n: number, update: () => void}) => {
         tabIndex={0}
         data-item={JSON.stringify({ type: "Num", properties: { significand: n.toString() } })}
         onClick={update}
-        onTouchEnd={update}
         onKeyUp={update}
     >
         <VShape/><Markup encoding={"latex"}>{`$${n.toString()}$`}</Markup>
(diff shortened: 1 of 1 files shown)


sjd210:
Just given this a test on iOS and it doesn't seem to be working at all as expected. 

As far as I can tell, pretty much everything is working without error (single-pressing) on the live site on iOS for all draggable menu items - however it is erroneous (double-pressing) for non-draggable items, which is just `plus-minus`.

This change removes one press from all interactions, so draggable items don't have any `onClick`/`onTouchEnd`/etc. behaviour, but `plus-minus` does seem to be working fine now.

Also as a separate but related issue: On the iOS live site any time a dragging action ends, it is registered as a click - even if the item is dragged from the menu into the main Inequality environment.  

---

I'll attach videos for clarity (and I can lend you my phone/we can work on it together for testing if needed):

**Live Site**

https://github.com/user-attachments/assets/834171eb-2c04-4ddc-9943-714cac3601cf

**This Branch**

https://github.com/user-attachments/assets/a263e62f-c67a-4bcd-8941-aaec6957e95d

jacbn:
Ah well, worth a shot. I'll rework it to use an internal state as to whether the user has pressed down rather than relying on a mixture of these wonky methods; it will be more architecture-independent then.

jacbn:
Closing this PR as this was fixed in a mostly separate fix for inequality on iOS in general.

PR #1069, opened 2024-08-21
Phy signup preferences form improvements

jacbn:
Applies changes to the final page on signup:
- Makes all fields mandatory on physics
- Disable "set preferences" button if not complete ("skip" is always accessible)


mlt47 (approved):
This PR also stops student users from registering an interest in more than one "stage" which was the initially intended behaviour. For those students that used to have 2 or more stages selected we now display their selected stages but do not allow them to delete the additionally recorded stages. This is OK for now, I will create a Trello card so that we address that after this week's release.

PR #1071, opened 2024-08-21
Adds warning when setting same quiz to class multiple times

sjd210:
Previously on re-submitting an active test, a toast would warn that _"You cannot reassign a test until the due date has passed."_ 

This change makes it fail earlier (on the front-end), displaying red warning text below the "**...group(s)**" input and disabling the "**Set Test**" button.

---

I am making use of a top-level `let` as a workaround to stop polling the backend on every change to the Set Tests modal. Since the value only changes on successfully setting a test, it does not need a re-render so this should be fine - however I'm very open to suggestions for another solution.

### DIFF
--- a/src/app/components/elements/modals/QuizSettingModal.tsx
+++ b/src/app/components/elements/modals/QuizSettingModal.tsx
@@ -7,7 +7,8 @@ import {
     useAppDispatch,
     useAppSelector,
     useAssignQuizMutation,
-    useGetGroupsQuery
+    useGetGroupsQuery,
+    useGetQuizAssignmentsSetByMeQuery,
 } from "../../../state";
 import {assignMultipleQuiz, isDefined, Item, selectOnChange, TODAY} from "../../../services";
 import range from "lodash/range";
@@ -59,6 +60,8 @@ export function QuizSettingModal({quiz, dueDate: initialDueDate, scheduledStartD
     const [scheduledStartDate, setScheduledStartDate] = useState<Date | null>(initialScheduledStartDate ?? null);
     const [feedbackMode, setFeedbackMode] = useState<QuizFeedbackMode | null>(initialFeedbackMode ?? null);
 
+    const allQuizAssignments = useGetQuizAssignmentsSetByMeQuery().data;
+
     const yearRange = range(currentYear, currentYear + 5);
 
     function addValidated(what: ControlName) {
@@ -86,7 +89,7 @@ export function QuizSettingModal({quiz, dueDate: initialDueDate, scheduledStartD
         });
     }
 
-    const groupInvalid = validated.has('group') && selectedGroups.length === 0;
+    const groupInvalid = validated.has('group') && selectedGroups.length === 0 || selectedGroups.some(group => allQuizAssignments?.some(assignment => assignment.quizId === quiz.id && assignment.groupId === group.value));
     const dueDateInvalid = isDefined(dueDate) && ((scheduledStartDate ? scheduledStartDate.valueOf() > dueDate.valueOf() : false) || dueDate.valueOf() < Date.now());
     const scheduledStartDateInvalid = isDefined(scheduledStartDate) && scheduledStartDate.valueOf() < TODAY().valueOf();
     const feedbackModeInvalid = validated.has('feedbackMode') && feedbackMode === null;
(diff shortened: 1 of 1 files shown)


src/app/components/elements/modals/QuizSettingModal.tsx
```diff
@@ -82,11 +87,12 @@ export function QuizSettingModal({quiz, dueDate: initialDueDate, scheduledStartD
                 setDueDate(null);
                 setScheduledStartDate(null);
                 setFeedbackMode(null);
+                allQuizAssignments = undefined;
             }
         });
     }
 
-    const groupInvalid = validated.has('group') && selectedGroups.length === 0;
+    const groupInvalid = validated.has('group') && selectedGroups.length === 0 || selectedGroups.some(group => allQuizAssignments?.some(assignment => assignment.quizId === quiz.id && assignment.groupId === group.value));
```
mlt47, line 95: Unfortunately the `quiz/assigned` endpoint returns all assigned quizzes whether or not it is past the due date. We will then need to either run that logic here or alter the endpoint to optionally receive a query parameter to specify the filter.
Either way readability might be improved by pulling out a well named function or constant so that `groupInvalid = validated.has('group') && selectedGroups.length === 0 && !alreadyAssignedToAGroup`.

src/app/components/elements/modals/QuizSettingModal.tsx
```diff
@@ -46,6 +47,8 @@ interface QuizSettingModalProps {
     feedbackMode?: QuizFeedbackMode | null;
 }
 
+let allQuizAssignments: QuizAssignmentDTO[] | undefined;
```
mlt47, line 50: As we have this information (perhaps retrieved in a different way given other comments) it might be useful to alter `groupOptions` to add " (already assigned)" to the group label in the drop-down so that the user gets a warning even earlier. It is possible that this will look overcrowded in which case there might be an alternative solution that is nicer - I don't know if the component supports disabled options, for example.

src/app/components/elements/modals/QuizSettingModal.tsx
```diff
@@ -59,6 +62,8 @@ export function QuizSettingModal({quiz, dueDate: initialDueDate, scheduledStartD
     const [scheduledStartDate, setScheduledStartDate] = useState<Date | null>(initialScheduledStartDate ?? null);
     const [feedbackMode, setFeedbackMode] = useState<QuizFeedbackMode | null>(initialFeedbackMode ?? null);
 
+    allQuizAssignments = !allQuizAssignments ? useGetQuizAssignmentsSetByMeQuery().data : allQuizAssignments;
```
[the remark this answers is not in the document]
mlt47: The RTK Query hooks have a lot of defaults set so that it does the "correct thing" when we're retrieving data from the API. It manages the redux state/cache for us. If assignments set by me hasn't changed since the last render it will not re-request it. It instead waits for other operations to invalidate that cache updating this subscription.

In this case all you need is `const { data: allQuizAssignments } = useGetQuizAssignmentsSetByMeQuery();` and because the `assignQuiz(...)` mutation specifies to [invalidate](https://github.com/isaacphysics/isaac-react-app/blob/42c82f7b5f581379a61f56763843df2fc22207dc/src/app/state/slices/api/quizApi.ts#L175) the `QuizzesSetByMe` tag, it will automatically re-request from that endpoint and update anywhere that is subscribed using `useGetQuizAssignmentsSetByMeQuery()` after an assignment. That means everything should just work and keep up-to-date by using that RTK Query hook.

It is the subject of this Friday's tech talk so hopefully that will help too!

mlt47 (changes requested):
This works and the implementation is along the right lines.
However, I believe that the RTK Query hook is actually doing what you want from it by default, so we should be able to simplify the implementation by switch to just using it! :)

src/app/components/elements/modals/QuizSettingModal.tsx
sjd210: I think this is a good call, and also that it works best with items being disabled in the `StyledSelect`. 

I initially had some issues with disabled items still being focused by default (annoyingly React has been working on a solution to this [for a while](https://github.com/JedWatson/react-select/pull/4080)), but the change to `options` in `styles` solves that in the scope of this page at least, so I'm pretty happy with it all now.

mlt47 (approved):

PR #1073, opened 2024-08-23
Add logged event for Question Finder search

mwtrew:
Adds a new logged event for Question Finder search, for example:

`QUESTION_FINDER_SEARCH`| ... | `"{""events"": {""book"": [], ""stages"": [""gcse""], ""topics"": [], ""examBoards"": [""ocr""], ""startIndex"": 0, ""difficulties"": [], ""searchString"": ""e"", ""questionStatuses"": {""complete"": false, ""tryAgain"": false, ""notAttempted"": false}}}"`

### DIFF
--- a/src/app/components/pages/QuestionFinder.tsx
+++ b/src/app/components/pages/QuestionFinder.tsx
@@ -216,7 +216,7 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
 
             dispatch(logAction({
                 type: "SEARCH_QUESTIONS",
-                events: {searchString, topics, examBoards, book, stages, difficulties, startIndex}
+                events: {searchString, topics, book, stages, difficulties, examBoards, questionStatuses, startIndex}
             }));
         }, 250),
         [nothingToSearchFor]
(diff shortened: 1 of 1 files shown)


src/app/components/pages/QuestionFinder.tsx
```diff
@@ -222,7 +214,10 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
                 limit: SEARCH_RESULTS_PER_PAGE + 1 // request one more than we need to know if there are more results
             }));
 
-            logEvent(eventLog,"SEARCH_QUESTIONS", {searchString, topics, examBoards, book, stages, difficulties, startIndex});
+            dispatch(logAction({
+                type: "QUESTION_FINDER_SEARCH",
+                events: {searchString, topics, book, stages, difficulties, examBoards, questionStatuses, startIndex}
+            }));
```
mlt47, line 220: The filters can be at the same level as the type.
I would also think that it is useful for Physics to also record `subjects` and `fields`.

mlt47 (changes requested):
Small requested change.

PR #1075, opened 2024-08-23
Chemistry and nuclear text entry

skyepurchase:
- Implement text entry for Chemistry questions
- Implement nuclear text entry
- Improve feedback for users

### DIFF
--- a/src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
+++ b/src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
@@ -47,9 +47,13 @@ function isError(p: ParsingError | any[]): p is ParsingError {
 }
 
 export const symbolicInputValidator = (input: string) => {
-    const openBracketsCount = input.split('(').length - 1;
-    const closeBracketsCount = input.split(')').length - 1;
-    const regexStr = /[^ 0-9A-Za-z()[\]{}*+,-./<=>^_]+/;
+    const openRoundBracketsCount = input.split("(").length - 1;
+    const closeRoundBracketsCount = input.split(")").length - 1;
+    const openSquareBracketsCount = input.split("[").length - 1;
+    const closeSquareBracketsCount = input.split("]").length - 1;
+    const openCurlyBracketsCount = input.split("{").length - 1;
+    const closeCurlyBracketsCount = input.split("}").length - 1;
+    const regexStr = /[^ 0-9A-Za-z()[\]{}*+,-./<=>^_\\]+/;
     const badCharacters = new RegExp(regexStr);
     const errors = [];
     if (badCharacters.test(input)) {
@@ -64,8 +68,11 @@ export const symbolicInputValidator = (input: string) => {
         }
         errors.push('Some of the characters you are using are not allowed: ' + usedBadChars.join(" "));
     }
-    if (openBracketsCount !== closeBracketsCount) {
-        errors.push('You are missing some ' + (closeBracketsCount > openBracketsCount ? 'opening' : 'closing') + ' brackets.');
+    if (openRoundBracketsCount !== closeRoundBracketsCount
+       || openSquareBracketsCount !== closeSquareBracketsCount
+       || openCurlyBracketsCount !== closeCurlyBracketsCount) {
+        // Rather than a long message about which brackets need closing
+        errors.push('You are missing some brackets.');
     }
     if (/\.[0-9]/.test(input)) {
         errors.push('Please convert decimal numbers to fractions.');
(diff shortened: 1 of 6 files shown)


sjd210 (changes requested):
I really liked seeing this all work together. There are just a few small front-end issues I think need solving here, but the grammar itself works perfectly!

src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
```diff
@@ -64,7 +228,35 @@ const IsaacSymbolicChemistryQuestion = ({doc, questionId, readonly}: IsaacQuesti
...
+                            {"NaCl(aq) -> Na^{+}(aq) +  Cl^{-}(aq)"}<br />
+                            {"^{238}_{92}U"}<br />
+                            <br />
+                            As you type, the box above will preview the result.
+                        </UncontrolledTooltip>
+                    </>
+                </InputGroup>
+                <QuestionInputValidation userInput={textInput} validator={symbolicInputValidator} />
+                {symbolList && <div className="eqn-editor-symbols">
+                    The following symbols may be useful: <pre>{symbolList}</pre>
```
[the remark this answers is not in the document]
skyepurchase: There's two ways about this:
1. Change it in Inequality to accept `\alphaparticle` rather than `alpha`
2. Change it in the front-end to convert `alpha` to `\alphaparticle`

The second is easiest, though ideally both would be in sync. There is also a case for allowing users to input `alpha`, `beta`, `n`, etc. But that will need to be done in the grammar.

src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
```diff
@@ -64,7 +235,43 @@ const IsaacSymbolicChemistryQuestion = ({doc, questionId, readonly}: IsaacQuesti
...
+                <InputGroup className="my-2 separate-input-group">
+                    <Input type="text" onChange={updateEquation} value={textInput}
+                              placeholder="Type your formula here"/>
+                    <>
+                        {siteSpecific(
+                            <Button type="button" className="eqn-editor-help" id={helpTooltipId} tag="a" href="/solving_problems#symbolic_text">?</Button>,
+                            <span id={helpTooltipId} className="icon-help-q my-auto"/>
+                        )}
+                        {doc.isNuclear
+                                ? <UncontrolledTooltip placement="top" autohide={false} target={helpTooltipId}>
```
sjd210, line 249: Nice examples! However I think that with the length of Nuclear answers, the overflow of an example over multiple lines makes it quite unclear when one ends and the next starts. I've got a couple different solutions in mind:

Either we can change each `<br />` to `<br className="mb-2" />` 

OR for the same effect with a more forward-thinking approach add a new className like `spacedTooltip` to `UncontrolledTooltip` 

```suggestion
                                ? <UncontrolledTooltip className="spacedTooltip" placement="top" autohide={false} target={helpTooltipId}>
```
and then in a css file (`questions.scss` should do, at the top-level due to how tooltips are handled): 
``` 
.spacedTooltip br {
  margin-bottom: 0.5rem;
}
```

sjd210: The Chemistry tooltip below, and probably even the Symbolic Maths [tooltip](https://github.com/isaacphysics/isaac-react-app/blob/98cbab5920b7976f7021c3470121d9cbc716ce73/src/app/components/content/IsaacSymbolicQuestion.tsx#L243C1-L243C25) could probably also do with this same `spacedTooltip` treatment, as they also have examples going over multiple lines - although just 2 lines instead of the 3 here.

(I'd also then remove the extra floating `<br />`s either side of the examples - otherwise it's taking up too much space)

skyepurchase: I think the maths tooltip is okay for now. If people get annoyed at the inconsistency then we can change it.

When this is merged I will also update the symbol help accordion `/solving_problems#symbolic_text` so that there is a more in depth symbol list for students.

src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
```diff
@@ -64,7 +235,41 @@ const IsaacSymbolicChemistryQuestion = ({doc, questionId, readonly}: IsaacQuesti
...
+                                ? <UncontrolledTooltip className="spaced-tooltip" placement="top" autohide={false} target={helpTooltipId}>
+                                    Here are some examples of expressions you can type:<br />
+                                    {"^{238}_{92}U -> ^{4}_{2}\\alphaparticle + _{90}^{234}Th"}<br />
+                                    {"^{0}_{-1}e"}<br />
+                                    {"\\gammaray"}<br />
+                                    As you type, the box above will preview the result.
+                                </UncontrolledTooltip>
+                                : <UncontrolledTooltip className="spaced-tooltip" placement="top" autohide={false} target={helpTooltipId}>
+                                    Here are some examples of expressions you can type:<br />
+                                    <br />
```
sjd210, line 258: As one final nit, I'd also remove this other hanging `<br />` as you did with the bottom ones (You don't have one on the Nuclear tooltip so there's no problem there). 

```suggestion

```
#  
After that, please do merge both PRs. I think we've had enough back-and-forth.

sjd210 (approved):

PR #1076, opened 2024-08-27
Move user context tooltip

axlewin:
The 'unusual viewing context' tooltip icon shows in different places depending on account type. This should move it to always be next to the stage (and exam board) selector.


jacbn (approved):
Looks good on both sites. Ada will probably want the hovertip removed soon but given it was already there this isn't making anything worse for them.

PR #1077, opened 2024-08-27
Remove connections step from phy registration

mlt47:


src/app/state/actions/index.tsx
```diff
@@ -215,8 +216,8 @@ export const registerNewUser = (
         if (isTeacherOrAbove(newUser)) {
             // Redirect to email verification page
             history.push('/verifyemail');
-        } else {
-            history.push('/register/connect')
+        } else  {
```
mlt47, line 219:
```suggestion
        } else {
```

jacbn (approved):

PR #1079, opened 2024-08-28
Fix connections and groups placeholder text

jsharkey13:
These all said "Search teachers" beforehand, even when shown on the students or groups connections lists.


jsharkey13:
Jacob did this simultaneously in https://github.com/isaacphysics/isaac-react-app/pull/1080 🤦

PR #1081, opened 2024-08-28
Hide deprecated tests

axlewin:
Hide tests with the 'deprecated' flag from all users on the 'set tests' page. Previously set tests which are now deprecated should still show up under 'manage tests' (and can be re-set from here, which should perhaps be disabled).

This also requires a change in the editor to show the 'deprecated' flag for tests.


mlt47 (approved):
This looks good to me, but we'll hold off merging until after the release in a few hours.

PR #1082, opened 2024-08-28
Switch email edit notice from alert to modal

sjd210:
Adds an `EmailChangeModal` to tell users upon changing email to verify their new address by following the verification link (previously a browser alert)

The logic of the rest of the `updateCurrentUser(...)` function is moved into two separate functions called depending on modal output (or `continueSettingsUpdate()` is always called if no modal is necessary).

### DIFF
--- a/src/app/state/actions/index.tsx
+++ b/src/app/state/actions/index.tsx
@@ -54,6 +54,7 @@ import {
     store,
 } from "../index";
 import {Immutable} from "immer";
+import { Button } from "reactstrap";
 
 // Utility functions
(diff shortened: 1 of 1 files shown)


jacbn (commented):
Design-wise, this is great, just two tiny points about styles. I wonder though if the text itself needs an update -- if the main issue is people being confused about their old email still appearing after a reload, perhaps we should rephrase it to highlight that their _new_ email address _won't_ work until they verify. Worth asking RPF for their opinion on the copy.

jacbn (approved):

PR #1085, opened 2024-08-29
Various spacing issues

jacbn:
Fixes a number of spacing and text overlap issues found during regression testing:

- Teacher connections panel rows must be fixed height (by virtue of React Window) so each line of text is now limited to that one line and will cut off with an ellipsis if longer;
- For the groups panel only, the full, non-cut-off group title will show on hover (and is accessible through screenreader, as are all cut-off strings);
- Fixes Ada Online Courses overlap on Firefox;
- Aligns the Phy board hexagon on Set Tests;
- Extends the max-height of toasts.

### DIFF
--- a/src/scss/common/header.scss
+++ b/src/scss/common/header.scss
@@ -34,7 +34,6 @@
 
   .fade {
     transition: opacity 0.4s linear, max-height 0.3s ease-out, margin-bottom 0.3s linear, border-width 0.3s linear;
-    max-height: 300px;
   }
   .toast:not(.show) {
     max-height: 0;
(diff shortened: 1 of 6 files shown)


src/scss/common/header.scss
```diff
@@ -34,7 +34,7 @@
 
   .fade {
     transition: opacity 0.4s linear, max-height 0.3s ease-out, margin-bottom 0.3s linear, border-width 0.3s linear;
-    max-height: 200px;
+    max-height: 300px;
```
jacbn, line 37: I'm not sure if we need this line at all -- if we make a yet longer toast it will still break. Is there a reason to limit toast size at all?

skyepurchase: I don't think there is a reason to limit the height. If something is too long we can edit the toast message or how it is shown to the user rather than having a problematic styling

skyepurchase (approved):
This is a good use of tooltips. Enough information is presented without the tooltip, and hard of seeing users do not interact with it!

PR #1089, opened 2024-09-03
Superseded / nofilter info bubbles

jacbn:
Adds "superseded" and "nofilter" info bubbles to questions when displaying on the question finder, site search and gameboard builder. 

"Superseded" questions can be seen by teachers in both the QF and site search, and will appear in the gameboard builder on questions they have previously set that have since been superseded. The "nofilter" tag can by design only be seen by staff, as these questions are not otherwise displayed.

### DIFF
--- a/src/app/components/elements/GameboardBuilderRow.tsx
+++ b/src/app/components/elements/GameboardBuilderRow.tsx
@@ -11,7 +11,8 @@ import {
     siteSpecific,
     stageLabelMap,
     TAG_ID,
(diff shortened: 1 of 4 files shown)


skyepurchase (approved):
Looks good and functions well. 

Only point is that it is hard to tell that the "superseded" tag is a button without hovering over it but as there are other mechanisms to reach the same information not important. Would be something to consider going forward if we use more tags.

PR #1099, opened 2024-09-05
React 18 Update

skyepurchase:
Upgrade react and related packages.

(edit: `react-leaflet` has been upgraded as well)


package.json
```diff
@@ -86,83 +86,83 @@
...
     "bootstrap": "^5.3.3",
     "bootstrap-block-grid": "^1.1.7",
     "clean-webpack-plugin": "^4.0.0",
     "copy-webpack-plugin": "^10.2.4",
-    "css-loader": "^6.10.0",
+    "css-loader": "^6.11.0",
     "cypress": "13.6.4",
     "dotenv": "^10.0.0",
-    "eslint": "^9.8.0",
+    "eslint": "^8.0.0-0",
```
mwtrew, line 138: Is there a reason to downgrade?

mwtrew (approved):
Looks good, but apparently the upgrade won't do much unless we also [move to the new rendering API](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis). Did you give that a try?

package.json
skyepurchase: `eslint-pluging-jsx-a11y` and `eslint-plugin-react-hooks` have a dependency requirement of `8.0.0-0` but I didn't find an upgrade for either of them to use the upgraded `eslint` package

PR #1101, opened 2024-09-05
Use cloze dropdowns on touchscreens

axlewin:
Use dropdowns instead of drop zones for cloze questions on any touchscreen device. Keep the size breakpoint for small screens of any type.

Using maxTouchPoints for this seems to work, but there's probably a neater way to actually detect the device type.

### DIFF
--- a/src/app/components/elements/markup/portals/InlineDropZones.tsx
+++ b/src/app/components/elements/markup/portals/InlineDropZones.tsx
@@ -125,9 +125,9 @@ function InlineDropRegion({id, index, emptyWidth, emptyHeight, rootElement}: {id
             </DropdownItem>
             {dropdownItems.map((item, i) => {
                 return <DropdownItem key={i}
-                className={!nonSelectedItemIds.includes(item.id) ? "invalid" : ""}
-                data-unit={item || 'None'}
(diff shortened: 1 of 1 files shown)


src/app/components/elements/markup/portals/InlineDropZones.tsx
```diff
@@ -139,7 +139,7 @@ function InlineDropRegion({id, index, emptyWidth, emptyHeight, rootElement}: {id
 
     if (dropRegionContext && droppableTarget) {
         return ReactDOM.createPortal(
-            deviceSize === "xs" ? dropdownZone : draggableDropZone,
+            ((navigator.maxTouchPoints) || (deviceSize === "xs")) ? dropdownZone : draggableDropZone,
```
jacbn, line 142: We actually have a helper method for this! It uses `maxTouchPoints` in the same way, but has a second check to cover a few more bases.
```suggestion
            (isTouchDevice() || (deviceSize === "xs")) ? dropdownZone : draggableDropZone,
```
(you'll need to import it in the file rather than just accepting this suggestion)

jacbn (commented):
Works well! Not urgent but we have a helper method for this (news to me too!) so we should probably stick to that for maintainability.

PR #1103, opened 2024-09-05
Deselect QF filters for multiple contexts

axlewin:
On Ada, if a user has more than one context set on their account preferences, don't set any question finder filters by default.

Users with a single context set should have the filters set as before.


skyepurchase (changes requested):
This unfortunately changes the viewing context for the entire site. To see the issue compare http://localhost:8003/questions/text_19 (with GCSE, AQA and Core, Ada) to https://staging.adacomputerscience.org/questions/text_19 (with the same setup). With these changes the user doesn't get a warning saying this content is not for them.

I think the place to deselect the QF filters is on the `QuestionFinder.tsx` page component. I believe only [this useEffect](https://github.com/isaacphysics/isaac-react-app/blob/master/src/app/components/pages/QuestionFinder.tsx#L117) needs to be changed.

skyepurchase (approved):

PR #1105, opened 2024-09-06
Fix padding on cloze items with latex within them

skyepurchase:


src/scss/common/questions.scss
jacbn, line 1: I am fairly against the changes here as they affect all latex on the site, so there's now extra line spacing around paragraphs with latex, unit dropdowns, accordion headers, etc. All to say I think this might be too wide-reaching for what we need. It looks fine to me with just the changes in the other two files (which only affect cloze questions), so I think we should just keep those.

jacbn (commented):
Cloze changes work great, but see comment about latex on the site in general.

PR #1112, opened 2024-09-09
Add fractional coordinate warning

axlewin:
For coordinate questions, give instant feedback if a fraction (or other invalid numerical expression) is input.

### DIFF
--- a/src/app/components/content/IsaacCoordinateQuestion.tsx
+++ b/src/app/components/content/IsaacCoordinateQuestion.tsx
@@ -5,6 +5,7 @@ import {Button, Input} from "reactstrap";
 import {isDefined, useCurrentQuestionAttempt} from "../../services";
 import {IsaacQuestionProps} from "../../../IsaacAppTypes";
 import {Immutable} from "immer";
+import QuestionInputValidation from "../elements/inputs/QuestionInputValidation";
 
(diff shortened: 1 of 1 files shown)


src/app/components/content/IsaacCoordinateQuestion.tsx
```diff
@@ -16,6 +17,16 @@ interface CoordinateInputProps {
     readonly?: boolean;
     remove?: () => void;
 }
+
+export const coordinateInputValidator = (input: string) => {
+    const errors = [];
+    const operatorExpression = new RegExp(".*[0-9][+/÷-]\\.?[0-9]+");
```
skyepurchase, line 23:
```suggestion
    const operatorExpression = new RegExp("[0-9]\s*[+\/÷\-x×]\s*[0-9]");
```
I think this regex can be simplified to this as:
* the leading consumer is not required
* multiplication may be a desired thing to warn about as well
* I wasn't sure why `\` was important to capture as well as the optional wild card
* We only need at least one number after the symbol

This additionally allows spaces around the binary operator symbol.

You can go further and use `/[0-9]\s*[+\/÷\-x×]\s*[0-9]/` instead of `new RegExp(...)`

skyepurchase (changes requested):
Looks/works really well I think the regex is just a bit too complex

skyepurchase (approved):

PR #1122, opened 2024-09-11
Attempt to add CodeCov bundle size analysis

jsharkey13:
I might have gotten these settings wrong, but I need to open a PR to find out!


jsharkey13:
Giving up on this for now; uploads are too unreliable (they always succeed, but only one of ~15 commits actually worked and showed up on the website or triggered the comment). Will revisit at some point to see if it has matured!

PR #1124, opened 2024-09-12
Ensure inline entries are rendered on soft link follow

jacbn:
Adds an `initialised` flag to the inline context to account for portals rendering order differing on soft links.

Soft and hard link redirects having different rendering behaviour is the real cause and this only patches over this, so we should perhaps consider an alternative portals approach in the future.


skyepurchase (approved):
Hopefully Inlines can be solved soon

PR #1128, opened 2024-09-13
Enable Plausible for assignment redirects

sjd210:
Correctly counts a separate Plausible pageview for  `/assignment/[gameboard_id]` and  `/gameboards?...`

---

The `/assignment/[gameboard_id]` link is a redirect to the `/gameboards?...` page. 

`TrackedRoute` was only tracking the finally reached component, therefore both links `TrackedRoute`s were activating for `/gameboards?...` and so Plausible was miscounting. The missing `/assignment` pageview can be recovered by tracking route during the redirect itself.

The additional `/gameboards?...` count has to be fixed separately while keeping the wrapper logic of `TrackedRoute`, hence the addition of a `disableTracking` prop.

---

This will have to be checked for its interaction with Plausible itself once pushed, but seems good on localhost.


sjd210:
@mlt47 This makes sense. I had tried to do something similar but without the `useLocation` hook, so I'm glad there was a way to make this work more simply. 

I've added your code directly and tested that everything seems to work with it, so this should be good unless there's a strange interaction with Plausible once pushed in.

PR #1129, opened 2024-09-16
Hotfix/bluesky icon

axlewin:
Add Bluesky link to footer.


src/app/components/elements/list-groups/SocialLinks.tsx
```diff
@@ -11,33 +11,36 @@ const getSocialIcon = (name: string) => {
 };
 
 export const SocialLinksRow = () => {
+    const linksByColumn = [Object.entries(SOCIAL_LINKS).slice(0, 2), Object.entries(SOCIAL_LINKS).slice(2, 4)];
```
jsharkey13, line 14: This hard-codes the fact that there are only 4 icons. Maybe it would be better to chunk them into pairs dynamically?
We do have lodash used elsewhere, and it has [a convenient chunk function](https://lodash.com/docs#chunk). Just make sure to use the `import chunk from 'lodash/chunk';` form of import, else you'll bundle the entirety of lodash unnecessarily.

jsharkey13 (approved):

PR #1130, opened 2024-09-17
Add new maths book

axlewin:
Add Pre-University Mathematics for Sciences (2nd edition) to various lists.

### DIFF
--- a/src/app/components/elements/modals/QuestionSearchModal.tsx
+++ b/src/app/components/elements/modals/QuestionSearchModal.tsx
@@ -187,8 +187,8 @@ export const QuestionSearchModal = (
                         {value: "physics_skills_19", label: "A Level Physics (3rd Edition)"},
                         {value: "physics_linking_concepts", label: "Linking Concepts in Pre-Uni Physics"},
                         {value: "maths_book_gcse", label: "GCSE Maths"},
-                        {value: "maths_book", label: "Pre-Uni Maths (1st edition)"},
                         {value: "maths_book_2e", label: "Pre-Uni Maths (2nd edition)"},
+                        {value: "maths_book", label: "Pre-Uni Maths (1st edition)"},
                         {value: "chemistry_16", label: "A-Level Physical Chemistry"}
                     ]}
                 />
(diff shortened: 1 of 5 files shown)


src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -187,7 +187,8 @@ export const QuestionSearchModal = (
                         {value: "physics_skills_19", label: "A Level Physics (3rd Edition)"},
                         {value: "physics_linking_concepts", label: "Linking Concepts in Pre-Uni Physics"},
                         {value: "maths_book_gcse", label: "GCSE Maths"},
-                        {value: "maths_book", label: "Pre-Uni Maths"},
+                        {value: "maths_book", label: "Pre-Uni Maths (1st edition)"},
+                        {value: "maths_book_2e", label: "Pre-Uni Maths (2nd edition)"},
```
mlt47, line 191: It doesn't look like the questions have been tagged with `maths_book_2e` in the editor yet, so we shouldn't merge this change before that has been done.

src/app/components/elements/views/ProgressBar.tsx
```diff
@@ -18,6 +18,7 @@ const BAR_COLOURS = new Map([
     ["physics_linking_concepts", "bar-alevel"],
     ["maths_book_gcse", "bar-maths"],
     ["maths_book", "bar-maths"],
+    ["maths_book_2e", "bar-maths"],
```
mlt47, line 21: Although this colour has been defined, there is no entry on the My Progress page for this 2nd Edition book yet.

mlt47 (changes requested):
I wonder if it might make sense to order these 2nd edition before the 1st in any list as that should be the preferred book going forward?

The maths book card from `/alevel` and some data on `/pages/order_books` need to be updated, so maybe we discuss this with the content team tomorrow so that we correctly coordinate the release of these changes.

Otherwise looks good!

mlt47 (approved):
`/alevel` > **Maths for Sciences Book** goes to `/books/pre_uni_maths` which is the 1st edition. We will let the content team know to change the `pre_uni_maths_intro` page fragment to clarify that there are two versions of the book to select from.

PR #1133, opened 2024-09-19
Remove q param syncing code for viewing context

mlt47:
The code change is straightforward. The implications for teachers sharing links to students need to be agreed upon by the content team before we commit to this.


mlt47:
We have implemented this in another PR which has now been merged.

PR #1137, opened 2024-09-20
Sanitise hexagon board IDs

axlewin:
Fixes a small bug where invalid hexagon IDs could be passed to HexagonGroupsButton. This wasn't affecting the live site, but could cause an error on staging.


jsharkey13 (approved):
I wonder if maybe this would be safer if it _replaced_ the invalid characters rather than removed them? (Because DOM element IDs should be unique and if _most_ of the gameboard ID is forbidden, the result may not be unique).
However, only staff can make bad gameboard IDs, and I guess replacing with e.g. `_` is no more or less likely to result in duplicated DOM IDs anyway!

PR #1138, opened 2024-09-20
Remove transparent background from wildcards

sjd210:
I believe we used to rely on the `ListGroup` background to show the white colour here but a change at some point has removed this. Now using the same base classes as a question.


jacbn (approved):
I'm not sure Ada have wildcards (especially considering the contents of the `showWildcard` function...), so the `siteSpecific` might be unnecessary, but seems useful if they do decide to them.

PR #1145, opened 2024-09-25
Fix About Us alignment and similar issues

sjd210:
Best viewed on the new beta version of the Isaac About Us page.

- Aligns all email icons in the same row to the same height.
- Re-crops some images (e.g. Nicki) from the Project Team so all circles are the same size.
- Allows small screens to have a two-profile-wide view rather than only one or three for better viewability.
- Ports co-founders to` isaacFeaturedProfile`s to allow for consistency in styling and adjustments based on view-width.

The new page will replace the old one in the content editor once this PR is merged.

(Since this is *sort of* a content page, I've had to use some funky CSS here to make adjustments. The changes can't be made in the editor itself without making it unusable for the next edit.)


src/scss/phy/featured-profile.scss
```diff
@@ -3,3 +3,52 @@
...
+div:has(> .featured-profile:first-child:nth-last-child(2)) {
+  @extend .flex-row;
+  @extend .row-cols-1;
+  @extend .row-cols-sm-2;
+
+  .featured-profile {
+    @extend .isaac-tab;
+
+    width: 47.5%;
+    @media (max-width: 575px) {
```
jacbn, line 30: In case we ever change the breakpoint values (defined in `isaac.scss`), we should use the breakpoint functions:
```suggestion
    @include media-breakpoint-down(sm) {
```

src/scss/phy/featured-profile.scss
```diff
@@ -3,3 +3,52 @@
...
+    .profile-image {
+      padding-bottom: 1rem;
+      width: 95%;
+    }
+  }
+
+  .featured-profile:first-child {
+    margin-right: 5%;
+  }
+}
```
jacbn, line 54: This fixes a small padding issue left over from changing the column width:
```suggestion
}

@include media-breakpoint-down(lg) {
  div.pad:has(> div > .featured-profile:first-child:nth-last-child(2)) {
    padding-left: 0;
  }
}

```

jacbn (approved):
Looks _so_ much better, great stuff! A couple of tiny points.

src/scss/phy/featured-profile.scss
sjd210: Personally I don't really see a problem with the padding as is, but this is simple enough to add so I'll just trust you 
¯\\_(ツ)_/¯

PR #1150, opened 2024-09-25
MSW Upgrade

jacbn:
Upgrades from MSW 0.49.3 to 2.4.9.

Major changes required going forwards can be found [here](https://mswjs.io/docs/migrations/1.x-to-2.x). 

Oddities from the upgrade:
- The jest config/new polyfills file act as workarounds for around a variety of issues [[1]](https://mswjs.io/docs/migrations/1.x-to-2.x#cannot-find-module-mswnode-jsdom) [[2]](https://github.com/mswjs/msw/discussions/1934#discussioncomment-9494612) with node modules not being found;
- ~~Mocked HTTP code 204s needed replacing with 200. 204 seems to leave the mocked API in an error state which, while allowing the offending test to pass, breaks all tests that ran after it.~~


src/test/pages/Groups.test.tsx
```diff
@@ -156,12 +157,12 @@ describe("Groups", () => {
...
             extraEndpoints: [
-                rest.delete(API_PATH + "/groups/:groupId", async (req, res, ctx) => {
-                    const {groupId} = req.params;
+                http.delete(API_PATH + "/groups/:groupId", async ({params}) => {
+                    const {groupId} = params;
                     if (parseInt(groupId as string) === groupToDelete.id) {
                         correctDeleteRequests++;
                     }
-                    return res(ctx.status(204));
+                    return HttpResponse.json(null, {status: 200,});
```
mwtrew, line 165: Strange that `204` in here doesn't work, must be a bug. This seems to work though:

```suggestion
                    return new HttpResponse(null, {status: 204, headers: {'Content-Type': 'application/json'}});
```

mwtrew (approved):
Looks good, thanks a lot for tackling this one!

src/test/pages/Groups.test.tsx
jacbn: Ah, nice. It even works without the header being explicitly set; presumably there's a bug with the JSON factory method in particular for empty objects.

PR #1152, opened 2024-09-27
Prevent selecting UNKNOWN as gender

jacbn:


jsharkey13 (changes requested):
Defaulting to "Female" on signup without interacting with the element is not going to work for us. (Even if this is only true on Chrome on Windows).

jsharkey13 (approved):
Well it _works_ . . .

PR #1154, opened 2024-09-30
Fix coordinate feedback

axlewin:
For coordinate questions: change the way coordinates are concatenated before being validated to avoid negative numbers erroneously being flagged as invalid.

Also implement invalid character checks in the same style as numeric questions.


jacbn (changes requested):
Unfortunately we certainly want to disallow commas (see comment), which messes with the validation -- this approach will need some restructuring.

jacbn:
I have some feedback about the numeric validation, but the requirements for that have changed a bit since we started. Since this is becoming increasingly separate from the original problem, I'll suggest that we undo the changes made to `IsaacNumericQuestion` here and make a separate PR for numeric cleanup. The coordinate stuff is looking good though!

jacbn (approved):

PR #1156, opened 2024-09-30
Set the height of the code editor according to language

jacbn:
Uses a smaller height for the SQL editor, given most SQL inputs on Ada do not require many lines.

Requires [this](https://github.com/isaacphysics/isaac-code-editor/pull/6) change to the code editor.


src/app/components/content/IsaacInteractiveCodeSnippet.tsx
```diff
@@ -27,7 +28,8 @@ export const IsaacInteractiveCodeSnippet = ({doc}: IsaacInteractiveCodeProps) =>
             test: doc.testCode,
             wrapCodeInMain: doc.wrapCodeInMain,
             language: doc.language,
-            dataUrl: doc.dataUrl
+            dataUrl: doc.dataUrl,
+            iFrameHeight: iFrameHeight - 96,
```
jsharkey13, line 32: `96` you say?

PR #1157, opened 2024-10-01
Request current user for MFA and email verification required users

mwtrew:
This means the session expiry time is set properly.


sjd210:
The change makes sense to me, and I've got it working on all account types locally 👍

PR #1160, opened 2024-10-02
Create new teaching order page

axlewin:
On Ada, create a new suggested teaching order page. This is in addition to the current page until the relevant fragments are created.

### DIFF
--- a/src/app/components/pages/TeachingOrders.tsx
+++ b/src/app/components/pages/TeachingOrders.tsx
@@ -12,10 +12,18 @@ enum TEACHING_ORDER_STAGE {
     ADVANCED_Y2 = "Advanced Y2"
 }
 
+const sanitiseStageTab = (stageTab: string) => stageTab.replace(" ", "_").toLowerCase();
+const sanitisedStageTabsMap = Object.fromEntries(Object.keys(TEACHING_ORDER_STAGE).map(stage => [sanitiseStageTab(stage), stage]));
+
+const getStageFromURL = () : TEACHING_ORDER_STAGE | undefined => {
+    const urlStage = window.location.hash.slice(1);
+    return sanitisedStageTabsMap[urlStage] as TEACHING_ORDER_STAGE | undefined;
+};
+
 export const TeachingOrders = () => {
     const TEACHING_ORDER_STAGES = Object.keys(TEACHING_ORDER_STAGE);
-    const [stageTab, setStageTab] = useState<typeof TEACHING_ORDER_STAGES[number]>(TEACHING_ORDER_STAGES[0]);
-    const [stageTabOverride, _setStageTabOverride] = useState<number | undefined>(TEACHING_ORDER_STAGES.indexOf(stageTab) + 1 || undefined);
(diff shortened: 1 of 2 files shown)


src/app/components/pages/TeachingOrders.tsx
```diff
@@ -0,0 +1,55 @@
...
+    CORE_Y1 = "Core Y1",
+    CORE_Y2 = "Core Y2",
+    ADVANCED_Y1 = "Advanced Y1",
+    ADVANCED_Y2 = "Advanced Y2"
+}
+
+export const TeachingOrders = () => {
+    const TEACHING_ORDER_STAGES = Object.keys(TEACHING_ORDER_STAGE);
+    const [stageTab, setStageTab] = useState<typeof TEACHING_ORDER_STAGES[number]>(TEACHING_ORDER_STAGES[0]);
+    const [stageTabOverride, _setStageTabOverride] = useState<number | undefined>(TEACHING_ORDER_STAGES.indexOf(stageTab) + 1 || undefined);
```
jacbn, line 18: This doesn't currently have any functionality to read the hash anchor. Something like this should work:

```suggestion

const sanitiseStageTab = (stageTab: string) => stageTab.replace(" ", "_").toLowerCase();
const sanitisedStageTabsMap = Object.fromEntries(Object.keys(TEACHING_ORDER_STAGE).map(stage => [sanitiseStageTab(stage), stage]));

const getStageFromURL = () : TEACHING_ORDER_STAGE | undefined => {
    const urlStage = window.location.hash.slice(1);
    return sanitisedStageTabsMap[urlStage] as TEACHING_ORDER_STAGE | undefined;
};

export const TeachingOrders = () => {
    const TEACHING_ORDER_STAGES = Object.keys(TEACHING_ORDER_STAGE);
    const [stageTab, setStageTab] = useState<typeof TEACHING_ORDER_STAGES[number]>(getStageFromURL() || TEACHING_ORDER_STAGE.CORE_Y1);
    const [stageTabOverride, _setStageTabOverride] = useState<number | undefined>(Object.values(sanitisedStageTabsMap).indexOf(stageTab) + 1 || undefined);
```

src/app/components/pages/TeachingOrders.tsx
```diff
@@ -0,0 +1,55 @@
...
+        activeTabOverride={stageTabOverride}
+        onActiveTabChange={(n) => {
+            setStageTab(TEACHING_ORDER_STAGES[n-1]);
+        }}
+    >
+        {tabTitlesToContent}
+    </Tabs>;
+
+    useEffect(() => {
+        const sanitisedStageTab = stageTab.replace(" ", "_").toLowerCase();
```
jacbn, line 40: If you're using the above suggestion, we should use the same sanitisation function here too.
```suggestion
        const sanitisedStageTab = sanitiseStageTab(stageTab);
```

jacbn (commented):
Looks good! We'll need to add some code to read the hash anchor from the URL though, in case people want to link to a specific tab.

src/app/components/pages/TeachingOrders.tsx
jacbn: (Admittedly, this won't read changes made to the hash anchor without reloading, but we don't do that anywhere. Something for another time maybe..)

PR #1165, opened 2024-10-03
Add front-end warnings and Inequality integration for new Chemistry options

sjd210:
Adds option for coefficient scaling:
- e.g. `2 H2 + O2 -> 2 H2O`  ===>>> `H2 + 1/2 O2 -> H2O`

and for disabling certain "available symbols" such as is done with trig functions for Symbolic Maths questions.

- e.g. `_state_symbols` controls `(s), (l), (g), (aq)`

Unavailable symbols are removed from the Inequality modal menu. If state symbols are unavailable and the user tries to use one in text-input, they are warned. 

<br/>

Both of these are not used as options for Nuclear Physics questions, although the logic is applied from available symbols to disallow state symbols, brackets, equilibrium arrows, and dots from its Inequality menu.

--- 

Also fixes a minor issue where processing the `_no_alphabet` meta symbol would skip the next meta symbol from being processed. The only place I could find this impacting was [here](https://isaacphysics.org/questions/ft_lower_q38d_stage2?stage=gcse) (note the "_trigs" in the available symbols list).


src/app/components/elements/modals/inequality/utils.ts
```diff
@@ -89,7 +90,7 @@ export function generateMathsBasicFunctionsItems(): MenuItemProps[] {
     ];
 }
 
-export function generateMathsTrigFunctionItem(name: string): MenuItemProps {
+export function generateMathsTrigFunctionItem(name: string): MenuItemProps { // ohh heck
```
jacbn, line 93: this is the general vibe inequality gives off...

jacbn (approved):
I can't speak for what categories content would like in each section so some of the testing here is best left to them. The code is looking good though!

src/app/components/elements/modals/inequality/utils.ts
sjd210: 😔 You're making me tempted to keep it

jacbn (approved):

PR #1178, opened 2024-10-08
Fix Teacher Connections' fixed width container problems

jacbn:


src/scss/common/my-account.scss
```diff
@@ -148,16 +148,19 @@
     margin-top: 2px;
   }
 
+  .connections-fixed-length-container {
+    max-width: calc(100% - 91.8px);
+  }
+
```
sjd210, line 154: This is ideal spacing for the "Leave" button with a tooltip icon on Isaac, but on Ada the font size is different and we don't have this tooltip so it leaves an awkward gap at the end. We should probably do something along these lines:

```suggestion
  .isaac-connections-fixed-length-container {
    max-width: calc(100% - 91.8px);
  }
  
  .ada-connections-fixed-length-container {
    max-width: calc(100% - 60px);
  }
```

(with a corresponding `classNames` change in the .tsx)

sjd210: On a similar note now that we have it, I would also add `connections-fixed-length-container` to lines [214](https://github.com/isaacphysics/isaac-react-app/blob/b3bf58207a03f18bc5257459aeb014b68e8f458b/src/app/components/elements/panels/TeacherConnections.tsx#L214) and [268](https://github.com/isaacphysics/isaac-react-app/blob/b3bf58207a03f18bc5257459aeb014b68e8f458b/src/app/components/elements/panels/TeacherConnections.tsx#L268) (although for Ada, we will again need another separate adjustment for its size). It's a less common problem than a group with multiple teachers, but students with particularly long names cause similar problems currently.

(From my testing, adding it directly to the `<span>` class alongside `connections-fixed-length-text` does the job with the different structure of that component).

sjd210 (changes requested):

src/scss/common/my-account.scss
jacbn: We tend to do the CSS changes with different .scss files (in /phy and /cs), but the idea is the same. I also was a bit unsure putting `-container` and `-text` on the same component was the right thing to do (even if it worked!) as the `max-width` lines were intended to be nested together, so I've cleaned that up a bit by making an actual container component.

sjd210: Of course - I promise I did know this! Regardless, I prefer this restructure of the container and everything now works well.

src/scss/phy/my-account.scss
```diff
@@ -18,3 +18,14 @@
     width: max-content;
   }
 }
+
```
sjd210, line 21: Nothing up with what you've done, but if we're fixing alignments here anyway then the icon in the header box on Physics has been offset and annoying me, so we might as well get this in at the same time.

```suggestion

.connect-list {
  .connect-list-header {
    .icon-person-active {
      margin-top: 7px;
    }
  
    h4 {
      padding: 0rem 1rem;
    }
  }
}

```

sjd210 (approved):
Another vaguely-related change I noticed and would like while we're touching this code anyway, but as long as you're happy with that then this can just be merged.

PR #1182, opened 2024-10-11
Expired test changes

axlewin:
Order tests in My Tests by due date and grey out 'Start Test' button for expired tests.

### DIFF
--- a/src/app/components/pages/quizzes/MyQuizzes.tsx
+++ b/src/app/components/pages/quizzes/MyQuizzes.tsx
@@ -40,12 +40,13 @@ enum Status {
     Unstarted, Started, Complete
 }
 
+const todaysDate = new Date();
+
 function QuizItem({item}: QuizAssignmentProps) {
     const assignment = isAttempt(item) ? null : item;
     const attempt = isAttempt(item) ? item : assignment?.attempt;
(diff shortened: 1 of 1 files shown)


src/app/components/pages/quizzes/MyQuizzes.tsx
```diff
@@ -130,8 +134,31 @@ const MyQuizzesPageComponent = ({user}: MyQuizzesPageProps) => {
...
+        if (a.dueDate && b.dueDate) {
+            if (a.dueDate < b.dueDate) {
+                return 1;
+            }
+            if (a.dueDate > b.dueDate) {
+                return -1;
+            }
+        }
+        else if (a.dueDate) {
+            return 1;
```
jacbn, line 147: If A has a due date and B doesn't, A < B, I agree. If they _both_ don't have a due date, A = B, which I disagree with -- I think comparing via set date in this case makes the most sense.

jacbn: There's also a question of the relative ordering between active tests, active-but-indefinite tests, and expired tests. _If_ we want to display them in one list, `active < indefinite < expired` makes the most sense to me, but I'll get this clarified.

jacbn (commented):
Need to get the sort order clarified with content, but the partitioning is working nicely. We should definitely use set date to compare two indefinite tests in the meantime!

PR #1184, opened 2024-10-14
Change cloze dropdown breakpoint

axlewin:
Use dropdowns for cloze questions on all small screens and on medium-sized touchscreens.


jacbn:
We have some helper functions¹ `above` and `below` (used via e.g. `below['md'](deviceSize)`, which might simplify these if you want! They're inclusive of the device size referenced.

¹ (well, objects containing functions)

PR #1192, opened 2024-10-18
Remove print shadows

axlewin:
Remove the shadows around non-accordion questions and above the license disclaimer.

### DIFF
--- a/src/app/components/pages/Question.tsx
+++ b/src/app/components/pages/Question.tsx
@@ -76,7 +76,7 @@ export const Question = withRouter(({questionIdOverride, match, location, previe
         const isFastTrack = doc && doc.type === DOCUMENT_TYPE.FAST_TRACK_QUESTION;
 
         return <GameboardContext.Provider value={navigation.currentGameboard}>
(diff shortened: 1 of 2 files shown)


src/app/components/pages/Question.tsx
```diff
@@ -76,7 +76,7 @@ export const Question = withRouter(({questionIdOverride, match, location, previe
         const isFastTrack = doc && doc.type === DOCUMENT_TYPE.FAST_TRACK_QUESTION;
 
         return <GameboardContext.Provider value={navigation.currentGameboard}>
-            <Container className={`${doc.subjectId || ""}`}>
+            <Container className={classNames(`${doc.subjectId || ""}`, "no-shadow")}>
```
jsharkey13, line 79: The `classNames` helper lets you not have to worry about whether something is a string or undefined or whatever; it will just build a valid value for you. So if you're moving to using it, we don't need the nasty workaround that was there:
```suggestion
            <Container className={classNames(doc.subjectId, "no-shadow")}>
```

jsharkey13 (approved):

PR #1195, opened 2024-10-21
Remove user context URL params

mwtrew:


mlt47 (approved):
This looks good.
Glad that there are tests to describe the expected behaviour more declaratively.
https://github.com/isaacphysics/isaac-react-app/pull/1204 should also be merged for this next release to minimise the downsides of removing this query param syncing feature.

PR #1197, opened 2024-10-24
Show test visibility in My Tests

axlewin:
Show the 'visible to students'/'hidden from teachers' labels from Set Tests on My Tests > Practice Tests.


src/app/components/pages/quizzes/MyQuizzes.tsx
```diff
@@ -150,6 +151,11 @@ const MyQuizzesPageComponent = ({user}: MyQuizzesPageProps) => {
         }
     };
```
sjd210, line 153: I like the comment on the original SetQuizzes file that this logic is from to explain:

```suggestion

    // If the user is event admin or above, and the quiz is hidden from teachers, then show that
    // Otherwise, show if the quiz is visible to students
```

Otherwise, everything looks good

PR #1208, opened 2024-10-30
Default exam spec tabs to user context

axlewin:
On Ada: if an exam specification page has stage/exam board tab options that match a user's context, default to those tabs when the page is loaded.

This works best for users with a single context set; if a user (teacher or above) has multiple contexts set, the tabs will default to the last match found.


mlt47 (approved):
This all looks good to me. In the past when users (teachers) have multiple account preferences set, we've defaulted to their first choice but recently moved to not trying to default when there is more than one option. I think here, because we need to choose something I think it is fine defaulting to any one of their account preferences.

PR #1213, opened 2024-10-31
Hide context picker for non-staff users

axlewin:
For non-staff users: hide context picker.
For staff users: show all stage/exam boards options on context picker.

### DIFF
--- a/src/app/components/elements/inputs/UserContextPicker.tsx
+++ b/src/app/components/elements/inputs/UserContextPicker.tsx
@@ -29,6 +29,9 @@ export const UserContextPicker = ({className, hideLabels = true}: {className?: s
     const user = useAppSelector(selectors.user.orNull);
     const userContext = useUserViewingContext();
 
+    const allExamBoardOptions = getFilteredExamBoardOptions();
+    const allStages = getFilteredStageOptions();
+
     const filteredExamBoardOptions = getFilteredExamBoardOptions({byUser: user, byStages: [userContext.stage], includeNullOptions: true});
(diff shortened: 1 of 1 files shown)


src/app/components/elements/inputs/UserContextPicker.tsx
```diff
@@ -43,93 +45,86 @@ export const UserContextPicker = ({className, hideLabels = true}: {className?: s
...
-                                    getFilteredExamBoardOptions({byUser: user, byStages: [stage], includeNullOptions: true})
-                                        .map(eb => eb.value);
-                                // If we have possible valid exam board options but All is not one of them, use one of those.
-                                if (possibleExamBoards.length > 0 && !possibleExamBoards.includes(EXAM_BOARD.ALL)) {
-                                    examBoard = possibleExamBoards[0];
-                                }
-                                // If we only have one possible exam board besides All, use that.
-                                else if (possibleExamBoards.length === 2 && possibleExamBoards.includes(EXAM_BOARD.ALL)) {
-                                    examBoard = possibleExamBoards.filter(eb => eb !== EXAM_BOARD.ALL)[0];
+    const [currentStage, setCurrentStage] = useState<STAGE>(userContext.stage);
```
mwtrew, line 48: I can see why we're doing this, but it would be nice to get rid of it since it will nearly always just follow the current user context. We might be able to do so once the "all" settings are removed, as I've experimented with [here](https://github.com/isaacphysics/isaac-react-app/compare/hotfix/context-picker-changes...temp/context-picker-changes-experiment?expand=1). We'll come back to it!

mwtrew (approved):

PR #1214, opened 2024-11-01
Limit multiple-digit numbers in Inequality

sjd210:
Forming a multiple-digit `numberInputValue` in the Inequality Modal was unlimited, which led to bad behaviour (e.g. overflowing from its menu container, large numbers facing integer limit issues, larger numbers being converted to scientific notation).

I've run a script to check all symbolic questions on the site and found 10 total questions with answers containing 4 or more digits:
- 7 of them would be fine with 4 positive digits
- The other 3 are just a single number and so can be answered very easily with just the existing text input (and don't even really need to be symbolic questions anyway!) [e.g. [Part A](https://isaacphysics.org/questions/iodate?stage=a_level)]

4 characters is also the amount that fit cleanly in the menu hexagon, so 4 positive digits / 3 negative digits is the safest answer.

--- 

Also to get in simultaneously for the Chemistry Checker, the `hasMetaSymbols` check now also tests for empty lists. This has already been tested by Content so shouldn't need too much checking.


jacbn (approved):
Might be worth adding a warning in the editor if the answer is a single value outside the permitted range to stop content accidentally making unanswerable questions, but that is separate from this. Otherwise looks good!

PR #1215, opened 2024-11-01
Make codeTabs handle VB

axlewin:
(Ada): If a user's default language is Visual Basic, check for both Visual Basic and VB when picking a default tab.

This is just a quick fix to handle the default tab issue, it doesn't address the mix of Visual Basic and VB elsewhere.


sjd210:
I know that you were only seeking to fix the default tab issue, but I have opinions on the title-changing code and would much rather see that root problem addressed rather than a symptom. 

Essentially the code had been written to assume that there would only be one child in a `codeTab` and so was only checking its first child (it was doing the same thing in another place recently, which has already been fixed), whereas it should have been checking its first `codeSnippet` child specifically. 

This makes the name consistent WHILE forcing the tab switching. I've made the change directly on the same branch, and since its basically writing over what you've done I'd appreciate if we swap and you now review this. It should be quite straightforward.

PR #1216, opened 2024-11-04
Prevent client error on old browsers on `/tests`

jacbn:
`toSorted` is undefined on [older browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted#browser_compatibility), causing client errors in some cases. While some of these are out the scope of browsers we support, this is a painless fix with no real reason not to do.

### DIFF
--- a/src/app/components/pages/quizzes/MyQuizzes.tsx
+++ b/src/app/components/pages/quizzes/MyQuizzes.tsx
@@ -176,17 +176,17 @@ const MyQuizzesPageComponent = ({user}: MyQuizzesPageProps) => {
 
     const [completedQuizzes, incompleteQuizzes] = quizAssignments ? partitionCompleteAndIncompleteQuizzes(quizAssignments) : [[], []];
     let [overdueQuizzes, currentQuizzes] = partition(incompleteQuizzes, a => a.dueDate ? todaysDate > a.dueDate : false);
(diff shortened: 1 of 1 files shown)


src/app/components/pages/quizzes/MyQuizzes.tsx
```diff
@@ -266,7 +266,7 @@ const MyQuizzesPageComponent = ({user}: MyQuizzesPageProps) => {
                             <RS.Col xs={12} className="mb-4">
                                 <RS.Input type="text" placeholder="Filter tests by name..." value={filterText} onChange={(e) => setFilterText(e.target.value)} />
                                 <button className={`copy-test-filter-link m-0 ${copied ? "clicked" : ""}`} tabIndex={-1} onClick={() => {
-                                    filterText.trim() && navigator.clipboard.writeText(`${window.location.host}${window.location.pathname}?filter=${filterText.trim()}#practice`);
+                                    if (filterText.trim()) navigator.clipboard.writeText(`${window.location.host}${window.location.pathname}?filter=${filterText.trim()}#practice`);
```
jacbn, line 269:
```suggestion
                                    if (filterText.trim()) {
                                        navigator.clipboard.writeText(`${window.location.host}${window.location.pathname}?filter=${filterText.trim()}#practice`);
                                    }
```

jsharkey13 (approved):

PR #1217, opened 2024-11-04
Remove "All Stages/Exam Boards" from account settings

axlewin:
Remove "All Stages" (both sites) and "All Exam Boards" (Ada) from account settings dropdowns. Keep these options in the context picker (for staff users).

This causes minor inconsistencies for users who currently have these options selected, but since all content is now shown by default this should be ok.


src/app/components/elements/inputs/UserContextAccountInput.tsx
```diff
@@ -109,7 +109,7 @@ function UserContextRow({
             >
                 {getFilteredExamBoardOptions({
                     byStages: [userContext.stage as STAGE || STAGE.ALL],
-                    includeNullOptions: onlyUCWithThisStage,
+                    includeNullOptions: onlyUCWithThisStage, excludeAllExamBoards: true,
```
mwtrew, line 112: Is there a reason not to use the existing `includeNullOptions` flag (`= false`) here?

mwtrew: Same goes for `getFilteredStageOptions()`

axlewin: good point - will change them now

mwtrew (approved):

mwtrew:
I'm not going to merge this just yet -  I feel like we'll get a lot of questions about the inconsistencies for users who currently have these options selected you mentioned, so it would be best to migrate their settings in the DB to a new default e.g. Core/Ada at the same time. We'll do this next week.

PR #1220, opened 2024-11-07
Disallow blank admin user searches

axlewin:
In the admin user manager: require at least one filter to be set when performing a search.


src/test/pages/AdminUserManager.test.tsx
```diff
@@ -153,7 +153,7 @@ describe("AdminUserManager", () => {
             ]
         });
         await navigateToUserManager();
-        await searchWithParams({}, {expectNumberOfResults: 1, searchHandler});
+        await searchWithParams({role: mockUser.role}, {expectNumberOfResults: 1, searchHandler});
```
sjd210, line 156: This test (although skipped anyway) has "ensure that default search parameters are set" in its title - which is no longer done. 

I'd suggest taking it out of this title and creating a straightforward new test along these lines with `expect(searchHandler).toHaveBeenCalledTimes(0);` or similar to make sure things are as comprehensive as before.

jacbn (approved):

jacbn (approved):

PR #1222, opened 2024-11-12
Add admin search postcode validation

axlewin:
Add a regex to verify that a postcode is of a sensible format in admin user search before sending it to the API. The regex should match complete postcodes (e.g. CB3 0FD) and outcodes (e.g. CB3).

This check doesn't verify that a postcode actually exists, so a server error can still be thrown in some cases, but this should reduce the chances of this happening.

### DIFF
--- a/src/app/components/pages/AdminUserManager.tsx
+++ b/src/app/components/pages/AdminUserManager.tsx
@@ -111,8 +111,13 @@ export const AdminUserManager = () => {
 
     const search = (event: React.FormEvent<HTMLFormElement>) => {
         event.preventDefault();
-        adminSearchResultsRef.current?.scrollIntoView({behavior: "smooth"});
-        searchUsers(searchQuery);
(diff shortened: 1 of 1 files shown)


jacbn (approved):

src/app/components/pages/AdminUserManager.tsx
```diff
@@ -111,8 +111,13 @@ export const AdminUserManager = () => {
 
     const search = (event: React.FormEvent<HTMLFormElement>) => {
         event.preventDefault();
-        adminSearchResultsRef.current?.scrollIntoView({behavior: "smooth"});
-        searchUsers(searchQuery);
+        if (searchQuery.postcode && !/^[A-Z]{1,2}[0-9]{1,2}[A-Z]{0,1} ?([0-9][A-Z]{2})?$/i.test(searchQuery.postcode)) {
```
jacbn, line 114: This does technically permit AANNA which is not a [valid format](https://assets.publishing.service.gov.uk/media/5a81ebbded915d74e6234d42/Appendix_C_ILR_2017_to_2018_v1_Published_28April17.pdf), but given this is a niche exception on an admin-only page I am not worried.

If you did want to fix this, AA?N(N|A)? _should_ be what you want.

PR #1224, opened 2024-11-12
Fix tooltip z-index issues

axlewin:
Fixes an issue where some tooltips fade in strangely due to elements behind them having a non-zero z-index.

### DIFF
--- a/src/scss/common/tooltip.scss
+++ b/src/scss/common/tooltip.scss
@@ -1,6 +1,6 @@
 div.fade:has(> .tooltip) {
     position: absolute;
-    z-index: 1040;
+    z-index: 1080;
     width: 100%;
   }
   
--- a/src/scss/common/tooltip.scss
+++ b/src/scss/common/tooltip.scss
@@ -1,4 +1,4 @@
-.tooltip-zindex {
+div.fade:has(> .tooltip) {
     position: absolute;
     z-index: 1040;
     width: 100%;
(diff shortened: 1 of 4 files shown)


src/scss/common/tooltip.scss
```diff
@@ -0,0 +1,6 @@
+.tooltip-zindex {
```
jacbn, line 1: Instead, I think we should apply this to tooltips in general like this:
```suggestion
div.fade:has(> .tooltip) {
```
This obviously had wider-ranging impact, but all `.fade`s with a `.tooltip` child should be specific enough to reference only BS tooltips.

src/scss/common/tooltip.scss
```diff
@@ -0,0 +1,6 @@
...
+    position: absolute;
+    z-index: 1040;
```
jacbn, line 3: I'm not sure where you originally got the 1040 from, but it led me down a fun rabbit hole. [This](https://getbootstrap.com/docs/5.0/layout/z-index/) documents how Bootstrap _should be_ defaulting `.tooltip`'s z-index to 1080, though Reactstrap's `UncontrolledTooltip` adds an extra div around this that seems to be ignoring the 1080. In truth, it only needs to be 20 or above to match the header. I don't really mind, but maybe we should be consistent with BS for now.

jacbn (commented):
We should probably apply these changes to all tooltips universally rather than each one individually; it can happen with the `/events` slot, for example, with the Help hoverable at the top in specific cases on mobile.

PR #1225, opened 2024-11-13
Account self-deletion

jacbn:
Adds a new modal `AccountSelfDeletionModal` and new pages `AccountDeletion` (at `/deleteaccount`) and `AccountDeletionSuccess` (`/deleteaccount/success`) to manage the deletion flow.

Logging out still uses the old action/reducers rather than slices, and since new endpoints should be using slices, this unfortunately requires mixing the two. The slices in `accountApi.ts` supply the new interaction with the API, and logging out is managed through the changes in `/state/actions/index.tsx`. We cannot use the existing logout functionality as this redirects to `/` (see `userConsistencyChecker`), so we make a parallel a/r pair. This allows us to create our own redirect.


src/scss/cs/button.scss
```diff
@@ -365,10 +365,8 @@ a:not(.btn):not(.nav-link) {
   }
 }
 
-strong {
-  a {
-    color: $dark-pink-300 !important;
-  }
+strong a, a strong {
+  color: $dark-pink-300 !important;
```
sjd210, line 369: Since you've styled `.btn.btn-inline-link` essentially identically to a `strong a` now anyway, should we combine both of the css expressions in this file to the same line too? I think it's quite likely that any future changes would want to be applied to both now, since they're mostly the same thing.

```suggestion
strong a, a strong, .btn.btn-inline-link {
  color: $dark-pink-300 !important;
```

src/app/components/pages/AccountDeletion.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const queryParams = useQueryParams();
+    const token = queryParams.token as string | undefined;
+    const user = useAppSelector(selectors.user.orNull);
+    const [deletionReason, setDeletionReason] = useState<DELETION_REASON>(DELETION_REASON.UNKNOWN);
+    const [confirmDeletion, setConfirmDeletion] = useState(false);
+    const [deleteAccount, {isLoading: isDeletingAccount}] = useDeleteAccountMutation();
+
+    return <Container className="pb-5">
+        <TitleAndBreadcrumb currentPageTitle={siteSpecific("Account Deletion", "Account deletion")} className="mb-4" />
+        {!user ? <p>You must be logged in to delete your account.</p> :
```
sjd210, line 50: Since this only checks for a logged-in user, it allows for an account other than the one being deleted to access this page. They won't then be able to actually delete the account because it's checked in the backend upon submitting the form (so it's not a security issue), but I could forsee this being confusing.

I'm particularly thinking about a case such as where they have multiple accounts and want to delete _**Account 1**_ and then log in with _**Account 2**_ because it is their new main account and THEN they click the link in the email. The page will show up as normal, but the deletion won't work.

I think that the page should either block a user from accessing it if they have a token which is invalid/expired for their account (so probably requiring an API interaction on page load) OR at least a confirmation of which account you're deleting somewhere on the page. Maybe both?

sjd210 (commented):
I've a couple of small style thoughts, and then a slightly broader one about accessing the page with an invalid token/the wrong account. 

Otherwise the changes look very good and the full through-line seems to work clearly and consistently.

src/app/components/pages/AccountDeletion.tsx
```diff
@@ -0,0 +1,69 @@
...
+            isTutorOrAbove(user) ? <p>Only student accounts can be deleted automatically. Please <Link to="/contact?preset=accountDeletion">contact us</Link> to request account deletion.</p> : 
+                !token ? <p>You must have a valid token to delete your account. Visit your <Link to="/account">account page</Link> to generate a new email with a token.</p> : 
+                    <>
+                        <PageFragment fragmentId="account_deletion_final_warning" />
+                        <p>Please select the reason for deleting your account:</p>
+                        <DeletionReasonForm className="ms-2" deletionReason={deletionReason} setDeletionReason={setDeletionReason} />
+                        <FormGroup check className="mt-4 ps-0">
```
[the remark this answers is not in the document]
jacbn: Hm, this looks like a Firefox bug; something about the `.tick` class used in the checkbox being `inline-block`. I'm not convinced it needs the `inline` part anyway, as it's positioned by a flex container. I've made the tick `display: block` (and lengthened the short line by 1px... it was unbalanced 😣); this seems to fix the Firefox issue and I can't see any regressions elsewhere on any browser.

The `check` prop is used on [all the checkbox examples](https://6-4-0--reactstrap.netlify.app/components/form/) on the RS docs, hence why I was a little cautious to remove it.

src/app/components/pages/AccountDeletion.tsx
jacbn: We certainly shouldn't have an endpoint for checking if a token is valid, this would be a fairly major security concern... but we could use a separate token for validity and deletion, for example. We discussed this in the standup, and the conclusion was that given the token is only active for 24 hours and we believe the number of people who actively use two accounts is small, it isn't worth the extra effort to build anything now. There is ample opportunity throughout the process for them to contact us if something doesn't work so they should know to email us if this scenario occurs and deletion fails. It's a great thought though, and something to keep in mind if we see any tickets about it!

src/scss/cs/button.scss
jacbn: Yep, good idea. There's also the content links on the line above this. Not quite sure the link ones belong in `button.scss`, but imo consistency is more important and combining them as you suggest guarantees that.

src/app/components/pages/AccountDeletion.tsx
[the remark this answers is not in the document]
sjd210: Hrm, I'm still not convinced that the `check` prop is necessary. It seems to just be used to add the `form-check` class to the component, the styling from which doesn't actually make a difference here.

That said, I agree it's working now and I can't find any other issues with checkboxes from that change, so let's stick with it.

src/app/components/pages/AccountDeletion.tsx
sjd210: Fair enough. I suppose that edge-case situation where this gets confusing isn't worse than our current situation of needing to manually process every delete request individually so there's no need to add extra complication (and I suppose the "`HELLO <NAME>`" in the nav-bar is still visible and should give *some* indication of which account you're logged into too, which should also help).

PR #1229, opened 2024-11-19
Distinguish Practice Tests

jacbn:
- Splits the listing of available practice tests into its own page, `/practice_tests`, accessible via the Learn nav;
- Creates a new default table view for `/tests` (though the original card view is still accessible), which displays the title, assigner and relevant dates;
- Splits assigned tests from practice tests into separate tabs at the topmost level of `/tests`; further filtering inside these categories (for started, not yet started, overdue and completed) can be achieved via the filters menu

### DIFF
--- a/src/app/components/pages/quizzes/MyQuizzes.tsx
+++ b/src/app/components/pages/quizzes/MyQuizzes.tsx
@@ -7,14 +7,13 @@ import {Link, RouteComponentProps, useHistory, withRouter} from "react-router-do
 import * as RS from "reactstrap";
 
 import {ShowLoading} from "../../handlers/ShowLoading";
-import {QuizAssignmentDTO, QuizAttemptDTO, QuizSummaryDTO, RegisteredUserDTO} from "../../../../IsaacApiTypes";
+import {QuizAssignmentDTO, QuizAttemptDTO, RegisteredUserDTO} from "../../../../IsaacApiTypes";
 import {TitleAndBreadcrumb} from "../../elements/TitleAndBreadcrumb";
 import {formatDate} from "../../elements/DateString";
 import {AppQuizAssignment} from "../../../../IsaacAppTypes";
 import {
     extractTeacherName,
     isAttempt,
-    isEventLeaderOrStaff,
     isFound,
     isTutorOrAbove,
     partitionCompleteAndIncompleteQuizzes,
@@ -22,12 +21,11 @@ import {
 } from "../../../services";
 import {Spacer} from "../../elements/Spacer";
 import {Tabs} from "../../elements/Tabs";
-import {useGetAvailableQuizzesQuery} from "../../../state";
 import {PageFragment} from "../../elements/PageFragment";
 import { CardGrid } from "../../elements/CardGrid";
 import partition from "lodash/partition";
 
-interface MyQuizzesPageProps extends RouteComponentProps {
+export interface QuizzesPageProps extends RouteComponentProps {
     user: RegisteredUserDTO;
 }
 
@@ -121,9 +119,8 @@ function QuizGrid({quizzes, empty}: AssignmentGridProps) {
     </>;
 }
 
-const MyQuizzesPageComponent = ({user}: MyQuizzesPageProps) => {
+const MyQuizzesPageComponent = ({user}: QuizzesPageProps) => {
 
-    const {data: quizzes} = useGetAvailableQuizzesQuery(0);
     const {data: quizAssignments} = useGetQuizAssignmentsAssignedToMeQuery();
     const {data: freeAttempts} = useGetAttemptedFreelyByMeQuery();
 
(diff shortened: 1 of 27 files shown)


src/app/components/pages/quizzes/MyQuizzes.tsx
```diff
@@ -222,82 +313,117 @@ const MyQuizzesPageComponent = ({user}: MyQuizzesPageProps) => {
...
+
+    const emptyAssignedMessage = <span className="text-muted">{!quizAssignments || quizAssignments.length === 0
+        ? "You have no tests in progress."
+        : "No tests match your filters."
+    }</span>;
 
-    return <RS.Container>
+    const emptyPracticeMessage = <span className="text-muted">
+        No practice tests match your filters. Take some new tests <Link to="/practice_tests">here</Link>!
+    </span>;
```
axlewin, line 384: This link appears on both sites, but `/practice_tests` doesn't exist on Ada.

I also like the distinction between "no tests at all" and "no filtered tests" as in `emptyAssignedMessage`, but a nested conditional here probably isn't the best solution.

```suggestion
        const emptyPracticeMessage = <span className="text-muted">{isPhy
        ? !freeAttempts || freeAttempts.length === 0
            ? <>You have no practice tests in progress. Take some new tests <Link to="/practice_tests">here</Link>!</>
            : <>No practice tests match your filters. Take some new tests <Link to="/practice_tests">here</Link>!</>
        : !freeAttempts || freeAttempts.length === 0
            ? <>You have no practice tests in progress.</>
            : <>No practice tests match your filters.</>
    }</span>;
```

src/app/components/pages/quizzes/PracticeQuizzes.tsx
```diff
@@ -0,0 +1,91 @@
...
+        {isEventLeaderOrStaff(user) && quiz.hiddenFromRoles && quiz.hiddenFromRoles?.includes("TEACHER") && <div className="small text-muted d-block ms-2">hidden from teachers</div>}
+        {isTutorOrAbove(user) && ((quiz.hiddenFromRoles && !quiz.hiddenFromRoles?.includes("STUDENT")) || quiz.visibleToStudents) && <div className="small text-muted d-block ms-2">visible to students</div>}
+    </>;
+
+    return <Container>
+        <TitleAndBreadcrumb currentPageTitle={siteSpecific("Practice Tests", "Practice tests")} />
+        <PageFragment fragmentId="help_toptext_practice_tests" />
+        <ShowLoading until={quizzes}>
+            {quizzes && <>
+                <h3>Available</h3>
```
axlewin, line 55: This heading was added to split the page up when we moved in progress & available practice tests under the same tab. Now that they're separated again, do we still need it?

axlewin (changes requested):

src/app/components/pages/quizzes/MyQuizzes.tsx
jacbn: Ah, it is supposed to exist on Ada. That simplifies the conditional enough that it feels nice. I'm also inclined to only have the link on if there are no tests at all, as if you're using the filters you're there to find an existing test, not take a new one.

src/app/components/pages/quizzes/PracticeQuizzes.tsx
```diff
@@ -0,0 +1,90 @@
...
+                        <div className="d-flex flex-grow-1 flex-column flex-sm-row align-items-center p-3">
+                            <div>
+                                <span className="mb-2 mb-sm-0 pe-2">{quiz.title}</span>
+                                {roleVisibilitySummary(quiz)}
+                                {quiz.summary && <div className="small text-muted d-none d-md-block">{quiz.summary}</div>}
+                            </div>
+                            <Spacer />
+                            {isTutorOrAbove(user) && <div className="d-none d-md-flex align-items-center me-4">
+                                <Link to={{pathname: `/test/preview/${quiz.id}`}}>
+                                    <span>Preview</span>
```
axlewin, line 76: When viewing this page as a teacher, the option to preview a test disappears completely below `md` - there could be an "Actions" dropdown as is used for small screens on Set Tests?

(This isn't a new issue though, so not critical for merging this).

src/app/components/pages/quizzes/MyQuizzes.tsx
```diff
@@ -1,213 +1,305 @@
...
+        <Row>
+            <Col xs={6}>
+                <Label className="w-100">
+                    <span className={"text-nowrap"}>Filter by quiz title</span>
+                    <Input type="text" data-testid="title-filter" onChange={(e) => setQuizTitleFilter(e.target.value)} />
+                </Label>
+            </Col>
+            <Col xs={6}>
+                <Label className="w-100">
+                    <span className={"text-nowrap"}>Filter by assigner</span>
```
axlewin, line 227: This filter is available for practice tests, which don't have assigners - is this intentional to keep everything aligned? Would greying it out be an option?

src/app/components/pages/quizzes/PracticeQuizzes.tsx
jacbn: I believe there's a plan to add filtering to tests pretty soon, which will involve overhauling it a bit; let's wait until then.

src/app/components/pages/quizzes/MyQuizzes.tsx
jacbn: Greying it out would probably best, to keep the page flow the same when switching tabs. There's also the problem of what to do about the `Not started` and `Expired` states in the state filter which are invalid states for practice tests, so will never do anything. If we switch from assigned tests to practice tests with these filters set, should we remove them? Grey them out? If we do remove them, should they return if we switch back to the assigned tests tab?

I don't think there's an easy answer, and it would be nice to get the team's feedback once they have had a play around. I've made a card so we can have this discussion separately and push this through for now.

axlewin (approved):

PR #1230, opened 2024-11-20
Boolean logic text entry for all

jsharkey13:
Enable text entry on symbolic logic questions for all users, and make it look slightly less cursed than it did before.


jacbn (approved):
We should make Inequality inputs consistent across all question types (perhaps with some varying text to indicate which question type it is), especially on Ada, but we can do that separately.

PR #1231, opened 2024-11-21
Link to teacher emails from teacher features

axlewin:


src/app/components/pages/TeacherFeatures.tsx
```diff
@@ -98,11 +98,12 @@ export const TeacherFeatures = () => {
             </Col>
             <Col xs={0} sm={3} className="d-md-none"/>
             <Col sm={6} md={4}>
-                <IsaacCard doc={{ clickUrl: "/pages/isaac_embedded_schools",
-                    image: {src: "/assets/phy/icons/teacher_features_sprite.svg#groups"},
-                    title: "Teacher Ambassadors",
+                <IsaacCard doc={{ clickUrl: isTeacherOrAbove(user)  ? "/teacher_emails" : "/pages/contact_us_teacher",
```
jsharkey13, line 101: Jacob is right, you don't need the conditional here. But only because there's an easier way to implement the _route_ that will give you the conditional redirect for free.

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -138,6 +139,7 @@ export const RoutesPhy = [
     <StaticPageRoute key={key++} exact path="/alevel" pageId="alevel" />,
     <TrackedRoute key={key++} exact path="/teacher_features" component={TeacherFeatures}/>,
     <TrackedRoute key={key++} exact path="/tutor_features" component={TutorFeatures}/>,
+    <TrackedRoute key={key++} exact path="/teacher_emails" component={TeacherEmails}/>,
```
jsharkey13, line 142: Instead of making a new component and making a new route like this, you could instead add it to the `Teacher Pages` bit below using `<StaticPageRoute>` like the mentoring pages.

I would also like it if we could force the fragment ID to be `teacher_emails_frag` so that it is more unique. Then the `pageId` you need for the route would be `fragments/teacher_emails_frag`.

jsharkey13 (changes requested):

jacbn (approved):

jsharkey13 (approved):

PR #1235, opened 2024-11-26
Autofix ESLint warnings and RS imports

jacbn:
Clean up various pieces of ESLint- and RS-related tech debt accumulated over the years.


mwtrew (approved):
Looks good to me. As for preventing this in future, it's best to address these things continuously. CodeQL does some of this, but not all - perhaps we can configure it (or something else) to use our ESLint config and surface these warnings on PRs.

PR #1237, opened 2024-11-27
`/tests` regression fixes

jacbn:
- Corrects the href value for practice test links
- Prevents the past tests toggle moving slightly when clicked


src/app/services/quiz.ts
```diff
@@ -353,7 +353,7 @@ export function convertAttemptToQuiz(attempt: QuizAttemptDTO): DisplayableQuiz |
         completedDate: attempt.completedDate,
         attempt: attempt,
         quizFeedbackMode: attempt.feedbackMode,
-        link: attempt.completedDate ? `/test/attempt/${attempt.id}/feedback` : `/test/attempt/${attempt.id}`,
+        link: attempt.completedDate ? `/test/attempt/${attempt.quizId}/feedback` : `/test/attempt/${attempt.quizId}`,
```
axlewin, line 356: This fixes the links for in progress tests, but the feedback links still seem to need the attempt id rather than the quiz id
```suggestion
        link: attempt.completedDate ? `/test/attempt/${attempt.id}/feedback` : `/test/attempt/${attempt.quizId}`,
```

axlewin: Ah, this is only the case for practice tests...

PR #1242, opened 2024-12-03
User context improvements

mwtrew:
Various updates to the context picker and viewing contexts:

- Remove now-inapplicable "unusual context" logic - always show the tooltip with an explanation for the current viewing context.
- Fix problems with "no content on this page" message not showing on Topic pages. This now relies on the audience field in the topic summary content object being correct, as with Concept pages.
- Add a reset button to the context picker, which clears the transient context if set.

Best reviewed with whitespace changes hidden.


src/app/components/elements/inputs/UserContextPicker.tsx
axlewin, line 1: When logged-out (Ada) users view a Topic page, the context picker defaults to Core/Ada, when really the context used is All stages/Ada. This isn't a new issue, but showing a tooltip which contradicts the context picker could be confusing.
image

mwtrew: Thanks, I missed that. I'll see what I can do!

src/app/components/elements/inputs/UserContextPicker.tsx
```diff
@@ -16,24 +18,36 @@ import {
 } from "../../../services";
 import {selectors, transientUserContextSlice, useAppDispatch, useAppSelector,} from "../../../state";
 
+const contextExplanationMap: {[key in CONTEXT_SOURCE]: string} = {
+    [CONTEXT_SOURCE.TRANSIENT]: "these context picker settings",
+    [CONTEXT_SOURCE.REGISTERED]: "your account settings",
+    [CONTEXT_SOURCE.GAMEBOARD]: `the ${siteSpecific("gameboard", "quiz")} settings`,
+    [CONTEXT_SOURCE.DEFAULT]: `${SITE_TITLE_SHORT}'s default settings`,
+    [CONTEXT_SOURCE.NOT_IMPLEMENTED]: "the site's settings"
```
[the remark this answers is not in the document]
axlewin: Just in the interest of useful test output, I'd suggest something like "(not implemented on phy)", but this would lose consistency with the others so maybe it's fine as-is.

PR #1243, opened 2024-12-05
Add new header

jacbn:
Replaces the old header and navigation with a new header (navigation to come). On desktop, this is displayed as a row of links to various pages and log in/out; on `md` and below this collapses into an offcanvas wiht the same links displayed as a column.


src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -1,105 +1,80 @@
...
+                                    {user && (user.loggedIn 
+                                        ? <Link to="/logout">Log out</Link>
+                                        : <Button color="solid" size="sm" tag={Link} to="/login">Sign up / log in</Button>
+                                    )} 
+                                </div>
+                            </>
+                            : <>
+                                {/* mobile hamburger menu */}
+                                <div className="d-flex justify-content-end align-items-center flex-wrap p-3 gap-3">
+                                    <Button color="solid" size="sm" tag={Link} to="/login">Sign up / log in</Button>
```
axlewin, line 46: This button is visible for logged-in users.

src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -1,105 +1,80 @@
...
 export const HeaderPhy = () => {
     const user = useAppSelector(selectors.user.orNull);
-    const streakRecord = useAppSelector(selectors.user.snapshot);
     const mainContentId = useAppSelector(selectors.mainContentId.orDefault);
     const deviceSize = useDeviceSize();
+    const [menuOpen, setMenuOpen] = useState(false);
+    const toggleMenu = () => setMenuOpen(m => !m);
+
     return <header className="bg-white" data-testid={"header"}>
         <Container className="container-fluid px-0">
```
axlewin, line 19: The designs have the header using more of the screen's width than it does currently (particularly noticeable on larger screens) - removing this container could be an option?

jacbn: Yep, that should fix itself when we change the containers. Haven't got there yet!

jacbn:
I am aware of the lack of admin settings cog next to "log out" on mobile view, but since it's built inside the `<Nav/>` for consistency with everywhere else it's a bit of a pain to split out. I'm tempted to suggest we don't need it, given it's still accessible from the offcanvas and it's unlikely we'd be using admin tools away from a desktop, but let me know if you think otherwise.

src/app/components/site/phy/HeaderPhy.tsx
```diff
@@ -1,105 +1,80 @@
...
-                                </React.Fragment>}
-                        </div> */}
+                        {above["lg"](deviceSize) 
+                            ? <>
+                                {/* desktop menu bar */}
+                                <div className="d-flex justify-content-end align-items-center flex-wrap p-3">
+                                    <NavigationPanelPhy className={classNames("flex-row")}/>
+                                    {user && (user.loggedIn 
+                                        ? <Link to="/logout">Log out</Link>
+                                        : <Button color="solid" size="sm" tag={Link} to="/login">Sign up / log in</Button>
```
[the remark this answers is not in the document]
axlewin: Hmm, still not fixed for me. It looks fine without the container, so possible it'll fix itself when those are changed, but there's probably a nicer solution.

PR #1245, opened 2024-12-06
Inequality producing empty canvases at root level on resize

jacbn:
If `makeInequality` is called with an undefined ref, p5 proceeds to generate a new canvas at root level with seemingly fullscreen sizing. With an adjacent change to [inequality](https://github.com/isaacphysics/inequality/pull/10), this forces the element to always be defined and will quietly cancel the setup in the known cases when this occurs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "html-to-text": "^9.0.5",
     "identity-obj-proxy": "3.0.0",
     "immer": "^9.0.21",
-    "inequality": "1.1.4",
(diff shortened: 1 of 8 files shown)


jacbn (commented):
Is not required, but should be paired with the updated version of inequality that forces the element type to be defined.

sjd210:
What strange behaviour from p5 :/ Anyway, this certainly seems to fix it and better typing is always good regardless 👍

PR #1246, opened 2024-12-06
Fix back button on consistency error using stale data

jsharkey13:
If you log out in another tab, a consistency error page is shown that should clear the local Redux state and make you logged out in these other tabs too. However we had issues with the browser back button, which for obscure reasons from the HTTP 1.1 specification, will ignore cache headers and use stale data to reload the previous page.

We could fix this by disabling caching for the current user data, but the back button does not misbehave on the tab where you actually log out. This suggests a fix; add an extra history entry for the error page and then hard-navigate to it. Then the back button takes you to the page from before the error, but without loading stale data.


mwtrew (approved):
Works nicely!

PR #1251, opened 2024-12-18
Support n-dimensional coord Qs

axlewin:
Extend coordinate questions to support any number of dimensions by replacing the  x and y properties with a list of strings (and similarly for placeholder values).

This should be merged after the corresponding [content changes](https://github.com/isaacphysics/rutherford-content/pull/548) and alongside the API [validator changes](https://github.com/isaacphysics/isaac-api/pull/663).


jsharkey13 (approved):

src/app/components/content/IsaacCoordinateQuestion.tsx
```diff
@@ -51,32 +50,26 @@ export const coordinateInputValidator = (input: string[][]) => {
...
-        {remove && <Button className="ms-3" size="sm" onClick={remove}>Delete</Button>}
+    const {value, placeholderValues, numberOfDimensions, onChange, readonly, remove} = props;
+    return <span className="coordinate-input">{[...Array(numberOfDimensions)].map((_, i) =>
+        <>
+            <Input
+                key={i}
+                type="text"
+                className="force-print"
+                placeholder={placeholderValues[i] ?? ""}
+                value={value.coordinates ? value.coordinates[i] : ""}
```
jacbn, line 61: `value.coordinates[i]` is undefined on first render. Since you're controlling the value of this, it's a controlled component, which [shouldn't have undefined values](https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable). 
```suggestion
                value={isDefined(value.coordinates?.[i]) ? value.coordinates[i] : ""}
```

src/app/components/content/IsaacCoordinateQuestion.tsx
```diff
@@ -51,32 +50,26 @@ export const coordinateInputValidator = (input: string[][]) => {
...
-            placeholder={placeholderYValue ?? "y"}
-            value={value.y ?? ""}
-            onChange={event => onChange({...value, y: event.target.value === "" ? undefined : event.target.value})}
-            readOnly={readonly}
-        />
-        )
-        {remove && <Button className="ms-3" size="sm" onClick={remove}>Delete</Button>}
+    const {value, placeholderValues, numberOfDimensions, onChange, readonly, remove} = props;
+    return <span className="coordinate-input">{[...Array(numberOfDimensions)].map((_, i) =>
+        <>
```
jacbn, line 55: You have a `map` iterator, so all direct children need a unique key. The "child" in this case is a fragment, which annoyingly means ESLint doesn't register this, but since both the `Input` and `span.coordinate-input-separator` are direct children in the _generated HTML_, there's a problem. You can either make sure both have a unique key (the Input already has `key={i}`, I suppose you could use something like ``key={`sep-{i}`}`` for the separator), or replace the fragment with a `span` and move the `key={i}` from the `Input` up to the `span`.

jacbn (commented):
Couple of React-y points. It's worth checking the console to see if anything is complaining as part of your testing :)

src/app/components/content/IsaacCoordinateQuestion.tsx
```diff
@@ -51,32 +50,26 @@ export const coordinateInputValidator = (input: string[][]) => {
...
-        <span className="coordinate-input-separator">,&nbsp;</span>
-        <Input
-            type="text"
-            className="force-print"
-            placeholder={placeholderYValue ?? "y"}
-            value={value.y ?? ""}
-            onChange={event => onChange({...value, y: event.target.value === "" ? undefined : event.target.value})}
-            readOnly={readonly}
-        />
-        )
```
jacbn, line 74: Did we mean to remove these brackets?

PR #1256, opened 2025-01-13
Upgrade app dependencies

jacbn:
All dev dependencies so not of great concern. A [new vulnerability](https://github.com/isaacphysics/isaac-react-app/security/dependabot/124) in `cross-spawn` is a dependency of `jest` which hasn't seen an update in a year.


jacbn:
Updating cypress seemed to add new focus boxes around elements that remain in the image, which was causing every image to fail. This isn't strictly bad, so long as it's consistent going forwards -- but the update also seemed to break my local cypress instance such that I wasn't able to start the tests (notably did _not_ break the git instance). I'll look into this more at some point, but no need to block the rest of the updates on this.

sjd210:
We've looked into the Cypress update a bit more now. To document that here, there seem to be issues with Mac's virtualisation of the Docker containers that Cypress uses. The update isn't pressing, so we can just hold off until this is fixed or we find a workaround.

As for the focus boxes, they _can_ be removed using `.blur()` but I agree that as long as the behaviour remains consistent we can just keep them.

PR #1262, opened 2025-01-14
Use GitHub Actions for Docker build

mwtrew:
See https://github.com/isaacphysics/isaac-adrs/blob/main/7-automated-docker-builds.md.


Dockerfile
```diff
@@ -1,10 +1,24 @@
-FROM nginx:stable
+ARG BUILD_TARGET="build-phy"
+
+FROM node:20.12.0-buster AS builder
```
jsharkey13, line 3: This is a major version change from our current process, right?

Also, I don't see `node:20.12.0-buster` listed on the Docker Hub page, so I'm not sure it's a recommended tag name? It seems to me that we should be using `22.13.0-bookworm` really (though that's another major version up again!). Otherwise, `20.18.1-bookworm`?

jsharkey13: Buster EOL'ed in 2022, so we should not be using it now. Bookwork has another 18 months yet.

mwtrew: It is [there](https://hub.docker.com/layers/library/node/20.12.0-buster/images/sha256-1970a4eb353baf1f6671ad5cdb368ad76a6443890f7705b74f1f9c1ecc2e4df1), but I agree that it would be better to use Bookworm now.

jsharkey13 (approved):

PR #1264, opened 2025-01-14
Implement abstract list view

sjd210:
An “abstract” component for displaying a list of items, akin to a group of `ContentSummaryListGroupItem`s in the current codebase. 

I've implemented this separately to `ContentSummaryListGroupItem` itself to rewrite it in a `<Row... <Col..`. structure without touching Ada, but once we know how the surrounding code works, it may be worth eventually re-merging these.

This will need an implementation wrapper for each of **Cards, Questions, Question Packs, Quick Quizzes, Concepts, Tests**, inputting their relevant icon and specifying which other fields are needed e.g. a test needs `title, previewUrl, testUrl` and uses the list icon.

For quick reference, XL in **physics** context and XS in **neutral** context:


jacbn (approved):
We might need to fiddle with a few things:
- padding on the level tag at `xs`
- vertical position of the title as `lg`+

but I think it'll be easier to do this when we have various concrete implementations in place. Looking really good for now so happy to merge!

PR #1265, opened 2025-01-14
Create pod component for news/events on homepage

axlewin:


src/app/components/pages/News.tsx
```diff
@@ -41,8 +41,8 @@ export const News = () => {
             /> : 
             <>
                 <Row className="d-flex flex-row card-deck row-cols-1 row-cols-sm-2 row-cols-lg-3 justify-content-center mt-4">
-                    {allNews.map((n, i) => <Col key={i} className="d-flex my-3 px-0 justify-content-center">
-                        <NewsCard key={n.id} newsItem={n} showTitle />
+                    {allNews.map((n, i) => <Col key={i} className="my-3 px-0 justify-content-center">
+                        <NewsPod key={n.id} newsItem={n} showTitle />
```
jacbn, line 45: This is used for both Isaac and Ada, so we're replacing Ada cards here too. `NewsCard` is already `siteSpecific`, and the only use for that card on Physics was here, so you're safe to put the `NewsPod` code inside `PhysicsNewsCard` (and I'd rather you did, as otherwise it's not being used).

src/scss/phy/cards.scss
```diff
@@ -1,62 +1,72 @@
...
+  flex: none;
+
+  .pod-title {
+    font-size: 16px;
+    font-weight: 700;
+  }
+  
+  .pod-img {
+    border-radius: 8px;
+    height: 158px;
```
jacbn, line 33: I'd use `@aspect-ratio` here (believe we're now aiming for 16:9 though most images will be 5:3 at the moment) rather than set an exact height, then have `object-fit: cover` on the `> img` so it doesn't stretch if there's an incorrectly sized image.

src/scss/phy/cards.scss
```diff
@@ -1,62 +1,72 @@
...
+      text-align: right;
+      color: white;
 
-// number of news cards per line for each breakpoint
-$news-cards-per-line: (
-  xs: 2,
-  lg: 3
-);
+      &.disabled {
+        z-index: 5;
```
jacbn, line 62: Does this do anything?

src/scss/phy/cards.scss
```diff
@@ -1,62 +1,72 @@
...
+      }
 
-@each $name, $breakpoint in $grid-breakpoints {
-  @if (map-has-key($news-cards-per-line, $name)) {
-    @include media-breakpoint-up($name) {
-      .card-deck .news-card {
-        flex: 0 0 calc(#{math.percentage(math.div(1, map-get($news-cards-per-line, $name)))} - #{$grid-gutter-width});
-        width: 100%;
+      > b {
```
jacbn, line 65: They seem to be using a different font to the body for this. It's embedded in the SVG, but it looks like Exo2 semibold to me.

```suggestion
      > b {
          font-family: $secondary-font-semi-bold;
```

src/app/components/elements/NewsPod.tsx
```diff
@@ -0,0 +1,45 @@
...
+        {image && <a href={url} className="focus-target">
+            <CardImg
+                top
+                src={image.src && apiHelper.determineImageUrl(image.src)}
+                alt={image.altText || `Illustration for ${title}`}
+                className="pod-img"
+            />
+        </a>}
+        <CardBody>
+            <b className="pod-title">
```
jacbn, line 24: Boldness doesn't seem to be working as intended on `hanken` fonts, I'll look into this separately. Don't worry about fixing it here!

jacbn (changes requested):
I think the most important point is that we need to make sure Ada stays intact while we're doing this, so the Ada pathways need to remain as they were. For news pods, you should just be able to apply the changes you've made in `NewsPod` to `PhysicsNewsCard` and undo the change from `NewsCard => NewsPod`. Event cards probably need their own component to keep Ada intact.

A second major point is the event and news cards have a longer form as shown on p16 of the designs. There's a longer comment about this below.

Nothing major about the vertical pod itself though, nice stuff.

src/app/components/elements/NewsPod.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {Link} from "react-router-dom";
+import {Card, CardBody, CardImg, CardProps, CardText} from "reactstrap";
+import {IsaacPodDTO} from "../../../IsaacApiTypes";
+import {apiHelper} from "../../services";
+
+interface NewsPodProps extends CardProps {
+    newsItem: IsaacPodDTO;
+}
+
+const NewsPod = ({newsItem, ...props}: NewsPodProps) => {
```
jacbn, line 11: See comment on EventPod about a longer layout; the same applies for news pods on `/news`.

src/scss/phy/cards.scss
```diff
@@ -1,62 +1,72 @@
...
+        width: 100%;
+      }
+    }
+  }
+}
+
+.pod {
+  background-color: transparent;
+  border: none;
+  width: 318px;
```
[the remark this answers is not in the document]
axlewin: ah, that was a hack to get the images to look ok - will fix this at the same time as the aspect ratio changes below

src/scss/phy/cards.scss
axlewin: Copied from the old implementation, but you're right - removing both instances of these lines doesn't seem to do anything.

PR #1267, opened 2025-01-16
Set Inequality text editor mode strictly

sjd210:
Currently, `editorMode` is not always instantiated for inequality’s text input. Often this doesn't matter, but when each mode has a different valid input (e.g. the exponent docking point being removed for Chemistry but kept in Maths) and `editorMode` has not been instantiated, it defaults to the less expressive input.

Text entry normally copies parameters from symbolic entry, but this causes a race condition. `editorMode` can simply be hard-coded into each editor’s text entry box to avoid this.


jsharkey13 (approved):
It's not great having the same logic encoded in multiple places, and it's not clear to me why hard-coding it at this point won't prevent it being overridden later by the race/issue. (But if the issue is caused by it being undefined, rather than wrong, I think I agree this will work reliably).

Let's try it for now!

PR #1271, opened 2025-01-21
Page context & page titles

jacbn:
- Adds a barebones implementation of subject landing pages
- Adds state management for the current page context for questions and landing pages
- Updates the existing `<TitleAndBreadcrumb/>` component to match bolder design, and add icons for major pages
- Revert to `<Container/>`s to lower maintenance burden of keeping Ada intact


src/app/components/elements/PageTitle.tsx
```diff
@@ -109,7 +117,10 @@ export const PageTitle = ({currentPageTitle, subTitle, disallowLaTeX, help, clas
         }));
     }
 
-    return <h1 id="main-heading" tabIndex={-1} ref={headerRef} className={`h-title h-secondary d-sm-flex ${className ? className : ""}`}>
+    return <h1 id="main-heading" tabIndex={-1} ref={headerRef} className={classNames("h-title h-secondary d-sm-flex", {"align-items-center py-4": isPhy}, className)}>
+        {icon && (
+            icon.type === "img" ? <img src={icon.icon} alt="" className="me-3"/> 
+                : icon.type === "hex" ? <PhyHexIcon icon={icon.icon} subject={icon.subject}/> : undefined)}
```
sjd210, line 123: I think part of the reason this looks so odd right now for pages that we don't have an icon for is that the height of this heading bar also gets considerably smaller from not being bulked out by the icon's height.

If we don't end up adding icons everywhere (which we should come to a decision about first), it may be worth adding a `min-height: 75px` to the heading div here? Food for thought.

src/app/services/context.ts
```diff
@@ -0,0 +1,7 @@
+import { Stage } from "../../IsaacApiTypes";
+import { Subject } from "../../IsaacAppTypes";
+import { HUMAN_STAGES, HUMAN_SUBJECTS } from "./constants";
+
+export function getHumanContext(pageContext?: {subject?: Subject, stage?: Stage}): string {
+    return `${pageContext?.stage ? (HUMAN_STAGES[pageContext.stage] + " ") : ""}${pageContext?.subject ? HUMAN_SUBJECTS[pageContext.subject] : ""}`;
+}
```
sjd210, line 7: Not a problem, but I'm very tickled by having a function called `getHumanContext()`. It feels like our codebase is an alien or robot trying to understand human emotion :)

src/app/components/elements/svg/PhyHexIcon.tsx
```diff
@@ -0,0 +1,16 @@
...
+    subject?: Subject;
+}
+
+export const PhyHexIcon = (props: PhyHexIconProps) => {
+    const {icon, subject, className, ...rest} = props;
+    return <div {...rest} className={classNames("d-flex pe-3", className)} {...(subject && {"data-bs-theme": subject})}>
+        <div className="phy-hex-icon"/>
+        <i className={classNames("icon", icon)}/>
+    </div>;
+};
```
sjd210, line 16: Since this is just for a set-size (75px) `phy-hex-icon`, this doesn't currently quite work for splashing this icon style everywhere on the site. It's fine in this branch now, but I've had to add a `size` parameter on my own branch to allow switching size - we can discuss that when it comes to it if necessary.

sjd210 (commented):
Looks good and feels intuitive to use in my branch already 👍 

A couple of comments here but nothing to act on.

PR #1272, opened 2025-01-22
Implement student dashboard

axlewin:
This uses the `sm` design on `md` to avoid the panels being too crowded.


jacbn:
This looks really good! I'm not convinced groups should be first, but this is a design thing I'll settle with content. I've suggested a few cleanup things [here](https://github.com/isaacphysics/isaac-react-app/pull/1275); I'd just go through commit-by-commit to make sure you know why I'm suggesting each (hopefully the commit messages are clear!), but otherwise this is looking super clean :)

jacbn (approved):

PR #1273, opened 2025-01-22
Subject-specific pages

jacbn:
Adds routes to subject-specific versions of:
- Question Finder
- Concepts listing
- Practice tests

Adds routes to and basic implementations of:
- Lessons and revision
- Quick quizzes
- Question packs

### DIFF
--- a/src/app/components/pages/Concepts.tsx
+++ b/src/app/components/pages/Concepts.tsx
@@ -10,6 +10,7 @@ import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
 import {ShortcutResponse, Tag} from "../../../IsaacAppTypes";
 import {IsaacSpinner} from "../handlers/IsaacSpinner";
 import { SubjectSpecificConceptListSidebar, MainContent, SidebarLayout } from "../elements/layout/SidebarLayout";
+import { useUrlPageTheme } from "../../services/pageContext";
 
 // This component is Isaac Physics only (currently)
 export const Concepts = withRouter((props: RouteComponentProps) => {
@@ -17,6 +18,7 @@ export const Concepts = withRouter((props: RouteComponentProps) => {
     const dispatch = useAppDispatch();
     const user = useAppSelector(selectors.user.orNull);
     const concepts = useAppSelector((state: AppState) => state?.concepts?.results || null);
+    const pageContext = useUrlPageTheme();
 
     const subject = useAppSelector(selectors.pageContext.subject);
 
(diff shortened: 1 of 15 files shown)


src/app/components/pages/Concepts.tsx
```diff
@@ -83,7 +85,7 @@ export const Concepts = withRouter((props: RouteComponentProps) => {
     const shortcutAndFilteredSearchResults = (shortcutResponse || []).concat(filteredSearchResults || []);
 
     return (
-        <Container id="search-page">
+        <Container id="search-page" { ...(pageContext?.subject && { "data-bs-theme" : pageContext.subject })}>
```
sjd210, line 88: I missed this issue in the general context branch, but for a neutral context on this page, the message from `SidebarLayout.tsx` now says 
> _"The concepts shown on this page have been filtered to only show those that are relevant to **undefined**."_

It also keeps the explanation and link to its own page. We could probably remove this entirely (or otherwise perhaps rewrite it and add links to other contexts?).

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -72,7 +87,7 @@ export const RoutesPhy = [
     // Student test pages
     <TrackedRoute key={key++} exact path="/tests" ifUser={isLoggedIn} component={MyQuizzes} />,
     <Redirect key={key++} from="/quizzes" to="/tests" />,
-    <TrackedRoute key={key++} exact path="/practice_tests" ifUser={isLoggedIn} component={PracticeQuizzes} />,
+    <TrackedRoute key={key++} exact path="/practice_tests" component={PracticeQuizzes} />,
```
sjd210, line 90: `/practice_tests` still exists as a general page but `lessons_and_revision`/`quick_quizzes`/`question_packs` doesn't.
(Of course so do `/questions` and `/concepts`, but those are more central and part of the proposed site structure).

Do we have plans for this? Is is something that needs to be discussed still? 
If not, are we keeping this as-is? Removing the old page? Adding new neutral ones? I don't have much opinion without context 🤷 but I'd like to know.

sjd210 (commented):
All the subject-specific routing seems to work, and the basic page implementations do in-fact lead to pages. 

Other than a small change to `SidebarLayout`, this is good to go. We've also already discussed the `data-bs-theme` ordering/inheritance strangeness, and I'd like to see that working, but it's not necessary for this to get merged.

src/app/components/pages/Concepts.tsx
jacbn: Ah true, I wrote that with the intention of only using that sidebar on subject-specific concept pages; the generic one has a fairly different sidebar that should be a `<GenericConceptsSidebar/>`. I built the `<GenericConceptsSidebar/>` template before generic concepts page was ready, so it hasn't been being used yet.

src/app/components/site/phy/RoutesPhy.tsx
jacbn: This is... a good question. I'll raise it with content. We no longer have a link to the existing `/practice_tests` in the nav, so it's impossible to get there without having the URL directly. I'd assume we'd be removing the neutral one in this case, but will get confirmation.

jacbn:
Also, the fix to colour inheritance is [here](https://github.com/isaacphysics/isaac-react-app/pull/1276)!

PR #1274, opened 2025-01-22
Implemented List Views

sjd210:
`<ListView>` should now replace every instance of `<LinkToContentSummaryList>` with a matching implementation. A list can be built up of multiple different types of item for e.g. Search Page and Wildcards, so this is handled with a single list component with a switch statement for individual items. 

It can also be used for new `<ListViewCard>` items like for a `<ListViewSubjectCard>` on the homepage of [this demo branch](https://github.com/isaacphysics/isaac-react-app/tree/redesign/homepage-card-list).

List views have been added to the following pages (just the list itself, nothing around them):
- Question Finder
- Concepts
- Set Tests
- My Progress
- Search

They have notably not been added to pages such as Gameboard/My Gameboards/My Tests as these each require more than a simple 1-to-1 replacement, however the ListViewItem for them exist for once those pages are rewritten.

This could also do with some new icons, but this branch is already big enough so I'm leaving that for later.


jacbn (commented):
This is really good! I wouldn't worry about the iconography etc, we can manage that when we build specific pages; the point of this branch was just to add the ability to do that in future, alongside some barebones implementations.

The only points I have are about the `ListViewCards` component and children; the `li` cards aren't wrapped in a `ul` like the others here, which is problematic for screenreaders. I've suggested a replacement structure to that whole component that uses a `row-cols-{x}` layout instead, it needs a little SCSS reformatting too but it does simplify the construction a bit too. Secondly, we can tidy up the `@color-mode` SCSS for the tags used on these cards quite a bit.

I've made these changes on a separate branch [here](https://github.com/isaacphysics/isaac-react-app/tree/redesign/list-view-suggestions). This branch was built off the one you used to test the homepage card view, so I'd recommend cherry-picking whichever of the commits from there into this branch; I don't think the first commit is applicable, but the others should be if you're happy with them.

sjd210:
I'm happy with all these changes, thanks for the work! I'll merge all of this in now 🎉

PR #1275, opened 2025-01-22
Dashboard suggestions

jacbn:
Applies some suggestions for the dashboard.

Couple of more important points than implementation detail; we'd much prefer managing spacing with margins rather than `<br/>` elements, as we can use margin everywhere and in all directions, which isn't the case with `<br/>`s. We also really shouldn't use fixed heights on dynamic elements like the assignment card, as if the title is long everything'll overflow.


src/app/components/elements/StudentDashboard.tsx
```diff
@@ -37,27 +38,29 @@ const GroupJoinPanel = () => {
...
+                type="text" placeholder="Enter code"
+                onChange={(e: React.ChangeEvent<HTMLInputElement>) => setAuthenticationToken(e.target.value)}
+                onKeyDown={(e) => {if (e.key === 'Enter') {
+                    e.preventDefault();
+                }}}
+            />
+            <Button onClick={processToken} outline color="secondary">
+                Connect
+            </Button>
+        </InputGroup>
```
axlewin, line 53: This definitely looks better for most screen sizes, but around 600px the input box gets too small for a group code (or the placeholder text) - it might look better with the button below the text input in this case?

jacbn: I noticed this, but my hope is that if we can get the group code box switched with the work to do one, this won't be an issue.

PR #1276, opened 2025-01-23
Fix colour inheritance

jacbn:
We use CSS variables to manage styling because they inherit, and a component will take the value defined by the closest parent with such a definition. This means if we have a `biology` region inside a `physics` region, components inside the `biology` region inherit that colour as we'd expect.

SASS variables, only existing at compile time, do not have any idea of inheritance; the last-defined variable will have the highest precedence as per standard CSS rules. Given the definition of `$isaac-color-palettes` and how we loop through these to generate CSS, this means the priority order is `neutral` < `biology` < `chemistry` < `maths` < `physics`. If any component styles are generated through this loop and only reference SASS vars, they will use this priority order.

The only downside to this approach is, because SASS *functions* generate their CSS at compile time, and CSS variables are dynamic (i.e. not known at compile time), we can't use CSS variables inside these functions. There isn't a universally-working way around this ([discussion](https://stackoverflow.com/q/51888205)), but I've been able to refactor all existing uses of these functions so it's not currently a problem.


jacbn:
(did not mean for the tests to run, they are expectedly broken on the redesign branch. only 35 failing so far though!

PR #1292, opened 2025-01-31
Implement new homepage design

axlewin:


jacbn:
Looking good! The image currently only shows for `xl`, whereas it should display with a different layout for the other sizes too. It's also the raw image, which is _enormous_ and may make people slightly sad. :( 

I've shrunk the image down and added some very rough layout code for the missing screen sizes. I'll give it another scan in a bit; given it's the homepage it's quite important that it's right on all screen sizes, all the time.

PR #1298, opened 2025-02-04
Redesign FAQ page

axlewin:
This includes some general changes to accordions; in particular, borders have been added for open accordions on non-content pages. If multiple accordions are open, they all have borders; it's not clear from the designs whether this is the expected behaviour.

The video tabs should be displayed inline rather than above one another, but I couldn't find a way to do this while keeping each video in its own component


jacbn:
I have a feeling that the "video tabs" are intended to be a regular tabs component (IsaacTabs), each with a video as its only child. This would still require a change to those content pages, but means we shouldn't have the code for tabs inside IsaacVideo but instead IsaacTabs. It _might_ be best to do this alongside e.g. the hint accordions on question pages so we have a solid example we can play around with while building.

For now, then, I'll revert the changes to IsaacVideo but leave a link to the commits here so we can reuse the code for Tabs when we get to them (I'll be building questions soon, will get to it then if noone else has).

jacbn:
Okay, I've made some suggestions [here](https://github.com/isaacphysics/isaac-react-app/pull/1307). Accordions with parts aren't perfect yet – will resolve when building questions.

PR #1300, opened 2025-02-07
Breadcrumbs & improved context

jacbn:
Changes the structure of breadcrumbs on Isaac to follow a more "path-based" structure where every link in the path is clickable. Main changes in `navigation.ts`.

Also restructures `pageContext` to allow a list of stages on a given page. This allows the page to better reflect the context of the question/context page and thus allows more flexibility in what should be displayed across the page.

Happy to walk someone through this if needed!


sjd210 (commented):
This all works nice and neatly. Probably worth keeping an eye on the concepts pages themselves to make sure nothing unexpected happens with the breadcrumbs once they've been tagged - but I trust it'll work the same as questions (and it certainly does for those that have already been tagged e.g. `cb_stats_bio_intro`)

I notice that all the `subjectSpecificPages` other than `/questions` (like Concepts, Quick Quizzes) don't have the breadcrumb yet. Is the plan just to show a demo on questions, then get it working on each of the other pages when they're developed?

Assuming this is the case (and once the merge conflict is dealt with), this looks good to add in :)

jacbn:
> I notice that all the `subjectSpecificPages` other than `/questions` (like Concepts, Quick Quizzes) don't have the breadcrumb yet

Ah, good point. I was testing it on questions; the line
```ts
const crumb = isPhy && isDefinedContext(pageContext) && generateSubjectLandingPageCrumbFromContext(pageContext);
```
should generate the `intermediateCrumb` for all of those pages. I've added the code for this to concepts and quick quizzes, I don't believe any other subject-specific pages exist yet (pre-made boards, lessons...).

sjd210:
Mhm, I was thinking of this list on RoutesPhy:

```ts
const subjectSpecificPages : Record<string, React.ComponentType<RouteComponentProps<any>> | React.ComponentType<any> | undefined> = {
    // at all valid paths matching `/:subject/:stage${string}`, render the given component
    "": SubjectLandingPage,
    "/questions": QuestionFinder,
    "/concepts": Concepts,
    "/lessons_and_revision": LessonsAndRevision,
    "/practice_tests": PracticeQuizzes,
    "/quick_quizzes": QuickQuizzes,
    "/question_packs": QuestionPacks,
};
```

but all of these still need to be implemented and/or rejigged. We ought to just merge this and deal with them later.

PR #1305, opened 2025-02-10
Question Finders

sjd210:
Adds new question finder page layouts and styling. 

I'm still not entirely happy with the actual filter panel specifics - we probably ought to change some font size/spacing/animation/etc. still to bring them more in-line with the designs - however I've been working on this for far too long without anyone else's eyes on it so I won't mess with these quite yet.

Probably the most significant implication of this branch for the redesign outside the QF pages themselves is that I've changed `StyledCheckbox` logic to use the new Figma iconset for Phy. Some of it is still probably quite messy (e.g. I've used an `icon-checkbox-bg` element to colour the background of an empty box, but the `background` property to do so on a partial box)

Points to note:
- This required rewriting the Topics search structure from a 2D array to a tree. Since depth matters for determining Subject/Field/Topic, and to better fit the existing code/site, this is in the form of an array of dictionaries. This is the example I wrote while developing:
```
When a user has selected both Maths as a subject and Biology > Ecology > Ecosystems as a topic:

-----------------------------------------------------------------------------
CHOICES (all of which are now visible simultaneously)
[
	{
		"Subject": ["Biology", "Chemistry", "Physics", "Maths"]
	},
	{
		"Biology": ["Cell Biology", "Biochemistry", "Genetics", "Physiology", "Ecology", "Evolution"], 
		"Maths": ["Number", "Algebra", "Geometry", "Functions", "Calculus", "Statistics"]
	}, 
	{
		"Ecology": ["Populations", "Ecosystems", "Nutrient Cycles", "Biodiversity"]
	}
]


-----------------------------------------------------------------------------

SELECTIONS (the checked boxes)
[
	{
		"Subject": ["Biology", "Maths"]
	},
	{
		"Biology": ["Ecology"], 
	}, 
	{
		"Ecology": ["Ecosystems"]
	}
]

-----------------------------------------------------------------------------
```
-  ~~The backend will need an update at some point in the future to be able to search across subject/field/topic lines. In this example case, anything from Maths OR Ecosystems as these are the leaf nodes. They are currently being sent to the API in the same 2D array format as before, but this can change fairly easily depending on how we implement it.~~

- ~~The books on a QF page are filtered depending on page context. I've manually labelled them here, but we might want a universal constant for these instead EDIT: This has been done on the [teacher dashboard branch](https://github.com/isaacphysics/isaac-react-app/blob/653a82f1118631467a69dc295b73ab568b66d70a/src/app/components/elements/modals/IsaacBooksModal.tsx). Shouldn't be hard to copy it over.~~

- ~~Question finder filters are stored in the URL (so it can be linked to, maintained on a hard reload, etc). Right now this includes the subject and stage for the current page context. I've left this as-is for now, but we could probably rewrite some of that logic to remove the page-inherent filters from the URL but still maintain the current behaviour.~~


src/app/components/elements/TitleAndBreadcrumb.tsx
```diff
@@ -54,7 +54,7 @@ type TitleAndBreadcrumbProps = BreadcrumbTrailProps & PageTitleProps & {
     breadcrumbTitleOverride?: string;
     children?: ReactElement | boolean;
 };
-export const TitleAndBreadcrumb = ({modalId, children, breadcrumbTitleOverride, currentPageTitle, subTitle, disallowLaTeX, className, audienceViews, help, collectionType, intermediateCrumbs, preview, icon}: TitleAndBreadcrumbProps) => {
+export const TitleAndBreadcrumb = ({modalId, children, breadcrumbTitleOverride, currentPageTitle, subTitle, description, disallowLaTeX, className, audienceViews, help, collectionType, intermediateCrumbs, preview, icon}: TitleAndBreadcrumbProps) => {
```
jacbn, line 57: I am not certain that the description should be here – indeed, in the designs even the _title_ is not in the breadcrumb, let alone any subtext. However, if we are moving things, we'll also need to move at least the subtitle too, so I'm not against keeping it here for now as then we can do all the moving to and from the same place.

jacbn (approved):
As you say, there is some work to be done with styling but I will comment on that once you've finished working on it. For now, I have a few minor suggestions on [this](https://github.com/isaacphysics/isaac-react-app/pull/1319) PR, but on the whole this is looking great so far! :)

I would also check the Ada VRTs are okay; these are quite broad changes and I would not be surprised if we had broken a couple of things.

sjd210:
This should now work with functional topic search logic, Ada VRTs, searchable sub-learning stages (e.g. "A-Level" and "Further A" for /a_level), and unredundant URLs.

I tried out a few different stylings, and this is the happiest I am with it all - even if it doesn't map exactly to the designs. In particular, using the smaller checkboxes universally in the sidebar made them feel too small and annoying to click. I did also *want* to add an expanding animation similar to `CollapsibleList` to the Topic menu, but nesting that within an existing `CollapsibleList` was causing problems so I've left it 🤷 Open to suggestions.

jacbn:
Looking good! A few things could still use some improved styling:
- The list of applied filters at the top should probably use the human-readable name of e.g. tags rather than the tag itself
    - (also the "clear all filters" ought to be capitalised)
-  Some of these tags are AND-ed, others are OR-ed; this is a somewhat eternal problem that we have yet to come to a nice way of solving, so perhaps we ignore it for now, but something to ponder;
- Should the A-level Physic**al** Chemistry book be called as such instead of Physic**s** Chemistry in the label?

As for your point about the expanding animation, `CollapsibleList` uses a `data-targetHeight` param that allows setting a child's *eventual* height instead of its current height, so that two collapsible lists can animate linearly at the same time (otherwise the changing height of the inner one changes the speed at which the outer one animates). If this is the problem you were having, you could *try* setting `data-targetHeight` in whichever child component you have. I remember this taking a good while to get right when I was building `CollapsibleList`, though, and it looks fine without so I wouldn't spend too much time on it.

sjd210:
Human-readable tags and the "A-Level Physical Chemistry" book-name have been fixed :)

---

In terms of the AND/OR-ing behaviour of the tags, that's a matter for the API - but the main offender seems to just be Books, and that's already been the case for a while. Probably worth fixing at some point, but I'm not concerned with it now.

As for animation, one of the main issues with this new layout is having more than one possible field on the same line in a container means you can't just add up all the sub-div heights to calculate the necessary `data-targetHeight` of the containing div (which is how we do it for `CollapsibleList`). I've got an idea for how we could eventually do this which I might bury in the backlog for later - but it'll be too time-consuming for now and agree it certainly isn't necessary for the page.

PR #1308, opened 2025-02-10
Create teacher dashboard

axlewin:


axlewin:
This includes changing the groups page to use a hash anchor to handle the tab logic so that individual group tabs can be linked to from the dashboard. This affects both sites;  it's not used for anything on Ada, but I don't think it's worth adding any site-specific logic.

Also note that the dashboard uses an existing component from the groups page - when the group page is redesigned there may be side effects here.

src/app/components/elements/TeacherDashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+                    {assignmentsCount > 0 && <span className="badge bg-primary rounded-5 ms-2">{assignmentsCount > 99 ? "99+" : assignmentsCount}</span>}
+                </Link>
+                <Link to="/progress" className='d-block panel-my-isaac-link'>
+                    My progress
+                </Link>
+                <Link to="/tests" className='d-block panel-my-isaac-link'>
+                    My tests
+                    {quizzesCount > 0 && <span className="badge bg-primary rounded-5 ms-2">{quizzesCount > 99 ? "99+" : quizzesCount}</span>}
+                </Link>
+            </div>
```
sjd210, line 135: I don't think we need all these student links here if we can simply toggle between student/teacher views to see them anyway.

Removing them also gives more horizontal space to work with, which should let us be able to make some more adjustments depending on screen-width.

(This was also causing problems at small screen-sizes anyway, overflowing the maximum height of a `dashboard-panel`)

src/app/components/elements/TeacherDashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+        </div>
+        <div className="section-divider" />
+        <Link to="/account" className="panel-my-isaac-link">
+            My account
+        </Link>
+        <Link to="/logout" className="panel-my-isaac-link">
+            Log out
+        </Link>
```
sjd210, line 165: We certainly don't need log out here. It's not in the student version, and is always available from the top menu-bar anyway.

src/app/components/elements/TeacherDashboard.tsx
```diff
@@ -0,0 +1,224 @@
...
+                    <AssignmentsPanel />
+                </Col>
+                <Col className="mt-4">
+                    <BooksPanel />
+                </Col>
+                <Col className="mt-4">
+                    <MyIsaacPanel />
+                </Col>
+            </Row>
+        </div>;
```
sjd210, line 222: I think we can use some similar logic to the `StudentDashboard` to make use of this space better (especially wrt the book panel), and make it less jarring to switch between the view in terms of components moving around the screen.

I've got something that I think works well in my suggestion branch.

src/scss/phy/homepage.scss
```diff
@@ -17,6 +17,13 @@
     text-wrap: nowrap;
   }
 
+  h5 {
+    font-size: 14px;
+    font-family: $primary-font;
+    color: $color-neutral-500;
+    line-height: 1.1;
+  }
```
sjd210, line 25: I'm not sure this was the right place to put this. I see that you wanted the `STUDENT`/`TEACHER` My Isaac text to match the one in the menu-bar, but then it was applying to all `h5`s on the page including the ones in the Current Work panel.

Either way, I'm still advocating for removing the STUDENT half of the My Isaac panel here so we can just remove this all.

src/scss/phy/homepage.scss
```diff
@@ -191,6 +211,25 @@ $dashboard-panel-padding: 30px;
   }
 }
 
+.scrollable-panel {
+  overflow-y: scroll;
```
sjd210, line 215: We ought to use `auto` rather than `scroll`. Otherwise the scrollbar sticks around even when it's not needed, which is a little ugly.

src/app/components/elements/StudentDashboard.tsx
```diff
@@ -174,19 +179,19 @@ export const StudentDashboard = () => {
                     </Row>
                     <Row>
                         <Col className="mt-4">
-                            <CurrentWorkPanel />
+                            <GroupJoinPanel />
```
sjd210, line 182: I'm not sure if the switching of Current Work/Join Group was meant to also apply to this `"lg"` layout, but I don't like it as much. The long searchbar looks odd compared to the long assignments. If I forget to ask Jacob about this later, remind me.

sjd210 (commented):
Nice! The panel logic is looking good, and it feels odd that we didn't already have book-subject mapping somewhere so I'm glad that now exists.

I've got a couple of thoughts, mostly about layout/spacing, that I've put into a [branch here](https://github.com/isaacphysics/isaac-react-app/pull/1323). As always, do what you will with it.

There's also a merge conflict, but it just seems to be over imports so should be simple enough to resolve.

src/app/components/elements/StudentDashboard.tsx
sjd210: Fair enough. I don't have a strong opinion so I'm happy being outvoted on this one 🤷

sjd210:
Awesome, I do definitely like the simplified assignment view more for that tab (I hadn't even noticed them before, but now that you've pointed it out they did look quite clunky!) 

This is good to go in 👍

PR #1314, opened 2025-02-13
Redesign question component

jacbn:
Uses accordion work started in [this](https://github.com/isaacphysics/isaac-react-app/pull/1298) PR; have this merged first!

Redesign the question component. This is used both inside and outside accordions, with various question types. This could use a keen eye when testing to make sure various questions of different types look okay on both sites!

This does **not** change the layout of question pages themselves; that is to come next.


axlewin (commented):
Looks good. A couple of minor layout points:

- The padding on quick questions now looks slightly strange, especially if there's other text in the accordion (see e.g. `/questions/rolling_objects`). This might look better with less vertical padding?
-  Accordions without titles (see regression test page) don't have chevrons and have very small headers. Not sure if this affects any live pages, though.

PR #1316, opened 2025-02-14
Redesign phy groups page

axlewin:
This uses some changes made in https://github.com/isaacphysics/isaac-react-app/pull/1308, which should be merged before this is reviewed.


src/app/components/pages/Groups.tsx
```diff
@@ -526,6 +525,120 @@ export const Groups = ({user}: {user: RegisteredUserDTO}) => {
...
+    }, [hashAnchor]);
+    const selectedGroup = (isLoading || isFetching) ? undefined : groups?.find(g => g.id === selectedGroupId);
+
+    const createNewGroup: (newGroupName: string) => Promise<boolean> = async (newGroupName: string) => {
+        setShowArchived(false);
+        return createGroup(newGroupName).then(async (result) => {
+            if (mutationSucceeded(result)) {
+                const group = result.data;
+                if (!group.id) return false;
+                dispatch(showGroupInvitationModal({group, user, firstTime: true, backToCreateGroup: () => setSelectedGroupId(undefined)}));
```
jacbn, line 634: This modal isn't showing up for me. I've had a quick look but nothing is jumping at me. Any ideas?

src/app/components/elements/layout/SidebarLayout.tsx
```diff
@@ -494,8 +495,44 @@ export const SetAssignmentsSidebar = (props: SetAssignmentsSidebarProps) => {
...
+    setSelectedGroupId: React.Dispatch<React.SetStateAction<number | undefined>>;
+    showArchived: boolean;
+    setShowArchived: React.Dispatch<React.SetStateAction<boolean>>;
+    groupNameInputRef: RefObject<HTMLInputElement>;
+    createNewGroup: (newGroupName: string) => Promise<boolean>;
+}
+
+export const GroupsSidebar = (props: GroupsSidebarProps) => {
+    const { user, groups, allGroups, selectedGroup, setSelectedGroupId, showArchived, setShowArchived, groupNameInputRef, createNewGroup } = props;
+    return <ContentSidebar buttonTitle="Select or create a group" {...props}>
```
jacbn, line 532: You're passing in a bunch of methods to a div here, which is causing a bunch of console warnings because they're not valid attributes. I think you mean to pass the `...rest` props, which are defined by `GroupsSidebarProps` extending `SidebarProps` (which itself extends divprops, iirc).

jacbn (commented):
Works nicely, have suggested some changes [here](https://github.com/isaacphysics/isaac-react-app/pull/1321). There's a bug I mention in there that's worth checking.

src/app/components/pages/Groups.tsx
axlewin: Good catch! This seems to be happening because of the `setSelectedGroupId` immediately after the modal is dispatched, which triggers a `useEffect` - the problematic line is `window.location.hash = selectedGroupId.toString()`, which causes an `ACTIVE_MODAL_CLOSE`. I have a fix [here](https://github.com/isaacphysics/isaac-react-app/compare/redesign/page/groups...groups-modal-bugfix?expand=1) which just moves the modal dispatch to after the hash is changed, but there might be a neater way to deal with this - any thoughts would be appreciated :)

jacbn: Ahhh, nice. Changing the hash causes a soft reload which removes the modal. We could just set the hash early, so the `useEffect` doesn't change anything:
```suggestion
                // setting the hash early here prevents the useEffect from changing the hash and causing a reload after the modal is dispatched, which would remove it.
                window.location.hash = group.id.toString(); 
                dispatch(showGroupInvitationModal({group, user, firstTime: true, backToCreateGroup: () => setSelectedGroupId(undefined)}));
```

PR #1325, opened 2025-02-20
Redesign question page

jacbn:
The designs are not quite finalised yet; will remove commented code once complete.


src/scss/phy/questions.scss
```diff
@@ -188,3 +188,21 @@
...
+  margin: 1rem 0;
+  font-size: 14px;
+
+  > div > span:first-child {
+    color: $color-neutral-800;
+    margin-bottom: 0.4rem;
+  }
+
+  > div + div {
+    border-left: 1px solid $color-neutral-200;
```
sjd210, line 206: I don't think this vertical bar should be here for this panel on `"xs"` screens (or `"sm"` with Stage & Difficulty specifically). It looks strange at the beginning of the line - in fact Fluent have horizontal bars at narrow widths, which I think looks even better.

sjd210 (changes requested):
A couple of small things but much cleaner! It's nice seeing major pages like this come together 🥳 

Also _(not actually touched on here and we've already discussed in-person briefly, but to note it somewhere)_, I really don't like how hints are displayed on single-part questions. Without a background to act as enclosure, they feel too disconnected from the Hint buttons - especially with the page credit right below. 

On a similar note, expanding these hints on <= `"sm"` screens causes the buttons to rise rather than the buttons remaining stationary and the hint contents coming down. I've figured out the specific problem in this case is to do with the sidebar layout and `order-0`/`order-1` classes. If the main content is first in HTML-order and page-order, this problem doesn't exist, but being first in page-order but second in HTML-order causes this weird rendering. It might be worth enforcing this ordering for content sidebars in TS directly rather than CSS.

sjd210 (approved):

PR #1328, opened 2025-02-21
Redesign events pages

axlewin:
There are a few aspects of the designs that would require content changes:

- Button styles - these are sometimes hardcoded in the content
- Consistent heading styles - also in the content
- "In-person events" filter - this would require a new tag (unless we assume anything not explicitly tagged as virtual is in-person?)

Also note that there's no VRT for Events, so there's likely some small changes I've missed that affect Ada - but since Ada don't have any events this should be fine.


src/app/components/pages/EventDetails.tsx
```diff
@@ -162,7 +164,251 @@ const EventDetails = ({match: {params: {eventId}}, location: {pathname}}: EventD
...
+            };
+
+            const PhyEventDetails = <Container>
+                <TitleAndBreadcrumb
+                    currentPageTitle="Events" icon={{type: "hex", icon: "page-icon-events"}}
+                    breadcrumbTitleOverride="Event details" intermediateCrumbs={[EVENTS_CRUMB]}
+                />
+                <EditContentButton doc={event}/>
+                <Row className="mb-4">
+                    <Col className="col-3 col-sm-2 col-lg-1 me-3 pt-2">
```
sjd210, line 336: I don't think this needs to be a Row/Col layout, we could just have two horizontally adjacent divs, which'll give us more control over the spacing rather than the big gap here.

From my quick play around in HTML, this should do:
```ts
<div className="d-flex mb-3">
   <div className="me-3">
      [... Event Hex Code ...]
   </div>
   <div className="mt-n2">
      [... Title Code ...]
   </div>
</div>
```

src/app/components/pages/EventDetails.tsx
```diff
@@ -162,7 +164,251 @@ const EventDetails = ({match: {params: {eventId}}, location: {pathname}}: EventD
 
             const checkTeacherStatus = isPhy && event.isATeacherEvent && !isTeacherOrAbove(user);
 
-            return <Container className="events mb-5">
+            const isTeacherEvent = event.tags?.includes("teacher") && !event.tags?.includes("student");
+            const isStudentEvent = event.tags?.includes("student") && !event.tags?.includes("teacher");
+
+            const KeyEventInfo = () => {
+                return <Table borderless className="my-3">
```
sjd210, line 171: I don't think this should be a `Table` anymore. It should be fairly easy to convert to a `<Row> <Col>` format (combined with the "Book event" button you already have like this), which'll give more control over the spacing rather than there being a huge gap on "xl" screens.

Alternatively, if you feel like playing with the css a little more, we could have it even more adaptive with a `flex-column` layout and some `Spacer`s, but I don't think this is necessary.

sjd210 (commented):
I've got a few specific points, and a few general thoughts. 

Otherwise really nice :) I think the events page was one of the most cramped-feeling pages on the site with the old layout, so it's nice seeing it given breathing room.

src/app/components/elements/layout/SidebarLayout.tsx
```diff
@@ -512,10 +514,108 @@ export const SignupSidebar = ({activeTab} : {activeTab: number}) => {
...
+            <h5>Event type</h5>
+            <ul>               
+                {Object.entries(EventStatusFilter)
+                    .filter(([_statusLabel, statusValue]) => (user && user.loggedIn) || statusValue !== EventStatusFilter["My booked events"])
+                    .filter(([_statusLabel, statusValue]) => (user && user.loggedIn && isTeacherOrAbove(user)) || statusValue !== EventStatusFilter["My event reservations"])
+                    .map(([statusLabel, statusValue]) =>
+                        <li className="list-unstyled" key={statusValue}>
+                            <Label className="label-radio d-flex">
+                                <Input                                   
+                                    id={statusValue}
```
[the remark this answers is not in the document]
axlewin: Hmm - I agree this could be using the space better, but I think it's fine to increase the actual spacing between components. In fact, the designs have the filters taking up the entire page height. I've added some padding (on the outer `Label` components, to keep the checkboxes in line with their labels). I still think having the inputs slightly closer together than in the designs looks better, but let me know what you think.

I do prefer the `FilterCheckbox` styling (a lot!), and it would be nice to use it for consistency, but I'm not sure it works here. I don't think multi-select is particularly helpful for anything other than `Stage`, and it raises questions about combining filters (e.g. if I select "Student" and "Online" events, do we AND or OR them?).

sjd210 (changes requested):
I think this is most of the way there now. I have a couple points about adding some on-hover css behaviour to some buttons (which we're pretty consistent about wanting in the rest of the redesign), and then I also have some thoughts about the image/map section of the page.

The designs don't have a layout for small screens on this page, but we shouldn't really be allowing the image/map to take up so much space on the RHS of the screen even when further down on the page than they're displayed. It's fine like that on large screens, but we ought to move them on small screens.

I've got a [branch](https://github.com/isaacphysics/isaac-react-app/pull/1345) where I did this myself, and while at it re-combined the Ada/Phy Event Details containers since they became barely any different anyway and it felt silly keeping 100s more lines of codes for a separate page that isn't even used.

sjd210 (approved):

PR #1336, opened 2025-03-03
Redesign Set Tests page

axlewin:


axlewin:
The "Manage Tests" table still doesn't look great, especially on smaller screens. So far I've prioritised keeping the existing `<tr>` layout to minimise the site-specific code needed here, but this might look better with a more fundamental change to use a `ListView` or similar.

PR #1339, opened 2025-03-07
Redesign My Tests page

axlewin:
The labels & placeholders for filters on this page still need to be made consistent with those on other quiz/assignment pages, but this will be easier to do separately once all of the relevant pages are done.


src/app/components/pages/quizzes/MyQuizzes.tsx
```diff
@@ -315,43 +404,10 @@ const MyQuizzesPageComponent = ({user}: QuizzesPageProps) => {
...
-        <span>Past tests</span>
-        <div className="h-100 align-content-center">
-            <StyledToggle
-                checked={showCompleted}
-                falseLabel="Hidden"
-                trueLabel="Shown"
-                onChange={() => {
-                    const target = !showCompleted;
-                    setShowCompleted(target);
-                    setQuizStatuses(s => target ? [...s, QuizStatus.Complete, QuizStatus.Overdue] : s.filter(status => ![QuizStatus.Complete, QuizStatus.Overdue].includes(status)));
```
jacbn, line 346: I will be quite sad to lose this, it sped up working with the filters quite significantly. I wonder if instead we could have two more "broader" filters in the same group as "All" for both "Active" and "Past", and put the rest of the tabs in a dropdown or make them smaller if you need the more fine-grained filtering. Don't worry about it for now though, this works well enough as is; I'll make a card to remind us to take another look when we've completed the bulk of the redesign.

jacbn (commented):
I have made a couple of changes to make a couple of things clearer that applied to both the old and new designs, but I like these new ones a lot. Made a couple of suggestions too for things to improve if you'd like.

src/app/components/pages/quizzes/MyQuizzes.tsx
axlewin: I agree that this is awkward to use at the moment,  "Active" and "Past" would definitely be useful. I wonder if the finer-grained filtering is needed at all; having a dropdown as well as the new-style filters might just be confusing.

PR #1341, opened 2025-03-10
Previous page context overhaul

jacbn:
This is dependent on the first few commits of the concept page redesign, which isn't finished yet; there will be some harmless but unrelated changes in `Concepts.tsx` if this is reviewed first (I wouldn't wait on the concepts PR though, this is otherwise entirely separate).

---

In the redesigns, there is an idea of a "previous context", whereby pages have some recognition of how the user got to this point. This is primarily used for the breadcrumb, on pages where the URL is not the ground truth for the page context (e.g. question pages). On these, we would like a link back to the subject/stage landing page if that was where we came from, so that the user journey is more clear.

We managed this before by simply not updating the context between page switches; if you were previously on the A Level Chem landing page, this would be the initial context on the next page load, and this could be compared with the `doc` properties if loading a e.g. question page to see whether this was still valid or whether it should change. However, a problem with this approach is that we are relying on the page we are visiting to continue managing the context. This would mean needing context-related code on *all* pages to work properly; otherwise, we would have (as we do currently) a problem where context is not updated on e.g. the homepage (which should be neutral, but does not have any code to "reset" the context), which causes e.g. the navbar at the top, which itself uses the context as a source of truth, to be incorrectly coloured.

The new approach proposed in this PR provides several changes to this method:
- All `pageContext` management should come from a hook defined in `pageContext.ts`. As of this PR, there are two; `useUrlPageTheme` for pages where the URL is the source of truth, and `usePreviousPageContext` for pages where the previous page's context should be maintained if possible.
- All hooks must manage both setup **and teardown** of the context on the relevant page (see the return statements in the `useEffect`s inside these functions). That is, `pageContext`'s `subject` and `stage` fields must both always be reset on leaving a page. This ensures we do not need any context management code on non-question, non-concept pages, which as we have a lot of hard-coded generic pages, is extremely useful.
- This necessarily required the addition of a new field, `previousContext`, to `pageContext`. This should always store the context that was reset during teardown. This can then be reused in e.g. `usePreviousPageContext` to check if this context is still valid.
- This simplification has also removed the need for the `resetIfNotFound` parameter on `useUrlPageTheme`, as the context is now always between page transitions.


src/test/pages/__image_snapshots__/phy/Assignment schedule should have no visual regressions #0.png
axlewin, line 1: There seems to be some strange git behaviour around this file...I have a feeling that restoring it now might just cause more issues, so maybe removing it here and replacing it once we update all the phy VRTs is fine.

axlewin (approved):

PR #1349, opened 2025-03-14
Bug/redesign 2024/subject landing page infinite requests

barna-isaac:
Turned out, the bug was only present when no user was logged in. I cannot explain why. On the upside, the react-testing-toolkit tests did reproduce the issue after all, they just needed to be adjusted so the user is logged out.

### DIFF
--- a/src/app/components/pages/SubjectLandingPage.tsx
+++ b/src/app/components/pages/SubjectLandingPage.tsx
@@ -25,15 +25,18 @@ const RandomQuestionBanner = ({context}: {context?: PageContextState}) => {
     const deviceSize = useDeviceSize();
     const dispatch = useAppDispatch();
 
+    const subjects = context?.subject;
+    const stages = context?.stage?.map(s => s === "11_14" ? "year_7_and_8,year_9" : s).join(',')
+
     const searchDebounce = useCallback(debounce(() => {
         dispatch(searchQuestions({
             searchString: "",
             tags: "",
             fields: undefined,
-            subjects: context?.subject,
+            subjects,
             topics: undefined,
             books: undefined,
-            stages: context?.stage?.map(s => s === "11_14" ? "year_7_and_8,year_9" : s).join(','),
+            stages,
             difficulties: undefined,
             examBoards: undefined,
             questionCategories: "problem_solving,book",
@@ -42,7 +45,7 @@ const RandomQuestionBanner = ({context}: {context?: PageContextState}) => {
             startIndex: undefined,
             limit: 1
         }));
-    }), [dispatch, context]);
+    }), [dispatch, subjects, stages]);
 
     const {results: questions} = useAppSelector((state) => state && state.questionSearchResult) || {};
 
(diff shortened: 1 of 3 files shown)


src/app/components/pages/SubjectLandingPage.tsx
```diff
@@ -42,7 +45,7 @@ const RandomQuestionBanner = ({context}: {context?: PageContextState}) => {
             startIndex: undefined,
             limit: 1
         }));
-    }), [dispatch, context]);
+    }), [dispatch, subjects, stages]);
```
jacbn, line 48: So, the issue is that `stages` is an array in the pageContext, which is regenerated on every retrieval because the hook that does this (`useUrlContext()`) recreates a **new array** for the context every time. This causes `searchDebounce` to rerender on every retrieval of the context because the two arrays are not equal; at least, they have the same contents, but they are not the same array. This solution maps the array into a string, for which shallow equality is enough and so this will not rerender (though note the ""preferred"" approach for doing this seems to be to use `JSON.stringify` in the deps array), but IMO this doesn't fix the underlying issue and we will run into this again if we want the context in a dependency array.

I believe a better fix would be to modify the hook so that it always returns the same array in the first place. We can do this by replacing the return statement of `useUrlPageTheme` (`pageContext.ts` line 178),
```ts
return determinePageContextFromUrl(location.pathname);
```
with
```ts
return useAppSelector(selectors.pageContext.context);
```
This creates one more rerender cycle where the data is stale, but we should be able to detect this and display a "loading..." div as we do elsewhere. I just think this approach is less prone to further rerender loops elsewhere in the code which may not be detectable (this one at least showed us a network request every time) and allows safely using the context in deps arrays, *as we are already doing elsewhere*.

mwtrew (approved):
Nice catch - `context` seems to be changing and triggering the re-render, but not the properties we're interested in. Good to see some tests too.

Now that's fixed, a more general question: do we need the `useCallback` and `debounce`? It's not clear to me why it would need to be debounced. I'm wondering if we could simplify it further and move the logic straight to the `useEffect`, e.g.:

```javascript
    useEffect(() => {
        dispatch(searchQuestions({
            searchString: "",
            tags: "",
            fields: undefined,
            subjects,
            topics: undefined,
            books: undefined,
            stages,
            difficulties: undefined,
            examBoards: undefined,
            questionCategories: "problem_solving,book",
            statuses: undefined,
            fasttrack: false,
            startIndex: undefined,
            limit: 1
        }));
    }, [dispatch, subjects, stages]);
```

It's worth asking @jacbn if there's a reason to keep it the way it is - it could be related to the "Get a different question" button which is not yet implemented.

src/test/pages/SubjectLandingPage.test.tsx
```diff
@@ -0,0 +1,55 @@
+import {screen, waitFor} from "@testing-library/react";
+import { renderTestEnvironment} from "../testUtils";
+import { mockQuestionFinderResults } from "../../mocks/data";
+import userEvent from "@testing-library/user-event";
+import { http, HttpResponse } from "msw";
+import { API_PATH, isAda } from "../../app/services";
+import { UserRole } from "../../IsaacApiTypes";
+
+describe("SubjectLandingPage", () => {
+    jest.setTimeout(10000);
```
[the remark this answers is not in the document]
barna-isaac: Good catch, removed.

src/app/components/pages/SubjectLandingPage.tsx
barna-isaac: Okay, I agree it'd be nice not to have this bug come back. I sent in a new commit with this solution.

PR #1352, opened 2025-03-17
Subject glossaries

jacbn:
Adds glossary pages for individual subject/stage pairs. Removes the superfluous URL params on these pages (while maintaining them for the main /glossary page).

Would appreciate some thoughts on:
- whether the sidebar layout works, as the glossary is (intentionally!) quite a long page so the sidebar leaves blank space for most of it;
  - (perhaps we could solve this by using a fixed-height scrollable window for the glossary?)
- whether the darker background for the results works, needs a border, etc. I'm trying to keep it consistent with elsewhere in the designs, but I don't love it.


axlewin (approved):
Looks good - I have a couple of general glossary-related points while we're looking at them, but they shouldn't block this being merged.

RE your comments:
- I think the sidebar layout is fine. A scrollable window might just make the page harder to use, and would be inconsistent with other potentially long pages (consider e.g. My Tests/Assignments when there are lots of assignment cards).
- I've tried a few options for borders etc on the results, and I think it looks best as-is. If anything, having the results a clearly distinct colour from the sidebar helps to make the empty sidebar spacing further down look better.

PR #1356, opened 2025-03-21
Redesign: assignment improvements

axlewin:
This PR ended up with more small miscellaneous changes than it should have done, but the three main changes are:
- Allow sorting on My Assignments
- Add a sidebar to gameboard overview pages
- Add a progress sidebar to question pages within a gameboard

Particularly controversial decisions that might need further discussion:
- Whether all of the assignment sort options are necessary, and what the exact wording for them should be
- What gameboard metadata should be in the "Question deck details" section of the sidebar
- Whether the styling on the progress sidebar works


jacbn (commented):
This looks really good! A couple of spacing fixes but only minor. As for your questions:
1. More options for sorting isn't going to do any harm, I'd rather all the date ones used the same wording but it's clear regardless. I might bring it up briefly with content but you can leave it for now.
2. Subjects and topics are enough metadata imo. I worry that big gameboards might have too many topics and it would look weird with all of them being pills; equally, it looks really good for smaller gameboards, so I say we keep it for now and see what content think.
3. I think we should remove these. They wrap weirdly atm, especially on mobile, and the information is already conveyed by the icon on the left and the green background for correct. Though, perhaps we should add the [correctness key](https://github.com/isaacphysics/isaac-react-app/blob/a6b2b23aeaaf51b68aacbbf3725e1e8f395b436a/src/app/components/elements/layout/SidebarLayout.tsx#L185-L190) in a new section underneath the rest of the sidebar? (note there are 4 states now!)

PR #1368, opened 2025-03-27
Question finder fixes

jacbn:
Fixes some context bugs with the question finder, then refactors a significant portion of the old code, including:
- moving tag hierarchy code to its own location
- ensuring questions are displayed for the correct QF on load
- remove unused code, including the old hex filter panel
- replace debounce functions' `useCallback` with `useMemo` to correctly return the _result_ from the call to debounce, rather than the debounce function
- rename various variables for clarity

---

One thing I know to be broken: on Ada, the shuffle button also applies filters. Fixing this requires separating out the shuffle logic to use its own `history.replace` call that only updates the `randomSeed`, as opposed to relying on `searchAndUpdateUrl`. This is a fairly small thing so I'm leaving it for now as that's a fair amount of work, but I've made a note.


jacbn:
@barna-isaac – I disagree with the failing test. IMO, switching filters **on Ada** should not reset the random seed until Apply Filters is clicked. This makes it more consistent with how otherwise changing the filters does not cause the question list to update. Any thoughts?

barna-isaac:
@jacbn and I looked at this, and have found that the tests already expect the behaviour suggested by @jacbn, it's just that the current implementation does not behave this way.

PR #1370, opened 2025-03-27
Add account initials icon

axlewin:


sjd210:
The fact this is only two letter characters makes it (mostly) silliness-proof, and I think it's absolutely fine that we're just filtering for the first character of each name - if someone is including unusual characters, that's on them.

The silliest things I could find are things like ඞ or ツ, which are ultimately fine and people could've been using in their names anyway if they wanted (even if the amogus does pain me a bit personally).

PR #1371, opened 2025-03-28
Implement initial design of book pages

jacbn:
Implements the result of many discussions with the content team about what pieces of the designs worked for books and what didn't. This is still somewhat incomplete (tabs styling should be dropdowns, missing assignment links for teachers) and cards have been made for this improvement, but it would be good to get the initial design through the door.


src/app/components/elements/Book.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    useEffect(() => {
+        const hash = history.location.hash;
+
+        if (!book?.id || !hash) {
+            setPageId(undefined);
+            return;
+        }
+
+        const fragmentId = book?.id + "_" + hash.replace("#", "");
```
jsharkey13, line 29: I think we should use full paths here (i.e. `/books/physics_skills_19/a1`) in case we later want to link to places _inside_ these section pages, and because these section pages really are full pages of their own.

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -119,18 +132,10 @@ export const RoutesPhy = [
...
-    <TrackedRoute key={key++} exact path="/books/pre_uni_maths" component={PreUniMaths}/>,
-    <TrackedRoute key={key++} exact path="/books/pre_uni_maths_2e" component={PreUniMaths2e}/>,
-    <TrackedRoute key={key++} exact path="/books/chemistry_16" component={Chemistry16}/>,
-    <TrackedRoute key={key++} exact path="/books/quantum_mechanics_primer" component={QuantumMechanicsPrimer}/>,
-    <TrackedRoute key={key++} exact path="/books/solve_physics_problems" component={SolvingPhysProblems}/>,
-    <TrackedRoute key={key++} exact path="/books/maths_book_gcse" component={MathsBookGcse}/>,
-    <TrackedRoute key={key++} exact path="/books/phys_book_yr9" component={PhysBookYrNine}/>,
-    <TrackedRoute key={key++} exact path="/books/step_up_phys" component={StepUpPhys}/>,
-    <TrackedRoute key={key++} exact path="/books/linking_concepts" component={LinkingConcepts}/>,
+    ...(Object.entries(books).map(([path, component]) => [
```
jsharkey13, line 135: This hardcoding sort of defeats the point of books as content-only objects. Can we not make a route, _after_ the existing hard-coded book routes, that just renders a book page based solely on the URL?

I feel like the `Book` component should be able to do that without the page ID being passed in?

jsharkey13: (It needs to be after the existing hardcoded ones so that we can migrate them without having to do it all at once; any _new_ book ID will not match the earlier hardcoded paths and will use the new style, but existing ones will keep using the existing model. Then we delete the hardcoded part once the book is migrated!)

sjd210 (changes requested):
I'm glad these book pages are finally getting somewhere with all the back-and-forth - and I think this looks good 😊! There are a couple of small things for changes local to here, then as you say we can move onto specific improvements in another card.

src/app/components/elements/layout/SidebarLayout.tsx
```diff
@@ -1202,6 +1205,54 @@ export const GlossarySidebar = (props: GlossarySidebarProps) => {
...
 
+export interface BookSidebarProps extends SidebarProps {
+    book: IsaacBookIndexPageDTO;
+    urlBookId: string;
+    pageId: string | undefined;
+};
+
+export const BookSidebar = ({ book, urlBookId, pageId }: BookSidebarProps) => {
+
+    const [expandedTab, setExpandedTab] = useState<number | undefined>(undefined);
```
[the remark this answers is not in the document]
jacbn: Oh, this is because we switched from storing page ids in hash anchors to hierarchy paths on Friday. Ugh.

Not sure it really matters, but calculating `activeChapter` inside the `useEffect` feels more React-y. We don't end up needing to calculate it on every render and the _actual_ dependencies are in the `useEffect`'s dep array this way.

```suggestion
    const [expandedTab, setExpandedTab] = useState<number | undefined>(undefined);
    useEffect(() => {
        const activeChapter = book.chapters?.map(chapter => chapter.sections?.some(section => section.bookPageId === pageId)).indexOf(true);
        setExpandedTab(activeChapter === -1 ? undefined : activeChapter);
    }, [book.chapters, pageId]);
```

jacbn:
(the build failed due to problems with npm? cloudflare?, it seems. i'll try rerunning in the morning)

I might look at stopping the constant scroll-to-tops before we merge this too. It gets quite annoying if you're scanning several pages...

jacbn:
I've made it scroll to the top of the title instead of all the way up, which is definitely less jarring. Do you think we should:
1. not scroll at all?
2. scroll only in non-reduced motion?
3. leave as is?

sjd210:
Option 2! With reduced motion enabled in the browser, `scrollTo` (and by extension `scrollVerticallyIntoView`) will abandon smooth scrolling and just snap immediately to that page position - this is very annoying. Otherwise this scrolling to top is a nice improvement.

jacbn:
Okay. I've refactored the scrolling into a hook so we can use the accessibility work from before. While testing this, I've noticed there is inconsistent behaviour on questions, concepts etc where on load, sometimes it scrolls to the top (i.e. (0, 0)), and others to the `h1#main-heading`, but this behaviour also happens on the live site. It appears to be caused by a race condition between [this](https://github.com/isaacphysics/isaac-react-app/blob/ff19924fdd2088aa716746f60cb68510d064afbc/src/app/components/elements/PageTitle.tsx#L98), and the scroll-to-top modified in this PR; we could modify the former to `element.focus({preventScroll: true});` to make it consistent everywhere. I think I'm going to leave this for now though as it is related to a change we need to make regarding where the "skip to main content" accessibility button points on different pages; just making you aware.

sjd210:
Quick note since this page is touching it: keyboard navigation is technically usable but really very messy with `CollapsibleList`s in general. Outlines are (for some reason) universally disabled on buttons such as these list headers, collapsed sections are in the tab order despite not being able to see whats selected, and hiding the overflow hides some of the focus outline.

This has already been this broken on the live site for quite a while and looking at `focus.scss`, I think we should be due a focus overhaul as part of this redesign process anyway. It's certainly unrelated enough to anything here that this can be merged regardless - just a note for if it wasn't already known.

sjd210 (approved):

PR #1372, opened 2025-03-31
Re-style reduced-motion into its own mixin

sjd210:
```css
@at-root main[data-reduced-motion="true"] & { ...
```
This was a neat way I found to read the attribute at root while applying rules at the current scope. The rest should be pretty standard Sass.

### DIFF
--- a/src/scss/common/accessibility.scss
+++ b/src/scss/common/accessibility.scss
@@ -1,8 +1,26 @@
-$reduced-motion: false;
+@mixin reduced-motion($show-content: true) {
+  // default values for all accessibility variables
+  --focus-outline: 2px solid #000; // TODO use this
 
-@mixin not-reduced-motion {
-  @if not $reduced-motion {
-    @content;
+  // if reduced motion is enabled in the browser or the app, apply these reduced-motion styles e.g. instant transitions
+  @media (prefers-reduced-motion: reduce) {
+    @if $show-content {
+      @content;
+    }
+  }
+  @at-root main[data-reduced-motion="true"] & {
+    @if $show-content {
+      @content;
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


jacbn:
I like this (because it works :sunglasses:) but I do think the readability has gone down a bit. `@include reduced-motion {}` is for things you want with reduced motion (fine), but `@include reduced-motion($show-content: false)` is for things without; you also end up having to check `$show-content` repeatedly. You basically have two different codepaths depending on the parameter, so why not just have two mixins?

```scss
@mixin reduced-motion {
  // if reduced motion is enabled in the browser or the app, apply these reduced-motion styles e.g. instant transitions
  @media (prefers-reduced-motion: reduce) {
    @content;
  }

  @at-root main[data-reduced-motion="true"] & {
    @content;
  }
}

@mixin not-reduced-motion {
  // if reduced motion is not enabled in the browser or the app, don't apply these motion styles e.g. hover scale effects
  @media (prefers-reduced-motion: no-preference) {
    @at-root main[data-reduced-motion="false"] & {
      @content;
    }
  }
}
```

We'd need to move the default variables to #main again, but IMO they belong here anyway; this is where we set all the defaults across most of the app. It also prevents us duplicating definitions that don't change.

```scss
#main {
  // default values for all accessibility variables
  --focus-outline: 2px solid #000; // TODO use this
}
```

As for `.collapsing`, yes, good point. We can reuse the compliance mixin though:
```scss
.collapsing { // a bootstrap class used in accordions
  @include reduced-motion-compliant-animation;
}
```

As a final note, sorry for making you go through this. I have twice now tested it with accordions and assumed it worked for everything else. :pensive:. At least we learnt something :)

sjd210:
Happy with your suggestions :) Splitting it into two different mixins makes good sense to me for readability

PR #1373, opened 2025-04-01
Overhaul redesign iconography

jacbn:
Previously, reusing an icon at a different scale required making a new CSS class with the same fg/bg images, but with a different size. Since several classes were representing the same image, the naming convention was not obvious, and changes to one icon need to be propagated to all classes.

This proposes a change to the structure. There is now a 1:1 mapping of SVG fg/bg pairs to `icon-xxx` classes, but the size parameter is now given by a CSS variable (default 16px). These can be modified by passing an additional class alongside the icon, `icon-lg` (60px, for ALVIs) or `icon-xl` (75px, for display cards on landing pages). Since all small icons had a mask size of 100%, but large icons had icon-specific sizes, this is managed via a similarly new variable `--mask-size`, which if *not* set falls back to the icon-specific value, but is set by default via `.icon` to 100% and unset through `icon-lg` and `icon-xl`.

The result is a greatly reduced number of icon classes, and each is given a more pleasant name (`icon-concept`! `icon-question`!). For icons with a thicker SVG variant (intended for inline icons), a new `-thick` icon naming convention exists (e.g. `icon-question-thick`).


axlewin (approved):
Nice, this is a lot neater now 👍

PR #1374, opened 2025-04-01
Fix assignment modal bug

axlewin:
Context: After creating a new gameboard and clicking "set as assignment", the assignment setting modal was not opening.

Explanation: When "set as assignment" is clicked, the ID of the new gameboard is stored in a hash anchor. The modal then works by taking the first 6 gameboards from the `useGameboards` hook and checking their IDs against the hash anchor to determine which gameboard to use. This previously worked because `useGameboards` returned gameboards sorted most-recent-first, so the just-created gameboard is always among the 6 returned. A previous change on the redesign branch reversed the order in which gameboards are returned, so the ID in the hash anchor was not found if the user had > 6 gameboards and no modal appeared.

The solution proposed here is just to revert the sorting changes to how it works on master. If we want to keep the new sorting logic, we could alternatively remove the 6-gameboard limit and search all of a user's gameboards for the hash anchor ID, but for users with many gameboards this would be lots of extra requests.

### DIFF
--- a/src/app/services/gameboards.tsx
+++ b/src/app/services/gameboards.tsx
@@ -205,10 +205,10 @@ export enum BoardLimit {
 }
 
 export const BOARD_ORDER_NAMES: {[key in AssignmentBoardOrder]: string} = {
-    "created": "Date created (oldest first)",
-    "-created": "Date created (recent first)",
-    "visited": "Date visited (oldest first)",
-    "-visited": "Date visited (recent first)",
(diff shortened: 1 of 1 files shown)


sjd210:
👍 Makes sense for it to be sorting by most recent by default anyway. 

Had an issue with the Docker build not running due to CamelGate, but re-running seems to have fixed that so we should be fine now.

PR #1375, opened 2025-04-02
Improvement/autofill assignment due

barna-isaac:
On the `Set assignment` modal, the due date now has a default value, which is set to 6 days from now. This way, an assignment set on Monday will be due by the end of Sunday.

I've included new notice text on the modal to clarify when during the day an assignment becomes available and when it's due. I've also moved the notice about due date now being a required field. It is now a validation message that we only show when the due date is empty.

There is a different version of this page, available through the `Assignment schedule` page, just for Physics. That page populates the fields from an existing assignment. I did not overwrite these fields with the default, but introduced the default on this page also for when an assignment is successfully set and the option is presented to set a second assignment with defaults. This page has no user guidance and I did not introduce that text here. However, I did convert the warning about due date now being required to a validation error on this page as well.

### DIFF
--- a/config/jest/jest.config.common.js
+++ b/config/jest/jest.config.common.js
@@ -4,6 +4,7 @@ module.exports = {
         "!src/**/*.d.ts"
     ],
     "resolver": "jest-pnp-resolver",
+    globalSetup: "<rootDir>/src/test/globalSetup.ts",
     setupFiles: [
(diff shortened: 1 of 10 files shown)


barna-isaac:
Thanks for the review! I don't have a take on `Due date reminder` vs `Due date`, but vaguely recall @jsharkey13 being in favour of the first? At any rate, I'll go ahead with the merge on both branches.

PR #1380, opened 2025-04-03
Make assignments/tests sidebars consistent with each other

axlewin:
Minor changes to ensure all assignments/tests-related sidebars use similar styling.

### DIFF
--- a/src/app/components/elements/layout/SidebarLayout.tsx
+++ b/src/app/components/elements/layout/SidebarLayout.tsx
@@ -682,27 +682,16 @@ export const MyGameboardsSidebar = (props: MyGameboardsSidebarProps) => {
     const deviceSize = useDeviceSize();
 
     return <ContentSidebar {...rest} className={classNames(rest.className, "pt-0")}>
-        <div className="section-divider"/>
+        {above["lg"](deviceSize) && <div className="section-divider"/>}
         <h5>Search gameboards</h5>
         <Input
-            className='search--filter-input my-4'
+            className='search--filter-input my-3'
             type="search" value={boardTitleFilter || ""}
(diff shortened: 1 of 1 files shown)


barna-isaac:
Here's a before-after comparison:
| Page       | Before | After |
| ---         | ---      |          --- |
| `/tests` | tests_before|tests_after |
| git diff     |






-

src/app/components/elements/layout/SidebarLayout.tsx
```diff
@@ -1009,15 +1007,6 @@ export const ManageQuizzesSidebar = (props: ManageQuizzesSidebarProps) => {
             value={quizDueDate && !isNaN(quizDueDate.valueOf()) ? formatISODateOnly(quizDueDate) : undefined} onChange={event => setQuizDueDate(new Date(event.target.value))}
             placeholder="Filter by due date" aria-label="Filter by due date"
         />
-    </div>;
```
barna-isaac, line 1012: I think the original version nicely highlighted that the `ManageQuizzesSidebar` consisted of a search, a `titleFilter`, a `groupFilter`, a `setDateFilter` and a `dueDateFilter`. Not a big deal though, I'm fine with the new version also.

axlewin: It's a good point, and something I thought about. Perhaps at some point it would be nice to refactor all of the input-based sidebars to use individual consts for each input (which could also facilitate some re-use of the same inputs), but since none of the others are structured like this, I prefer it this way for consistency for now.

axlewin:
@barna-isaac I'll make a Trello card for the glitches you mentioned - otherwise I'm happy for this to be merged

PR #1381, opened 2025-04-04
Redesign: rename gameboards to question decks

axlewin:


axlewin:
The changes in the first commit should be uncontroversial. The subsequent commits have more potential to break things - in particular, `/gameboards#id` doesn't redirect anywhere.

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -179,7 +179,8 @@ export const RoutesPhy = [
 
     // Legacy Routes
     <Redirect key={key++} exact from="/mission" to="/about" />,
-    <Redirect key={key++} exact from="/boards" to="/my_gameboards" />,
+    <Redirect key={key++} exact from="/boards" to={PATHS.MY_GAMEBOARDS} />,
+    <Redirect key={key++} exact from="/my_gameboards" to={PATHS.MY_GAMEBOARDS} />,
```
jsharkey13, line 183: We probably need a redirect for `add_gameboard` here too, since that is a very well-used route!

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -179,8 +179,10 @@ export const RoutesPhy = [
...
     <Redirect key={key++} exact from="/game_builder" to={PATHS.GAMEBOARD_BUILDER} />,
+    <Redirect key={key++} exact from="/gameboard_builder" to={PATHS.GAMEBOARD_BUILDER} />,
     <Redirect key={key++} exact from="/board/:id" to={`${PATHS.GAMEBOARD}#:id`} />,
```
jsharkey13, line 186: I think much like this redirect, we _do_ need one for `/gameboard#id` too. That way, later on, we can redirect `isaacphysics.org/anything` to `isaacscience.org/anything` without having to worry about what the path is.

src/app/state/slices/api/gameboards.ts
```diff
@@ -159,13 +159,13 @@ export const unlinkUserFromGameboard = createAsyncThunk<string, {boardId?: strin
                 const hasAssignedGroups = (assignmentsByMe?.filter(a => a.gameboardId === boardId) ?? []).length > 0;
                 if (hasAssignedGroups) {
                     if (reduxState && reduxState.user && reduxState.user.loggedIn && isAdminOrEventManager(reduxState.user)) {
-                        if (!confirm(`Warning: You currently have groups assigned to ${boardTitle}. If you delete this your groups will still be assigned but you won't be able to unassign them or see the ${siteSpecific("gameboard", "quiz")} in your assigned gameboards or 'My gameboards' page.`)) {
+                        if (!confirm(`Warning: You currently have groups assigned to ${boardTitle}. If you delete this your groups will still be assigned but you won't be able to unassign them or see the ${siteSpecific("question deck", "quiz")} in your assigned ${siteSpecific("question decks", "quizzes")} or ${siteSpecific("My question decks", "Manage assignments")} page.`)) {
```
jsharkey13, line 162: The final `${siteSpecific("My question decks", "Manage assignments")}` seems wrong. Those are two different pages?
Should it be "Set assignments" instead of "My question decks"? It is true that it will not be on "My question decks" either, but if for Ada you think the assignments page is the more useful one to refer to (and I don't disagree!) then I think it should be consistent!

```suggestion
                        if (!confirm(`Warning: You currently have groups assigned to ${boardTitle}. If you delete this your groups will still be assigned but you won't be able to unassign them or see the ${siteSpecific("question deck", "quiz")} in your assigned ${siteSpecific("question decks", "quizzes")} or ${siteSpecific("Set assignments", "Manage assignments")} page.`)) {
```

jsharkey13: Only staff can do this anyhow, so maybe the wording doesn't matter _much_; but just in case it ends up used more widely.

src/app/services/searchResults.ts
```diff
@@ -178,10 +178,10 @@ const siteShortcuts: SearchShortcut[] = siteSpecific([
         type: SEARCH_RESULT_TYPE.SHORTCUT
     }, {
         id: "my_gameboards",
-        title: "My Gameboards",
-        terms: ["gameboards", "gameboard", "my gameboards"],
-        summary: "View your saved gameboards.",
-        url: "/my_gameboards",
+        title: "My Question Decks",
+        terms: ["gameboards", "gameboard", "my gameboards", "question deck", "question decks", "my question decks"],
```
jsharkey13, line 182: Maybe we could add "boards" and "decks" too?
```suggestion
        terms: ["gameboards", "gameboard", "my gameboards", "boards", "question deck", "question decks", "my question decks", "decks"],
```

src/app/components/elements/cards/BoardCard.tsx
```diff
@@ -159,9 +159,9 @@ export const BoardCard = ({user, board, boardView, assignees, toggleAssignModal,
     function confirmDeleteBoard() {
         if (hasAssignedGroups) {
             if (isAdminOrEventManager(user)) {
-                alert(`Warning: You currently have groups assigned to this gameboard. If you delete this your groups will still be assigned but you won't be able to unassign them or see the ${siteSpecific("gameboard", "quiz")} in your assigned ${siteSpecific("gameboards", "quizzes")} or '${siteSpecific("My gameboards", "My quizzes")}' page.`);
+                alert(`Warning: You currently have groups assigned to this ${siteSpecific("question deck", "quiz")}. If you delete this your groups will still be assigned but you won't be able to unassign them or see the ${siteSpecific("question deck", "quiz")} in your assigned ${siteSpecific("question decks", "quizzes")} or '${siteSpecific("My question decks", "My quizzes")}' page.`);
```
jsharkey13, line 162: This is the same message as the other place, but ends differently. Can we be consistent?
Is there potential for code re-use, even, given the duplicated logic?

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -179,9 +179,13 @@ export const RoutesPhy = [
...
     // Legacy Routes
     <Redirect key={key++} exact from="/mission" to="/about" />,
-    <Redirect key={key++} exact from="/boards" to="/my_gameboards" />,
+    <Redirect key={key++} exact from="/boards" to={PATHS.MY_GAMEBOARDS} />,
+    <Redirect key={key++} exact from="/my_gameboards" to={PATHS.MY_GAMEBOARDS} />,
     <Redirect key={key++} exact from="/game_builder" to={PATHS.GAMEBOARD_BUILDER} />,
+    <Redirect key={key++} exact from="/gameboard_builder" to={PATHS.GAMEBOARD_BUILDER} />,
+    <Redirect key={key++} exact from="/add_gameboard/:id" to={`${PATHS.ADD_GAMEBOARD}/:id`} />,
     <Redirect key={key++} exact from="/board/:id" to={`${PATHS.GAMEBOARD}#:id`} />,
+    <Route key={key++} exact path="/gameboards" render={(props) => <Redirect to={`${PATHS.GAMEBOARD}${props.location.hash ?? ""}`}/>} />, // Route wrapper needed to handle hash
```
jsharkey13, line 188: I think this is _slightly_ cleaner, but I understand that you liked neither approach!
```suggestion
    <Redirect key={key++} exact from="/gameboards" to={{pathname: PATHS.GAMEBOARD, hash: window.location.hash}} />,
```

src/app/state/slices/api/gameboards.ts
axlewin: Hmm. I had assumed that `your assigned ${siteSpecific("question decks", "quizzes")}` referred to the Set/Manage Assignments page. I went with "Manage assignments" for Ada because  `/quizzes` doesn't seem to be accessible for teachers via any of the menus (unless I'm missing it somewhere), but this does introduce some redundancy for Ada.

Assuming that "your assigned quizzes" and "Manage assignments" are the same thing, we could just remove the last part of the sentence entirely for Ada?

jsharkey13 (approved):

PR #1391, opened 2025-04-10
Convert to slices – page fetch

jacbn:
Continuing a big piece of tech debt we inherited, this continues the move to slices from action/reducers. This was brought on since this automatically fixes issues with newer requests storing stale state before completion.

A previous iteration of this PR did not account for other components checking the page type through redux, so a new `docSlice` was built to allow this functionality.

### DIFF
--- a/src/app/components/handlers/ShowLoadingQuery.tsx
+++ b/src/app/components/handlers/ShowLoadingQuery.tsx
@@ -28,6 +28,7 @@ export const DefaultQueryError = ({error, title}: {error?: FetchBaseQueryError |
 interface ShowLoadingQueryInfo<T> {
     data?: T | NOT_FOUND_TYPE;
     isLoading: boolean;
+    isFetching: boolean;
     isError: boolean;
     error?: FetchBaseQueryError | SerializedError;
 }
@@ -36,6 +37,7 @@ export function combineQueries<T, R, S>(firstQuery: ShowLoadingQueryInfo<T>, sec
     return {
         data: isFound<T>(firstQuery.data) && isFound<R>(secondQuery.data) ? combineResult(firstQuery.data, secondQuery.data) : undefined,
         isLoading: firstQuery.isLoading || secondQuery.isLoading,
+        isFetching: firstQuery.isFetching || secondQuery.isFetching,
         isError: firstQuery.isError || secondQuery.isError,
         error: firstQuery.error ?? secondQuery.error,
     };
@@ -71,12 +73,12 @@ type ShowLoadingQueryProps<T> = ShowLoadingQueryErrorProps<T> & ({
 //  - `placeholder` (React element to show while loading)
 //  - `query` (the object returned by a RTKQ useQuery hook)
 export function ShowLoadingQuery<T>({query, thenRender, children, placeholder, ifError, ifNotFound, defaultErrorTitle}: ShowLoadingQueryProps<T>) {
-    const {data, isLoading, isError, error} = query;
+    const {data, isLoading, isFetching, isError, error} = query;
     const renderError = () => ifError ? <>{ifError(error)}</> : <DefaultQueryError error={error} title={defaultErrorTitle}/>;
     if (isError && error) {
         return "status" in error && typeof error.status === "number" && [NOT_FOUND, NO_CONTENT].includes(error.status) && ifNotFound ? <>{ifNotFound}</> : renderError();
     }
-    if (isLoading) {
+    if (isLoading || isFetching) {
         return placeholder ? <>{placeholder}</> : loadingPlaceholder;
     }
     return isDefined(data)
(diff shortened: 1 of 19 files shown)


src/app/components/handlers/ShowLoadingQuery.tsx
```diff
@@ -71,12 +73,12 @@ type ShowLoadingQueryProps<T> = ShowLoadingQueryErrorProps<T> & ({
...
 //  - `query` (the object returned by a RTKQ useQuery hook)
 export function ShowLoadingQuery<T>({query, thenRender, children, placeholder, ifError, ifNotFound, defaultErrorTitle}: ShowLoadingQueryProps<T>) {
-    const {data, isLoading, isError, error} = query;
+    const {data, isLoading, isFetching, isError, error} = query;
     const renderError = () => ifError ? <>{ifError(error)}</> : <DefaultQueryError error={error} title={defaultErrorTitle}/>;
     if (isError && error) {
         return "status" in error && typeof error.status === "number" && [NOT_FOUND, NO_CONTENT].includes(error.status) && ifNotFound ? <>{ifNotFound}</> : renderError();
     }
-    if (isLoading) {
+    if (isLoading || isFetching) {
```
jacbn, line 81: I'm struggling to believe this ever worked, since `isLoading` is only ever true on the first request; `isFetching` is true while any request is acive..? It would have been making any pages that use this display stale data momentarily, before the new data request had succeeded.

src/app/components/pages/Generic.tsx
```diff
@@ -60,38 +61,43 @@ export const Generic = withRouter(({pageIdOverride, match: {params}}: GenericPag
...
-                            <ShareLink linkUrl={`/pages/${doc.id}`}/>
-                        </div>
-                        <div className="question-actions mt-3 not-mobile">
-                            <PrintButton/>
+    return <ShowLoadingQuery 
+        query={pageQuery}
+        defaultErrorTitle="Unable to load page"
+        ifNotFound={<NotFound />}
+        thenRender={supertypedDoc => {
+            const doc = supertypedDoc as IsaacQuestionPageDTO & DocumentSubject;
```
jsharkey13, line 69: I know this isn't your change, but is there a reason this is still (a) necessary, and (b) an `IsaacQuestionPageDTO` and not a `SeguePageDTO`?

src/app/state/selectors.tsx
```diff
@@ -18,7 +18,10 @@ export const selectors = {
     },
 
     doc: {
-        get: (state: AppState) => state?.doc || null,
+        get: (state: AppState) => state?.page?.doc || null,
+        getWithout404: (state: AppState) => {
+            return state?.page?.doc === NOT_FOUND ? null : state?.page?.doc ?? null;
```
jsharkey13, line 23: Is this right? Above you replaced 
```
state?.doc && state.doc !== NOT_FOUND ? state.doc : undefined
```
with this `getWithout404`. That used `undefined` for the 404 case. Is `null` better than `undefined`?

src/app/components/handlers/ShowLoadingQuery.tsx
jsharkey13: I don't think it was. The group token data for groups is loaded this way, and it does not display stale data no matter how many times I switch groups with heavily throttled network speeds. Event detail pages likewise do not show stale data :/

src/app/state/selectors.tsx
jacbn: It doesn't matter which we use, as the only operations we do with the result of `getWithout404` are `if ({state}.page) {...}`, i.e. nothing where the behaviour between undefined and null is different. I was only trying to make this one consistent with `get`, and given [this](https://github.com/isaacphysics/isaac-react-app/blob/6f21c92be67c7fbad84e433b47369a1e8ebe4a89/src/test/state/reducers.test.tsx#L36), I believe null has been preferred in the past so I'm trying to keep it to that.

src/app/components/pages/Generic.tsx
jacbn: Nope. This isn't the first place I've seen the wrong type here (believe Concept pages were like this too), so presumably just a copy-paste error.

src/app/components/handlers/ShowLoadingQuery.tsx
jacbn: Not 100% sure, but I believe that both the groups token modal and event detail pages have a different component for each instance, so each time the query loads it is the "first time". The case it doesn't work is when a fixed component (e.g. a wrapper) has a query object with a parameter that can change without creating a new instance of this component. It seems like almost all of our use cases for `ShowLoadingQuery` we use a query with e.g. an ID parameter drawn from the content object specified by the path in the URL, which will never change and so the "first time"-only case is enough. Regardless, we should definitely be checking for `isFetching`.

mwtrew (approved):

PR #1392, opened 2025-04-11
Improve admin pages styling

axlewin:
- Remove borders around "Useful links" (both sites)
- Add borders to buttons (phy)
- Improve padding & alignment of buttons (both sites)


jsharkey13:
This seems to miss the "Find users" in the Admin event management section "Add users to booking". Was there any pattern to the buttons that we could search for to find other instances?

jsharkey13 (approved):

PR #1395, opened 2025-04-11
Adapt teacher dashboard for tutors

axlewin:


sjd210:
Seems reasonable :+1: I did notice some weird spacing in the teacher/tutor dashboard group layout for an account with no groups, but that's pre-existing and can be dealt with separately. I've just added it to the bug document.

PR #1399, opened 2025-04-14
Redesign: change dashboard due date logic

axlewin:
The homepage dashboards weren't showing assignments/quizzes due today because they were considered overdue. This was inconsistent with My Assignments, which doesn't consider an assignment overdue until midnight at the _end_ of the due date. Both places now use the same `isOverdue` function which maintains the current My Assignments behaviour.

(This also fixes some edge cases for the student dashboard where having assignments due today prevented the next-soonest assignments from being shown).

### DIFF
--- a/src/app/services/assignments.ts
+++ b/src/app/services/assignments.ts
@@ -15,13 +15,13 @@ function createAssignmentWithStartDate(assignment: AssignmentDTO): AssignmentDTO
     return {...assignment, startDate: assignmentStartDate};
 }
 
+const now = new Date();
+const midnightLastNight = new Date(now);
+midnightLastNight.setHours(0, 0, 0, 0);
+
 type AssignmentStatus = "inProgressRecent" | "inProgressOld" | "allAttempted" | "allCorrect";
 export const filterAssignmentsByStatus = (assignments: AssignmentDTO[] | undefined | null) => {
-    const now = new Date();
     const fourWeeksAgo = new Date(now.valueOf() - (4 * 7 * 24 * 60 * 60 * 1000));
-    // Midnight last night:
-    const midnightLastNight = new Date(now);
-    midnightLastNight.setHours(0, 0, 0, 0);
 
     const myAssignments: Record<AssignmentStatus, (AssignmentDTO & {startDate: Date | number})[]> = {
         inProgressRecent: [],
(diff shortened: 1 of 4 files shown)


barna-isaac:
The fix works as advertised, and the PR even makes things simpler. Before the PR, there had been (just in these files) 3 implementations that checked whether an assignment was due:
- an `isOverdue` function in `assignments.ts`, which is what caused the bug described in the description
- another implementation that checked whether an assignment is overdue in `filterAssignmentsByStatus` in `assignments.ts`, which was implemented differently from but worked consistently with the implementation in `MyAssignments.ts`
- an implementation  in `MyAssignments.tsx`

This PR changes the implementation of `isOverdue` in `assignments.ts` so it's consistent with the implementation in `MyAssignments`, and uses that function in both places, thereby fixing the bug, and only 2 implementations no remain.

I've manually checked that the behavior on `MyAssignments` did not change using 4 test cases: in the BST (+ offset) and EST (- offset) time zones, I checked that we work as expected half an hour before and after the due date. In all 4 cases, `MyAssignment.tsx` works the same way as before, so congrats on not breaking it.

I'd like to point out that both implementations are still quite difficult to follow, and also inconsistent:
- in `+`-offset time zones such as British Summer Time (London), compared to the date on the server, the due date is midnight, local time of the following day (so `2025-01-01T00:00:00Z` will be due `2025-01-01T23:00:00Z`, giving the user an extra 23 hours compared to what's stored on the server)
- in `-`-offset time zones such as Eastern Summer Time (New York), compared to the date on the server, the due date is midnight, local time of THAT DAY (so `2025-01-01T00:00:00Z` will be due `2025-05-01T04:00:00Z`, giving the user just 4 extra hours compared to what's stored on the server).

This is inconsistent: there is only a 5-hour difference between London and New York, but the due dates are 19 hours apart. Also, neither due dates are when the server would consider it due (which is `2025-01-02T00:00:00Z`).

I'd just like to have this recorded, so when we decide to fix this, we'll know what we're dealing with. For now, I think it's enough that the bug is fixed and at least the dashboard is consistent with "My Assignments".

barna-isaac (approved):

src/app/services/assignments.ts
```diff
@@ -15,13 +15,13 @@ function createAssignmentWithStartDate(assignment: AssignmentDTO): AssignmentDTO
     return {...assignment, startDate: assignmentStartDate};
 }
 
+const now = new Date();
+const midnightLastNight = new Date(now);
+midnightLastNight.setHours(0, 0, 0, 0);
+
```
barna-isaac, line 21: Now that these are variables in the module, they run at slightly different times than they would as part of the function body. In the function body, these would re-evaluate each time the component re-renders (so after a user interaction). This way, they are calculated just once, when the page loads. Especially for sites that might be displayed for a ling time without the user refreshing, the difference might add up.

The existing behaviour could be preserved by making these functions (so `now()` instead of `now`). Just wanted to make sure we're aware of this, but I don't insist on restoring the old behavior.

PR #1400, opened 2025-04-14
Feature/quiz view redesign

barna-isaac:
Besides applying https://github.com/isaacphysics/isaac-react-app/pull/1396, which introduces the `test/view` page, contains the following changes
- `QuizSidebar` now shows the `subject` and `topics`
- `QuizFooter` now consistently ends at `QuizSidebar` across the `attempt` and `preview` pages

### DIFF
--- a/src/app/components/pages/quizzes/QuizView.tsx
+++ b/src/app/components/pages/quizzes/QuizView.tsx
@@ -18,7 +18,7 @@ const pageHelp = <span>
 const Error = buildErrorComponent("Unknown Test", "There was an error loading that test.", rubricCrumbs);
 
 const FooterButton = ({link, label}: {link: string, label: string}) => <Col className="d-flex">
-    <Button className="flex-fill d-flex align-items-center justify-content-center mb-3" color="secondary" tag={Link} to={link}>
+    <Button className="flex-fill d-flex text-nowrap align-items-center justify-content-center mb-3" color="secondary" tag={Link} to={link}>
         {label}
     </Button>
 </Col>; 
--- a/src/app/components/pages/quizzes/QuizView.tsx
+++ b/src/app/components/pages/quizzes/QuizView.tsx
@@ -12,7 +12,7 @@ import { Spacer } from "../../elements/Spacer";
 const pageLink = () => '';
 
 const pageHelp = <span>
-    View information about a test without adding it to "My Tests". This page page does not show any questions.
+    View information about a test without adding it to &quot;My Tests&quot;. This page page does not show any questions.
 </span>;
 
 const Error = buildErrorComponent("Unknown Test", "There was an error loading that test.", rubricCrumbs);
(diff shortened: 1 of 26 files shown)


src/app/components/pages/quizzes/QuizView.tsx
```diff
@@ -38,7 +38,7 @@ export const QuizView = ({user}: {user: RegisteredUserDTO}) => {
         quizId: quizRubricQuery.data?.id,
     };
 
-    return <Container className={`mb-5 ${attempt?.quiz?.subjectId}`}>
+    return <Container data-testid="quiz-view" className={`mb-5 ${attempt?.quiz?.subjectId}`} data-bs-theme={getThemeFromTags(attempt.quiz?.tags)}>
```
jacbn, line 42: If `attempt?.quiz?.subjectId` was exclusively for theming in the old world, we don't need it in the `className` anymore. Same goes for the other Quiz components.

src/test/pages/QuizAttempt.test.tsx
```diff
@@ -53,6 +53,15 @@ describe("QuizAttempt", () => {
             await studentAttemptsQuiz();
             await expectLinkWithEnabledBackwardsNavigation("Continue", `/test/attempt/${quizId}/page/1`, `/test/attempt/${quizId}`);
         });
+
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+        isPhy && describe('sidebar on redesigned Physics site', sideBarTestCases(studentAttemptsQuiz));
```
jacbn, line 58: We probably shouldn't just be ignoring these. I know they are unproblematic but it's just adding to the pile of errors you wanted to get rid of; you can replace them with
```ts
if (isPhy) {
    describe('sidebar on redesigned Physics site', sideBarTestCases(studentAttemptsQuiz));
}
```

jacbn (commented):
Also see https://github.com/isaacphysics/isaac-react-app/pull/1396; there are a couple of redesign points there.

src/test/pages/QuizAttempt.test.tsx
barna-isaac: done in `bea2e6ea`. I was wondering what to do about this, but have seen many other tests also using these -- figured we might just disable the rule. At any rate, I'm also happy with the rule enabled -- just let us get to a point where we start taking these rules seriously and have code style enabled and enforced in the build^^

src/app/components/pages/quizzes/QuizView.tsx
barna-isaac: fixed in `da279089`. Did this across `Quiz` components, and also fixed a bug where the `AttemptFeedback` components only had the deprecated classes, which did nothing.

barna-isaac:
Some suggestions I challenged, but most of them I did! Please have a look again, and let me know if I should maybe join -- using the correct types meant I needed to adjust `QuizAttemptComponent` quite a bit. Again, please also tell me if you think there's a way for this to be easier.

src/test/pages/QuizAttempt.test.tsx
jacbn: I'm not against disabling the rule, but we should do it globally rather than by-line if we were going to. I think I slightly prefer it with, but happy to be overruled.

jacbn (approved):
Looks good!

PR #1406, opened 2025-04-25
Accessibility Improvements

jacbn:
See related card for more; happy to discuss the reasons behind any of these changes if you are interested, but will do in a talk in a week or two. 

---

Reworks the bulk of the `:focus` styling on the site to use one consistent mixin, `@mixin focus-outline`, which uses a box shadow and an invisible outline (for compatibility with Windows' high contrast mode). Switches much of the old-style `:focus` styling, which activates on any element which has any focus (including when clicked), to `:focus-visible`, which activates when the browser determines that styling is necessary (usually just when tabbed to), removing the styling when it is not needed. This means the styling can be made more obvious and assistive for those who need it, without impacting those who do not. Also makes use of landmark elements on pages with more complicated layouts (e.g. those with sidebars), so that SR users can more easily move between distinct sections.


axlewin (commented):
This is a big improvement, especially for physics. Just a few places where the `focus-visible` styling doesn't quite look right:

Physics:
- Quick questions: show/hide answer button has no tabbed styling

Ada:
- Focus on "Create new group" button (`/groups`) behaves strangely. On Firefox, I get stuck on this element and can't tab away from it.
- Tabbed styling looks strange on some elements: "Published" indicator, assignment names (`/assignments`), footer links on concepts/questions within assignments
- Share/delete buttons have no tabbed styling (see e.g. `/quizzes/set`)
- Print/report button icons become illegible when in focus (white on a pale background - this is already the case on the live site, but it looks stranger now that the share button is styled differently).
- `/students`/`/teachers` have sections where the background colour is very similar to the buttons' focus outline colour

axlewin (approved):

PR #1408, opened 2025-04-28
Ada iconography consistency

jacbn:
Adds the basis for icons on Ada to work in the same way as they do on the Physics redesign.
Also adjusts the `ascent-override` property of Ada fonts to correctly vertically align the font, which has been ~2px too high since the site's inception, to make sure text-aligned elements (like icons) actually look aligned.


sjd210 (approved):
Since this touches lots of things very minorly, I've had a good poke around the site to make sure everything works across different pages at various resolutions, and it all seems pretty consistent. 

I won't be *surprised* if we find something odd down the line because of the `ascent-override` property - but there aren't any problems anywhere I can think of, so I'm not too concerned 👍 LGTM

PR #1410, opened 2025-04-29
Close modal after setting assignment

axlewin:


axlewin:
This was an Ada request, but it seems sensible to apply to both sites.

sjd210 (approved):
Seems reasonable. I do wonder whether we should be doing the same thing for the "Set Test" modal too, but for the sake of getting this into the release in time I'll merge this now and we can ask about that later.

PR #1411, opened 2025-04-29
Use delete icon on gameboard builder

axlewin:
On Ada: Use the same icon as deleting an assignment for deleting gameboard questions.

### DIFF
--- a/src/app/components/elements/GameboardBuilderRow.tsx
+++ b/src/app/components/elements/GameboardBuilderRow.tsx
@@ -25,6 +25,7 @@ import { Spacer } from "./Spacer";
 import { LLMFreeTextQuestionIndicator } from "./LLMFreeTextQuestionIndicator";
 import { StyledCheckbox } from "./inputs/StyledCheckbox";
 import { Markup } from "./markup";
+import { Button } from "reactstrap";
 
 interface GameboardBuilderRowInterface {
     provided?: DraggableProvided;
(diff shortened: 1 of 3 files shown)


src/app/components/elements/GameboardBuilderRow.tsx
```diff
@@ -62,38 +63,47 @@ const GameboardBuilderRow = (
...
-                                currentQuestions.setQuestionOrder(newQuestionOrder);
-                                if (provided) {
-                                    undoStack.push({questionOrder: currentQuestions.questionOrder, selectedQuestions: currentQuestions.selectedQuestions});
-                                    redoStack.clear();
-                                }
-                            }
-                        }}
-                    />
+                    {isAda && provided
+                        ? <Button outline className="bin-icon d-inline-block outline"
```
sjd210, line 93: `bin-icon` is probably a bit big for this table. I think this ought to be no larger than 32px diameter to maintain the pre-existing height of the row (I reckon it's suitable to create a separate `bin-icon-small` class to be more suitable size for this).

src/app/components/elements/GameboardBuilderRow.tsx
```diff
@@ -62,38 +63,47 @@ const GameboardBuilderRow = (
...
+                            id={`gameboard-builder-include-${question.id}`}
+                            title="Deselect question"
+                            aria-label="Deselect question"
```
sjd210, line 96: Now that this isn't a select box, I don't think this should say "Deselect". We should make it consistent with the new row heading.

```suggestion
                            title="Remove question"
                            aria-label="Remove question"
```

sjd210 (changes requested):

sjd210 (approved):
Nice 👍 LGTM

src/app/components/elements/GameboardBuilderRow.tsx
sjd210: I think it's fine. The circle probably isn't as *necessary* at this point, but it's consistent and not causing any issues so 🤷‍♀️

PR #1415, opened 2025-05-01
Update subject-stage landing pages

axlewin:


src/app/components/pages/subjectLandingPageComponents.ts
```diff
@@ -159,11 +184,10 @@ export const getLandingPageCardsForContext = (context: PageContextState, stacked
...
 
-    const baseCards: LandingPageCard[] = [
-        QuestionFinderCard,
-        ConceptPageCard,
-        PracticeTestsCard
-    ];
+    const baseCards: LandingPageCard[] = 
+        context.stage.includes("11_14") && context.subject === "physics"
+            ? [StepIntoPhyCard, ConceptPageCard, QuestionFinderCard]
+            : [QuestionFinderCard, ConceptPageCard, PracticeTestsCard];
```
sjd210, line 190: I think there's an easy way to remove the practice tests card for GCSE Maths/Chemistry without needing to majorly restructure things here (and we ought to - those combinations don't *have* practice tests). 

We can add another special case for them in `baseCards` that's only two elements long, and removing the ending `null`s from their entry in `subjectSpecificCardsMap` so they're also only two elements long. I think that by itself will work exactly as we need.

sjd210 (changes requested):

src/app/components/pages/subjectLandingPageComponents.ts
axlewin: Yeah, that works. It kind of defeats the point of having base cards, but I suppose having 2 special cases isn't much worse than having 1.

Fwiw, there are other combinations that don't have practice tests yet either, but presumably only removing this card for gcse maths/chem was an intentional content decision.

sjd210: I think two special cases is the reasonable limit; if we need to change it further it's probably best to have another mapping (which uses the current base case most of the time).

And hrm you're right. I'm working under the assumption that they have plans - if not, it won't be a big deal to also remove the card for those other combinations later on 🤷‍♀️

sjd210 (approved):
Cool. These are definitely better now 👍 Much more consistent

PR #1417, opened 2025-05-01
Redesign: general improvements #5

jacbn:
- Display book page errors more cleanly
- Add a new breakpoint **for containers only** to increase content readability on huge screens
- Adjust ALVIs spacing to prevent overflowing on smaller screens
- Open the topics field by default on Phy QF
- Various minor fixes for Assignment Schedule


axlewin (commented):
Looks good. The only thing I've noticed wrt the new container breakpoint is that the dashboard is now significantly wider than the rest of the homepage on really big screens - it could probably use some padding on `3xl`?

PR #1422, opened 2025-05-02
Update group management modal

axlewin:


axlewin:
This was an Ada request, but the copy changes seem sensible to apply to physics too.

PR #1426, opened 2025-05-08
Visible question correctness

jacbn:
- Adds a background colour for fully correct questions in the QF;
- Moves the status symbol to a notification-style bubble next to the Q icon (would appreciate any views on whether we should be using this for in progress questions also)
- Applies the same background colouring for gameboards (these do not currently use AVLIs, perhaps a task for the future)


sjd210 (approved):
All* the changes here look good and straightforward enough :+1:

---

> (would appreciate any views on whether we should be using this for in progress questions also)

I think this looks odd, and really doesn't read intuitively as being in-progress. I think it would be comparatively less odd to have the in-progress status back where it had originally been on the ALVI while still keeping the correct one in its new position, which does look quite nice!

src/app/components/elements/list-groups/AbstractListViewItem.tsx
```diff
@@ -115,9 +124,6 @@ export const AbstractListViewItem = ({icon, title, subject, subtitle, breadcrumb
                 {audienceViews && fullWidth && <div className="d-flex mt-1"> 
                     <StageAndDifficultySummaryIcons audienceViews={audienceViews} stack/> 
                 </div>}
-                {status && (below["lg"](deviceSize) || fullWidth) && <div className="d-flex mt-1">
-                    <StatusDisplay status={status}/>
-                </div>}
```
sjd210, line 120: Removing this instance of `StatusDisplay` means we have no In Progress marker for <= "sm" screens anymore, as they aren't considered `fullWidth` for ALVI purposes. It used to be split between this position and the right position as <= "lg" and >= "xl" respectively (which is how the designs had it originally), but now it's only the main right position, and for >= "md" instead.

I don't mind too much which position it's kept in at what sizes, but it really ought to be _somewhere_ for <="sm" screens.

sjd210 (approved):

PR #1431, opened 2025-05-09
Redesign gameboard builder filters

axlewin:
In the question search modal:
- Use QF-style hierarchy filters for topic selection
- Move filters to "sidebar" on large screens

A few points/questions:
- The topic filter is now styled differently to the other filters, which looks strange on Ada. Should all of the filters be changed to their QF equivalents?
- The `lg` layout on physics isn't great; we could just stack all of the filters vertically as with <`lg`, but this requires a lot of scrolling to reach the results.

### DIFF
--- a/src/app/components/elements/modals/QuestionSearchModal.tsx
+++ b/src/app/components/elements/modals/QuestionSearchModal.tsx
@@ -46,7 +46,7 @@ import { HierarchyFilterTreeList } from "../svg/HierarchyFilter";
 import { ChoiceTree, getChoiceTreeLeaves } from "../panels/QuestionFinderFilterPanel";
 import { CollapsibleList } from "../CollapsibleList";
 import { StyledCheckbox } from "../inputs/StyledCheckbox";
-import { updateTopicChoices, initialiseListState, listStateReducer } from "../../../services/questionSearch";
+import { updateTopicChoices, initialiseListState, listStateReducer } from "../../../services";
 
 // Immediately load GameboardBuilderRow, but allow splitting
 const importGameboardBuilderRow = import("../GameboardBuilderRow");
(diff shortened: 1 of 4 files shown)


mwtrew (approved):
Looks good. I also noticed we're using `searchQuestions` differently [here](https://github.com/isaacphysics/isaac-react-app/blob/bd9866bb5ae1dddca6984d321f09616e4adcd2c4/src/app/components/elements/modals/QuestionSearchModal.tsx#L124) vs. the [question finder](https://github.com/isaacphysics/isaac-react-app/blob/8c34083508cd6deb159e8831e19869ffbe00ea23/src/app/components/pages/QuestionFinder.tsx#L222), have you looked into that and are happy with it? From a quick check it does seem to give the same results.

axlewin:
> I also noticed we're using `searchQuestions` differently [here](https://github.com/isaacphysics/isaac-react-app/blob/bd9866bb5ae1dddca6984d321f09616e4adcd2c4/src/app/components/elements/modals/QuestionSearchModal.tsx#L124) vs. the [question finder](https://github.com/isaacphysics/isaac-react-app/blob/8c34083508cd6deb159e8831e19869ffbe00ea23/src/app/components/pages/QuestionFinder.tsx#L222), have you looked into that and are happy with it? From a quick check it does seem to give the same results.

Hmm. The only thing I'm not sure about here is how the `tags` value should be updated. The QF approach of calling [`getChoiceTreeLeaves`](https://github.com/isaacphysics/isaac-react-app/blob/8c34083508cd6deb159e8831e19869ffbe00ea23/src/app/components/pages/QuestionFinder.tsx#L224) directly in the search query is definitely neater than [this `useEffect`](https://github.com/isaacphysics/isaac-react-app/blob/bd9866bb5ae1dddca6984d321f09616e4adcd2c4/src/app/components/elements/modals/QuestionSearchModal.tsx#L210), and I tried it that way, but we're using `searchTopics` in a few other places in the question search modal so I decided updating it everywhere with the `useEffect` might be cleaner. From a React perspective though, maybe we should be using the QF approach in both places 🤔.

axlewin:
I'm not sure why the tests are failing. It's giving `TypeError: subjectSpecificTagService is not a constructor`, but nothing here should affect the tag services...

PR #1437, opened 2025-05-12
Various visibility fixes for small screens

sjd210:
Mostly small changes to overflow behaviour (see commit messages), but most notably:

- Adds an `optionBar` for buttons like _Published_ and _Share_ to be displayed in the same row as the collapsed sidebar on small screens for pages like **Glossary** or **Terms of Use** to avoid a lot of vertical empty space.
- Allows events at the bottom of subject-stage landing pages to be displayed vertically for sizes that fit it better.
- A scrollbar for the dashboard assignment card. We're already limiting the titles on this card to 4 lines, but there are many potential overflow problems unless they're limited to 2 - and even then there are some niche ones e.g. for particularly small screens. The scrollbar should catch all of these, but did require adding some negative padding to preserve `on-hover` behaviour, so it'd be good to check that doesn't cause any oddities.


axlewin (approved):
This all looks good. I'm not totally convinced that [`cloneElement`](https://github.com/isaacphysics/isaac-react-app/blob/redesign/hotfix/small-visibility/src/app/components/pages/Generic.tsx#L81) is the safest way to deal with the option bar, but all of the alternatives in the [React docs](https://react.dev/reference/react/cloneElement) need much bulkier solutions so I think it's ok as-is.

(I'm not going to fix the VRT conflict for now considering how many different PRs want to update it).

sjd210:
@axlewin Yeah, I wasn't entirely happy with `cloneElement` myself either. I'll have another quick look for an alternative way while we're waiting on release now to see if there's an easy cleaner path I can find, but I think it'll work otherwise.

PR #1440, opened 2025-05-12
Metadata Update

jacbn:
Update the metadata of various pages, phasing out "Isaac Physics" in favour of "Isaac" or "Isaac Science".

The second commit sets the default subject on the (to-be) isaacscience domain to be Physics, to stay in line with the current functionality. I would rather this had no default tag, but there may be a number of existing pages relying on this.


sjd210 (approved):

cypress/support/component-index-phy.html
```diff
@@ -18,15 +18,15 @@
     <link rel="preload" as="image" href="/assets/phy/line.svg">
     <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no, shrink-to-fit=no" />
     <meta name="theme-color" content="#000000" />
-    <meta name="application-name" content="Isaac Physics" />
-    <meta name="description" content="Isaac Physics is a project designed to offer support and activities in physics problem solving to teachers and students from GCSE level through to university." data-react-helmet="true" />
-    <meta property="og:site_name" content="Isaac Physics" />
-    <meta property="og:title" content="Isaac Physics" data-react-helmet="true" />
+    <meta name="application-name" content="Isaac Science" />
+    <meta name="description" content="Isaac Science is a project expanding out of Isaac Physics to provide resources to develop confidence in problem-solving not only physics (year 7 through to university) and maths (GCSE to university) but also in chemistry (GCSE to university) and biology (A level)." data-react-helmet="true" />
```
jsharkey13, line 22: Aside from the fact that this is unwieldly and will be truncated by previews and search engines, I don't think "... provide resources to develop confidence in problem-solving not only physics (year 7 through to university) and maths ..." makes any sense!

src/app/components/pages/Support.tsx
```diff
@@ -122,8 +122,8 @@ export const SupportPageComponent = ({match: {params: {type, category}}}: RouteC
 
     const metaDescriptionMap = siteSpecific(
         {
-            "student": "Find answers to your questions about Isaac Physics in our FAQ for students.",
-            "teacher": "Find answers to your questions about Isaac Physics in our FAQ for teachers.",
+            "student": "Find answers to your questions about Isaac Science in our FAQ for students.",
+            "teacher": "Find answers to your questions about Isaac Science in our FAQ for teachers.",
```
jsharkey13, line 126: Surely this could be plain "Isaac" and not be confusing?

PR #1449, opened 2025-05-14
Improve Ada printing behaviour

sjd210:
- Fixed behaviour from the redesign causing the [With hints | Without hints] to overlap with the share button in an unreadable way
- Changed the code snippet "View on GitHub" text to the GitHub link itself when printing
- Show tab titles above all individual tabs when printing. The original request was for `codeTabs` specifically, but having tested different options I think this makes sense for all tabs on both sites (helps for a question like [sort_31_v2](https://adacomputerscience.org/questions/sort_31_v2) which isn't using `codeTabs` (or even something like the [About Us](https://adacomputerscience.org/about) page, if you really feel like it)). I also checked viability of doing similar work for our dropdown and button `<Tab>` styles, but they're less straightforward and don't tend to be used on printable pages anyway.

    _Note: I found it was clearer to show all titles with the selected tab highlighted rather than just the name of the selected tab by itself. It helps show that the tabs are collected together in a block._

### DIFF
--- a/src/app/components/content/IsaacCodeSnippet.tsx
+++ b/src/app/components/content/IsaacCodeSnippet.tsx
@@ -56,7 +56,8 @@ const IsaacCodeSnippet = ({doc}: IsaacCodeProps) => {
         </div>
         {doc.url && <Row>
             <Col className="text-center mb-2">
-                <a href={doc.url} onClick={logViewOnGitHub} target="_blank" rel="noopener noreferrer">View on GitHub</a>
+                <a className="no-print" href={doc.url} onClick={logViewOnGitHub} target="_blank" rel="noopener noreferrer">View on GitHub</a>
+                <a className="only-print mb-2" href={doc.url} target="_blank" rel="noopener noreferrer">{doc.url}</a>
             </Col>
         </Row>}
     </>;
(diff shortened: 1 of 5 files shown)


src/app/components/content/IsaacCodeSnippet.tsx
```diff
@@ -56,7 +56,8 @@ const IsaacCodeSnippet = ({doc}: IsaacCodeProps) => {
         </div>
         {doc.url && <Row>
             <Col className="text-center mb-2">
-                <a href={doc.url} onClick={logViewOnGitHub} target="_blank" rel="noopener noreferrer">View on GitHub</a>
+                <a className="no-print" href={doc.url} onClick={logViewOnGitHub} target="_blank" rel="noopener noreferrer">View on GitHub</a>
+                <a className="only-print mb-2" href={doc.url} target="_blank" rel="noopener noreferrer">{doc.url}</a>
```
axlewin, line 60: I don't think the `mb-2` here does anything - removing it doesn't seem to affect the spacing.
```suggestion
                <a className="only-print" href={doc.url} target="_blank" rel="noopener noreferrer">{doc.url}</a>
```

axlewin (commented):
Looks good. The only issue I can see is that this adds some unnecessary padding above the first code snippet in each section:

Before:
image

After:
image

sjd210:
Oh, nice spot. In which case, I'll bring the padding for the whole heading section more in-line with how it is on the non-print site. It makes it all more consistent this way. (In fact, I think this is what I was trying to do with that rogue `mb-2`, but clearly I'd forgotten about it)

PR #1463, opened 2025-05-16
Redesign: concept finder improvements

axlewin:
- Add stage filtering to main concept finder
- Don’t automatically deselect subject when fields are deselected (concept finder behaviour should match QF behaviour here)
- Move “these concepts have been filtered…” message out of sidebar


axlewin:
> For a specific narrow viewport size, the "Browse all questions" button spans 3 lines on the Question Finder page. I'm not sure whether this is something we need to fix, but still felt it was worth pointing out. The same happens on the new Concept page as well.

Good spot! [This](https://github.com/isaacphysics/isaac-react-app/pull/1463/commits/61371f7e9d5a1371860c2fbf64163f3107479f2e) should fix it, but it would be useful to test a few different pages on your browser too.

barna-isaac:
A good thing that I'd like to highlight is that the "reselect parent" fix is also applied to the `Practice Quizzes` page, which also had the same bug.

axlewin:
> A good thing that I'd like to highlight is that the "reselect parent" fix is also applied to the `Practice Quizzes` page, which also had the same bug.

Yeah - we're not _really_ using that page (it isn't linked to from anywhere), but the behaviour might as well be consistent in case anyone does use it.

barna-isaac:
When performing a text search, subject tags narrow as more specific queries produce less matches, but filter tags remain constant. I think the behavior should be consistent.

https://github.com/user-attachments/assets/54d5cce5-4df1-4ad5-9ac7-9d1d24959999

barna-isaac (changes requested):

axlewin:
Thanks for looking at this! The latest batch of commits should address all of your comments, plus a few other inconsistencies I found; there's justification for some choices in the commit messages, but let me know if you have any more questions/suggestions.

barna-isaac:
@jacbn , can you please look at this? I've been trying to get these alginments right, but I just hate how unevenly and how much white space we waste. Does "Search and Filter" really need these huge margins? I'd be happiest just removing them, but this is what it looks like on every other page too, and I don't want to introduce inconsistencies.

Screenshot 2025-05-22 at 17 05 02

Screenshot 2025-05-22 at 17 09 04

axlewin:
@barna-isaac : Whilst I agree the inline style wasn't the ideal solution, I'm not sure that [this](https://github.com/isaacphysics/isaac-react-app/pull/1463/commits/7401ef6400b7cf64bd2a3e49ca2278f7da22b061) is either. It still looks strange on certain screen widths, just differently:
image

barna-isaac:
@axlewin, do you mean it's looking weird because the text is now three lines? Maybe we could just change the button text so it's to just "All questions" and "All concepts"?

barna-isaac:
Okay, the code for concept filtering looks very pretty now, but I still wonder whether this result is correct. On this example, we have 2 physics questions and 1 chemistry question, but the stages only show 2 questions. One of the questions is a draft, but why is it that the draft question is still counted for the subject, and not counted for the stages? Also, should we even show the draft question?

Screenshot 2025-05-23 at 09 58 36

axlewin:
> do you mean it's looking weird because the text is now three lines?

Yeah - compared to before, where this was 2 lines of text & 2 lines for the button. I'm not sure it makes a big difference though, if you prefer it this way.

I'm still not sure I understand your changes though. Does the `mb-3` on the `<p>` do anything?

axlewin:
> @axlewin: no, I just put the `mb-3` on `<p>` because there also needs to be an `mb-3` on the button, and this highlights the symmetry. But `<p>` gets `mb-3` by default. What do you think should we do to make it look better? Or should we maybe just leave it as is?

It's not the neatest solution, but we could just remove the hardcoded style and then change the classes on the outermost `div`  to allow the button to wrap below the text on the problematic screen widths:
`<div className="d-flex align-items-baseline flex-wrap flex-md-nowrap flex-lg-wrap flex-xl-nowrap ">`

This at least stops either element wrapping to 3 lines, and avoids complicating the DOM by adding another wrapper `div` around the `AffixButton`.

There are a handful of screen sizes where it will look strange either way though, so we can leave it as-is if you prefer.

axlewin:
> However, I think this is a valid issue: if a selected filter becomes empty, it remains applied but disappears from the screen. This leads to invalid screens, where "Physics" is partially selected but doesn't actually show any selected descendants, or the final screen of the attached recording, where it looks like only the Biology filter is selected but we don't show any results.
> 
> I think the easiest solution would be if we always showed empty filters.

As discussed, always showing empty stage filters isn't ideal because it leads to cases where you can't deselect a field (try selecting physics > thermal > year 7 & 8). We could show empty field filters too, but this might be confusing for students (e.g. if they've selected year 7 & 8, and calculus is showing up as an option). I also worry that when only published content is shown, this problem would be worse, because there would be more stage/field combinations with no results.

My proposal [here](https://github.com/isaacphysics/isaac-react-app/pull/1463/commits/a5ff701bb8199a0948b6d57ce7b226d96a3b7854) is to always show _selected_ stage filters, even if there are no results. From a quick test this seems to avoid any cases where users can get 'stuck' and unable to deselect a selected filter.

It's worth noting that this is different from the implementation discussed on `/practice_tests`, but I think it's clearer.

(As always, please see if you can find any more cases that still don't work!)

barna-isaac:
@axlewin I like the idea of always showing selected stage filters, and propose that we extend it to always show selected field filters as well.

axlewin:
I'm not concerned about the whitespace issues you mentioned (as you said, the spacing matches other pages). Up to you if you'd like to address those before merging.

PR #1466, opened 2025-05-19
submit inline questions in a sequence

barna-isaac:
This avoids a race condition where the API could not save multiple answers that were submitted at the same time for anonymous users. This is because, for anonymous users, when saving an answer, the API creates a lock using the user's ID, see:

https://github.com/isaacphysics/isaac-api/blob/51296605cef5b532f6f6b90d9a9575001172d4a6/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java#L92.


barna-isaac:
I've reproduced and checked that the issue on `dev`.

Before applying the fix, some questions would mark but others would fail and a notification is shown. After refreshing the page, only some of the answers are saved.

https://github.com/user-attachments/assets/218ab552-eb80-4dde-8c62-9b7f458be322

After applying the fix, all questions mark and no error is shown. After refreshing the page, all answers are saved.

https://github.com/user-attachments/assets/ea43c7de-2a53-495f-9a78-ac057aec58cc

jacbn (approved):

PR #1468, opened 2025-05-20
Redesign/general-improvements-9

jacbn:
Yet more bugfixes!

Of the larger changes:
- Reset the status of our Redux store's `doc` on page switch so that calls to `doc.get` (used in various content components to determine the page type) are not stale;
- Move the styling of accordion-less questions one component out (to the surrounding `<Form/>`) so that hints are included and appear in the same "block";
- Refactor QF to make spacing around checkboxes consistent, and to fix an issue with the book filters

### DIFF
--- a/src/scss/common/questions.scss
+++ b/src/scss/common/questions.scss
@@ -24,7 +24,7 @@ $cloze-scrollbar-height: 4px;
   min-width: auto;
 }
 
-.question-component {
+form:has(> .question-component) {
   margin-bottom: 2rem;
   clear: both;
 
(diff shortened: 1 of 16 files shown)


src/scss/common/questions.scss
```diff
@@ -70,6 +70,12 @@ $cloze-scrollbar-height: 4px;
   }
 }
 
+.question-component {
+  background-color: white;
+  box-shadow: 0 2px 30px 0 rgba(0, 0, 0, 0.08);
```
axlewin, line 75:
```suggestion
  background-color: white;
  border-radius: $border-radius-lg;
  box-shadow: 0 2px 30px 0 rgba(0, 0, 0, 0.08);
```
minor visual thing with the top corners on accordion-less questions:
image

axlewin (commented):
nice, one small visual bug but otherwise this all looks good 👍

src/scss/common/questions.scss
jacbn: Hm. That CSS block should be switched to the form styling, otherwise it doesn't give a background to hints..? Not sure why it wasn't.

axlewin (approved):

PR #1475, opened 2025-05-21
Subject Overview Page cleanup

sjd210:
General changes to line things up with the mock ups for each Subject Overview Page:
- Adding symbols to explain practice and challenge difficulties
- Including a greyed out option for GCSE Biology in the card list
  - (+ removing blanks from the card list on small screens)
- Adding the example question deck to the bottom of the screen.
  - Physics hasn't made one yet. We can confirm it has the same id as the commented one and add it in fairly easily once available.


src/app/components/pages/SubjectOverviewPage.tsx
```diff
@@ -112,15 +149,28 @@ export const SubjectOverviewPage = withRouter((props: RouteComponentProps) => {
...
+            <p className="mt-3">
                 In Isaac {humanSubject},
                 <ul>
                     <li>Practice questions are those that require one concept or equation to solve.</li>
                     <li>Challenge questions are those that require one or more concepts, or require more creativity to solve the problem, helping to develop important problem solving skills. </li>
                 </ul>
             </p>
 
-            {/* <ExampleQuestions/> */}
+            <ExampleQuestions subject={pageContext.subject} />
```
axlewin, line 173:
```suggestion
            <div className="d-flex justify-content-center">
                <div className="w-100 w-md-50">
                    <ExampleQuestions subject={pageContext.subject} />
                </div>
            </div>
```
The mockups have this roughly half-width and centered, which imo does look better for bigger screens (especially with the difficulty icons above centered anyway). Happy to merge as-is if you disagree, though.

axlewin (commented):
Lgtm, just a minor layout suggestion for the sample question deck

src/app/components/pages/SubjectOverviewPage.tsx
sjd210: Hrm, I partly agree. I think we can make this code itself cleaner by passing the new `className` into the `ListView` directly rather than two layers of divs - however more notably I think that frequently switching width between 100->50->100->50 is quite jarring, but we can ease that and create a better flow than either other option by using `w-75`.

I assume this still matches your wishes nicely, but if it _doesn't_, I'd prefer `w-100` to `w-50`.

PR #1479, opened 2025-05-22
Show clearer error message when site loaded offline

sjd210:
This doesn't cover every case, but [window.navigator.onLine](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine) cannot cause false negatives so this is a safe improvement over the current system. I've briefly tried it with Firefox, Chrome and Edge, and I trust the documentation that says its widely available. 

If we were sufficiently concerned about edge cases (e.g. being connected to a network but not receiving internet from it), we could try also reading JS errors or pinging the client, but I don't think its worth the effort.

### DIFF
--- a/src/app/components/elements/PageFragment.tsx
+++ b/src/app/components/elements/PageFragment.tsx
@@ -20,12 +20,12 @@ export const PageFragment = ({fragmentId, ifNotFound}: PageFragmentComponentProp
             <small>
                 {"We're sorry, page fragment not found: "}
                 <code>{fragmentId}</code>
-                <p>
+                {!window.navigator.onLine && <p>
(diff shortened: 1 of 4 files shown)


src/app/components/elements/PageFragment.tsx
```diff
@@ -19,6 +20,12 @@ export const PageFragment = ({fragmentId, ifNotFound}: PageFragmentComponentProp
             <small>
                 {"We're sorry, page fragment not found: "}
                 <code>{fragmentId}</code>
+                <p>
```
jsharkey13, line 23: This appears to be unconditional, instead of checking if you're offline!

PR #1489, opened 2025-05-29
Redesign/markbook

mwtrew:


jacbn:
There are a _lot_ of VRT changes; I am not merging them yet for the sake of the reviewer(s) – though I am happy to if you don't spot anything. The bulk of them come from:
- On Ada, generic card styling has been changed to slightly increase padding and remove the outline, according to designs. This doesn't seem to affect the readability of anything, and is more consistent with the theming.
- On Physics, there was a bug with button styling that was turning neutral-themed icons green. We probably do prefer it the old way, but we should solve this properly rather than relying on the bug.

src/app/components/pages/quizzes/QuizTeacherFeedback.tsx
```diff
@@ -129,10 +132,26 @@ export const QuizTeacherFeedback = ({user}: {user: RegisteredUserDTO}) => {
 
                 <div className={`assignment-progress-details bg-transparent ${pageSettings.colourBlind ? " colour-blind" : ""}`}>
                     <AssignmentProgressPageSettingsContext.Provider value={pageSettings}>
-                        <AssignmentProgressLegend showQuestionKey />
-                        <div className="p-4">
+                        {/* <AssignmentProgressLegend showQuestionKey /> */}
```
jsharkey13, line 135: Is this still here for future work?

src/scss/cs/icons.scss
```diff
@@ -50,24 +50,16 @@
...
-.icon-group {
-  display: inline-block;
-  overflow: hidden;
-  width: 1.5rem;
-  height: 1.25rem;
-  background: url(/assets/cs/icons/group.svg) no-repeat;
-  margin-left: 10px;
-  margin-right: 10px;
-}
+// TODO: fix instances of this
```
jsharkey13, line 53: Has this been done? Does it still need to be here?

PR #1492, opened 2025-06-02
Redesign: Login modal / Logout button / "Browse all" buttons

sjd210:
Various spacing and layout adjustments:
- The _"Log out"_ button was inconsistent with other navbar buttons on >="lg" screens. It should now be the same size.
- The login/signup modal (which appears after answering 2 questions logged out) has had a slight overhaul. It had needed reconfiguring for redesign-specific changes like the new logo and buttons, but has also had other small issues for a while  (e.g. using old checkboxes) which have now been cleaned up. This _will_ also effect Ada, but I think only for the better.
- The _"Browse all questions"_/_"... concepts"_ buttons on their respective finder pages were creating a lot of empty space on <="md" screens that made the finders difficult to use (since you'd have to scroll to the top to access filters). Bringing the button inline with the sidebar button removes this empty space and generally looks a bit cleaner.


src/app/components/pages/QuestionFinder.tsx
```diff
@@ -464,14 +477,7 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
                         {(pageContext?.subject && pageContext.stage)
                             ? <div className="d-flex align-items-baseline flex-wrap flex-md-nowrap flex-lg-wrap flex-xl-nowrap">
```
jacbn, line 479: Speaking of odd spacing, `align-items-start` seems to work better here. Stops the `p` starting lower down than the button.

src/scss/phy/button.scss
```diff
@@ -30,6 +30,10 @@
     padding: 0.8125rem 1.625rem;
     min-width: 120px;
   }
+
+  &.btn-horizontal {
+    min-width: max-content;
```
jacbn, line 35: Would it be worth making this a helper class? We have `w-max-content`, would be happy with `mw-max-content` too. This has come up before is all.

jacbn (changes requested):
This breaks a couple of small things but looks good on the whole!

src/app/components/pages/QuestionFinder.tsx
```diff
@@ -454,7 +467,7 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
                     selections, setSelections,
                     applyFilters: searchAndUpdateURL, clearFilters,
                     validFiltersSelected, searchDisabled, setSearchDisabled
-                }} />
+                }} optionBar={BrowseAllButton}/>
```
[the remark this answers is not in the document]
sjd210: I agree with everything you're saying. It feels odd either way, but I feel it's at least less of an _active_ problem here.

src/scss/phy/button.scss
jacbn: I'll leave it for now. If it happens again it's easy enough to search the codebase for it to replace old ones!

PR #1496, opened 2025-06-02
Redesign: Account settings cleanup

sjd210:
Most significantly, this moves the 2FA setting from being below the "Save" button in its own column to be inline with all other security settings (on both Ada and Isaac). The reason this hadn't been done initially had been due to issues about nesting a `<form>` within a `<form>` which is forbidden under [HTML spec](https://html.spec.whatwg.org/multipage/forms.html#the-form-element) so can have undefined behaviour.

Currently, we use a hacky workaround for this for the Teacher Connections page (with a similar issue) which works for us but still includes nested `<form>`s so is inherently unreliable. This PR splits the settings page into two sections (one for the wider form content, one for content with its own form) and lifts the submit button to be after both using the [form attribute](https://www.impressivewebs.com/html5-form-attribute/) to associate controls. The only issue I could see with this is that Internet Explorer doesn't support it, but we don't support IE anyway so it should all be fine.

---

Also general code cleanup and small fixes, such as:
- Recombining `AdaMyAccount` and `PhyMyAccount` into one component, as they were already almost identical.
- Replacing `autoFocus` on Teacher Connections search with a `ref`-based manual focus for accessibility (it was already behind a button so wasn't _too_ bad but this also protects us from issues like screen resizing, and gets rid of those pesky red squiggly lines).
- Stopping the Teacher Connections searchbar from slightly resizing the entire header on Isaac, and unnecessarily blocking the group icon on Ada.
- Removing the universal vertical offset to btn-links that had misaligned things, and replacing it with a specific offset for just `<p>` blocks where the original issue had been.


src/scss/cs/my-account.scss
```diff
@@ -28,7 +28,7 @@
 .my-account-tab {
   select.form-select, input.form-control, div.form-control {
     height: 45px;
-    border-radius: 10px !important;
+    border-radius: 10px;
```
axlewin, line 31: This was keeping the teacher connections input border radius consistent with the profile inputs. Even though this is an `!important` overriding another `!important`, I'd vote for keeping it the way it was.

src/app/components/pages/MyAccount.tsx
```diff
@@ -323,193 +323,119 @@ const AccountPageComponent = ({user, getChosenUserAuthSettings, error, userAuthS
...
-                                    <UserBetaFeatures
-                                        displaySettings={myUserPreferences?.DISPLAY_SETTING ?? {}} setDisplaySettings={setDisplaySettings}
-                                        consentSettings={myUserPreferences?.CONSENT ?? {}} setConsentSettings={setConsentSettings}
+                                </TabPane> 
+                            </TabContent>}
+                            <div className={classNames({"py-4 card-footer": isAda})}>
+                                {isPhy && <div className="section-divider-bold"/>}
+                                <div className={classNames("d-flex justify-content-center", {"col-12 col-md-6 offset-md-3": isAda})}>
+                                    <Input
+                                        form="my-account" type="submit" value="Save" className={classNames("btn btn-solid border-0", {"w-100": isAda})}
```
axlewin, line 428: Not introduced by this PR, but while we're restructuring this page - thoughts on adding an `activeTab !== ACCOUNT_TAB.teacherconnections` check for displaying this button? I can see the argument for keeping it consistent across tabs, but it seems confusing to have a button which (as the comment that was previously here noted) is never actually used.

axlewin (commented):
Nice, it's good to finally have the 2FA in the right place. I have a couple of small suggestions, but this all looks good 👍

src/app/components/pages/MyAccount.tsx
axlewin: Ahh, fair enough then 👍

PR #1505, opened 2025-06-04
Update deploy script to support Isaac Science

jsharkey13:
This needs reviewing; it is hard to test the behaviour of except on the live server.


mwtrew (approved):
Looks good, we're using `subject` and `site` in the right places respectively - except perhaps `deploy_etl()` which should be using `subject`, but I see we're skipping that entirely for `site=sci` so that's fine.

PR #1517, opened 2025-06-09
Revision & content-controlled sidebars

jacbn:
Adds support for revision pages, using a new content-controlled sidebar format. Also applies this sidebar to book pages and generic pages if they have the `sidebar` content prop, with the potential for more page types later.


src/app/components/elements/Book.tsx
```diff
@@ -56,7 +54,10 @@ export const Book = ({match: {params: {bookId}}}: BookProps) => {
                 defaultErrorTitle="Unable to load book contents."
                 thenRender={(definedBookIndexPage) => {
                     return <>
-                        <BookSidebar book={definedBookIndexPage} urlBookId={bookId} pageId={pageId} />
+                        {book?.sidebar
+                            ? <ContentControlledSidebar sidebar={book?.sidebar}/>
+                            : <BookSidebar book={definedBookIndexPage} urlBookId={bookId} pageId={pageId} />
```
jsharkey13, line 59: Since the API cannot return pages that have `chapters` any longer, surely this fallback will be just as broken as the other missing sidebar?
Potentially you could fall back to the _books_ page sidebar listing all the books, if you think it needs something?

src/app/components/elements/BookPage.tsx
```diff
@@ -5,7 +5,7 @@ import { IsaacBookDetailPageDTO } from "../../../IsaacApiTypes";
 import { TeacherNotes } from "./TeacherNotes";
 import { EditContentButton } from "./EditContentButton";
 
-const BookSectionLink = ({id, title}: { id: string; title: string }) => {
+export const MetadataContainerLink = ({id, title}: { id: string; title: string }) => {
```
jsharkey13, line 8: Should this not now live in its own component file?

src/app/components/elements/layout/SidebarLayout.tsx
```diff
@@ -1507,6 +1508,89 @@ export const BookSidebar = ({ book, urlBookId, pageId }: BookSidebarProps) => {
     </ContentSidebar>;
 };
 
+const calculateSidebarLink = (entry: SidebarEntryDTO): string | undefined => {
+    switch (entry.pageType) {
+        case "isaacBookDetailPage": {
+            const detailPageSplit = entry.pageId?.split("__");
```
jsharkey13, line 1514: Ditto about a constant!

src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -243,14 +243,13 @@ interface BookDetailListViewItemProps extends Extract<AbstractListViewItemProps,
...
-    const itemLabel = itemBook ? item.id?.slice(`book_${itemBook.tag}_`.length) : undefined;
 
     return <AbstractListViewItem
         {...item}
         icon={{type: "hex", icon: "icon-generic", size: "lg"}}
-        title={`${itemLabel ? (itemLabel?.toUpperCase() + " ") : ""}${item.title}`}
+        title={`${item.subtitle ? (item.subtitle + " ") : ""}${item.title}`}
         subtitle={itemBook?.title}
-        url={itemBook ? `/${documentTypePathPrefix[DOCUMENT_TYPE.BOOK_INDEX_PAGE]}/${itemBook.tag}/${itemLabel}` : undefined}
+        url={itemBook ? `/${documentTypePathPrefix[DOCUMENT_TYPE.BOOK_INDEX_PAGE]}/${item.id?.replace("__", "/").slice("book_".length)}` : undefined}
```
jsharkey13, line 252: This URL is also computed in `SidebarLayout`, but without using the `documentTypePathPrefix`. Since it is duplicated, is it worth pulling it out into a function somewhere?

jsharkey13: It could also use a constant for `__`.

PR #1518, opened 2025-06-10
Glossary filters suggestions

sjd210:


src/app/components/pages/Glossary.tsx
```diff
@@ -145,19 +154,21 @@ export const Glossary = () => {
     useEffect(() => {
         if (isFullyDefinedContext(pageContext) && isSingleStageContext(pageContext)) {
             setFilterSubject(tags.getById(pageContext.subject as TAG_ID));
-            const potentialStage = stageByValue([pageContext.stage.map(x => x === "11_14" ? "year_9" : x)[0]], stagesOrdered.slice(0,-1));
-            setFilterStages(potentialStage ? [potentialStage] : undefined);
+            const potentialStage = stagesByValue([pageContext.stage.map(x => x === "11_14" ? "year_9" : x)[0]], stagesOrdered.slice(0,-1));
```
axlewin, line 157: It's a change from the current behaviour, but while we're looking at this, could we replace this logic with something like `stagesByValue(pageContext.stage.flatMap(x => LEARNING_STAGE_TO_STAGES[x]), stagesOrdered.slice(0,-1))` (and similarly for L101)? I realise that glossary terms have (mostly) been tagged to work consistently with learning stages so far, but from a content perspective it's not obvious to me that tagging a term as just `year_7_and_8` or `further_a` would prevent it from showing up here.

(Your Argand Diagram example illustrates this well, but even if we don't want Further A to count as A Level, I think year 7&8 should definitely count as 11-14).

axlewin (approved):
I have a small question about stages vs learning stages, but this all works nicely, happy for this to be merged with the other branch 👍

src/app/components/pages/Glossary.tsx
sjd210: Oh absolutely agreed, yes. This isn't working consistently with our current contexts and it really ought to 👍

PR #1520, opened 2025-06-12
Rename Ada to Ada CS

mwtrew:
Also replaces hard-coded strings with `SITE_TITLE_SHORT` where possible.

### DIFF
--- a/src/app/components/pages/ClientError.tsx
+++ b/src/app/components/pages/ClientError.tsx
@@ -2,7 +2,7 @@ import React, {useEffect} from "react";
 import {Link} from "react-router-dom";
 import {Col, Container, Row} from "reactstrap";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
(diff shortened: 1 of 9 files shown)


mwtrew:
> This certainly does cover the specific cases from the card, but I can still find a lot of instances of just "Ada" that we ought to clean up if the idea is doing it universal. Most of these are 15 more instances of `siteSpecific("Isaac", "Ada")` that we could replaced with `SITE_TITLE_SHORT` (which I trust you can find yourself) _(and a couple of which are on admin pages to be fair)_, but then also one more specific case:
> 
>     * `StudentResources.tsx` -> "With an Ada account..."
> 
>     * _The same page also includes "I love Ada!", but it's part of a quote which I don't think we should disrupt_
> 
> 
> For interest, I did a regex search using `(?<!is)Ada`, and then manually checked each instance of "Ada" that didn't have an "is" before it. These were my findings and I'm pretty confident that'll have caught everything.

Thanks, clearly I had some holes in my regex. I'll address these.

PR #1530, opened 2025-06-16
Redesign header consistency

jacbn:
Aims to make page headers as consistent as possible across the site.


src/app/components/elements/PageMetadata.tsx
```diff
@@ -0,0 +1,79 @@
...
+                        {doc?.subtitle && <h5 className="text-theme-dark"><Markup encoding="latex">{subtitle ?? doc.subtitle}</Markup></h5>}
+                    </div>}
+                    {isAda && <EditContentButton doc={doc} />}
+                    <ActionButtons />
+                </div>
+                {children}
+            </>
+        }
+        {isPhy && <>
+            {sidebarButtonText && below['md'](deviceSize) && <SidebarButton className="my-2" buttonTitle={sidebarButtonText}/>}
```
jacbn, line 73: I don't _love_ this. I think I'd prefer another prop e.g. `showSidebar` that enables the button with default "Search and filter" text, and then `sidebarButtonText` can exist exclusively as a text override. 

Alternatively, we could put the button text in redux alongside the open/closed state? This way we can set the button text _inside the sidebar_, which feels more natural, and this button can exist as a display for that.

src/app/components/elements/PageMetadata.tsx
```diff
@@ -0,0 +1,79 @@
...
+                            </h3>
+                            {badges}
+                        </div>
+                        {doc?.subtitle && <h5 className="text-theme-dark"><Markup encoding="latex">{subtitle ?? doc.subtitle}</Markup></h5>}
+                    </div>}
+                    {isAda && <EditContentButton doc={doc} />}
+                    <ActionButtons />
+                </div>
+                {children}
+            </>
```
jacbn, line 70: Some pages (e.g. glossary) look a bit odd as they don't have a title nor any text to put in here. Generic pages also suffer from this, though these are more difficult to do anything about; still stuck on what to do (if anything) about these.

src/app/components/elements/PageMetadata.tsx
axlewin: I agree that having to pass "Search and filter" into so many different instances of this isn't ideal. I think either of your solutions are fine; the first feels more intuitive to me since it is a property of `PageMetadata`, but equally it would make sense to keep it in the same place as the open/closed state.

src/app/components/elements/PageMetadata.tsx
axlewin: For pages like the glossary, can we just get some text to put at the top? It feels strange that e.g. /concepts has some text here but /glossary doesn't (we could almost just re-use the same wording).

No idea on generic pages though :/

PR #1532, opened 2025-06-16
Redesign: Add top scrollbar wrapper for tall elements

sjd210:
For elements such as the **"My Question Decks"** list in a Table view that may be incredibly tall (e.g. when the table has 100s of elements) and may also be horizontally cut-off at certain screen widths (and so require a scrollbar), the user should not need to scroll to the bottom of the element to be able to horizontally scroll.

This change adds a wrapper that provides a top-scrollbar _as well as_ the pre-existing bottom-scrollbar by adding a `div` above the element of the same horizontal size and syncing their `scrollLeft` values. This syncing makes it slightly laggier than scrolling would ordinarily be and so should be reserved only for when its actually needed. It is therefore disabled for touchscreen since it shouldn't require a scrollbar, and uses an `enabled` prop for all other instances to determine the number of entries/size of element necessary for the top-scrollbar to appear in each case. I've picked some numbers that seem appropriate, but this may require some tuning.


src/app/components/pages/AdminContentErrors.tsx
```diff
@@ -125,31 +126,33 @@ export const AdminContentErrors = () => {
...
-                                    <tr>
-                                        <th>File</th>
-                                        <th title="Is this file published?">Published</th>
-                                        <th title="Files with critical errors will not be available on Isaac!">Critical Error</th>
-                                        <th>List of Error Messages</th>
-                                    </tr>
-                                    {errors.errorsList
-                                        .filter((error) => error.listOfErrors.reduce(errorReducer, false))
-                                        .filter(
+                            <HorizontalScroller enabled={errors.errorsList.length > 10} className="mb-3">
```
axlewin, line 129: This is using the length of the unfiltered errors list - could we handle this similarly to gameboards and move the filtering out of the table so we can use the filtered length here?

(That said, this is an admin-only page, so if there's too much overhead with getting this to work with the ShowLoading query it's probably better just to leave it).

src/app/components/pages/SetAssignments.tsx
```diff
@@ -305,7 +306,7 @@ const PhyTable = (props: SetAssignmentsTableProps) => {
                 </Col>
             </Row>
 
-            <div className="overflow-auto mt-3">
+            <HorizontalScroller enabled={boards ? boards.boards.length > 6 : false} className="mt-3">
```
axlewin, line 309: Similarly to `AdminContentErrors` - this should probably be the filtered board count.

PR #1535, opened 2025-06-19
Bootstrap spacing changes

jacbn:
I used the regex ``((gap|["'` ][mp][xytbse]?)-((xs|sm|md|lg|xl|xxl)-)?)5(?=[^0])`` to find existing instances of level 5 spacers, and replaced this with `$17` (i.e. the first matching group, so everything before "5", followed by 7 instead). 
This was intended to match all variations of strings of the format:
- `p-5`
- `pt-5`
- `pt-md-5`
- `gap-5`
- `column-gap-md-5`

but NOT those of the form:

- `-50`
- `col-5`
- `col-md-5`
- (any other `-md-5`, e.g. `block-grid-md-5`)

Where, in the above examples, `p` (padding) is replacable with `m` (margin), `t` is replaceable with `[xytbse]`, and `md` is replaceable with `xs|sm|md|lg|xl|xxl`.

The same regex found no instances with 6 in place of 5, so I am content that my first runthrough caught all of these.

There is one instance of `my-5` in the privacy policy. It being published on Physics is slightly problematic, but I think it looks fine at `2rem` so happy to leave as is for now.

### DIFF
--- a/src/IsaacAppTypes.tsx
+++ b/src/IsaacAppTypes.tsx
@@ -598,6 +598,7 @@ export interface QuestionSearchQuery {
     startIndex?: number;
     limit?: number;
     randomSeed?: number;
+    querySource: string;
 }
 
 export interface ContentSummary extends ContentSummaryDTO {
--- a/src/app/components/elements/modals/QuestionSearchModal.tsx
+++ b/src/app/components/elements/modals/QuestionSearchModal.tsx
@@ -70,7 +70,7 @@ export const QuestionSearchModal = (
     const deviceSize = useDeviceSize();
     const sublistDelimiter = " >>> ";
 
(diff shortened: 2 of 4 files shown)


jacbn:
The VRT change to Ada's homepage was intentional; see [this](https://github.com/isaacphysics/isaac-react-app/pull/1536#pullrequestreview-2942393477) comment.

sjd210 (approved):
All looks good here - I'm definitely very happy with more consistency like this! 👍 

I think its a _slight_ shame that our fixed-size classes `wf-` and `hf-` are misaligned with these, but this was already the case and they're abundant enough in content that its not an easy fix. This also makes translation between them easier than it was anyway, so I think we're good🤷‍♀️

PR #1537, opened 2025-06-19
Enable topic page accordion logging

mwtrew:
This fixes two problems to enable Topic page accordion logging:
- Accordion logging in general has been broken for some time, due to the debounced logging function returned by `debounce` not being called.
- Accordion logging relies on the current `doc` state in Redux, which Topic pages didn't set. I'm reasonably sure setting this won't cause any problems, and it's updated on any navigation.

This also moves current Topic state to using RTK Slices.


jacbn (approved):
Looks good!

PR #1550, opened 2025-06-25
Redesign sign in with microsoft

barna-isaac:


src/app/components/elements/MicrosoftSignInButton.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+// Button prompting the user to sign in via Microsoft
+export const MicrosoftSignInButton = () => {
+    const dispatch = useAppDispatch();
+
+    const logInWithMicrosoft = () => {
+        dispatch(handleProviderLoginRedirect("MICROSOFT"));
+    };
+
+    return <Button className={"position-relative"} block outline color={siteSpecific("primary", undefined)} style={{borderColor: "#545461", borderWidth: "1.5px"}} onClick={logInWithMicrosoft}>
```
jacbn, line 14:
```suggestion
    return <Button color="keyline" onClick={logInWithMicrosoft}>
```
This is the Fluent-style approach to outlined buttons we have on the redesign. Looks identical!

src/app/components/elements/MicrosoftSignInButton.tsx
```diff
@@ -0,0 +1,17 @@
...
+        <img className="google-button-logo" src={"/assets/common/logos/microsoft-logo.svg"} alt={"Microsoft logo"}/>Microsoft
```
jacbn, line 15: Should we rename `google-button-logo`? I appreciate this is not yet going on Ada, but they have their own `rpf-button-logo`. 

TBH, I think I'd prefer if these were all the same class `authenticator-logo` or something, or even just use BS classes on the logo and have no classes at all. But one each is also fine. It's more just a point for future maintainability, if a CSS class is called one thing and actually modifies several it isn't a helpful name.

src/app/components/pages/LogIn.tsx
```diff
@@ -286,11 +287,16 @@ export const LogIn = () => {
                                                 <RaspberryPiSignInButton/>
                                             </Col>
                                         </Row>}
-                                    <Row className="mb-3 justify-content-center">
+                                    <Row className={classNames("justify-content-center", siteSpecific("mb-2", "mb-3"))} >
                                         <Col sm={9}>
                                             <GoogleSignInButton/>
                                         </Col>
                                     </Row>
+                                    {isPhy && <Row className="mb-3 justify-content-center">
```
jacbn, line 295: Should this not be `mb-2` to mimic the Google one?
```suggestion
                                    {isPhy && <Row className="mb-2 justify-content-center">
```
I appreciate as it's the last one it might have needed additional spacing, but the container does that in this case.

jacbn (commented):
Looks good, just a few points about spacing and styling.

src/app/components/elements/MicrosoftSignInButton.tsx
barna-isaac: for now I'll just go ahead with inline styles. for later, have you got any suggestions for cleaning this up?

PR #1552, opened 2025-06-26
Add Ada card to homepage

axlewin:
NB this shouldn't be merged until the corresponding page is made (and the link added to this card).


src/app/components/elements/list-groups/AbstractListViewItem.tsx
```diff
@@ -167,7 +167,7 @@ export const AbstractListViewItem = ({title, icon, subject, subtitle, breadcrumb
     const cardBody =
     <div className="w-100 d-flex flex-row">
         <Col className={classNames("d-flex flex-grow-1", {"mt-3": isCard, "mb-3": isCard && !typedProps.linkTags?.length})}>
-            <div className="position-relative">
+            <div className="position-relative" style={{minWidth: "80px"}}>
```
axlewin, line 170: This isn't nice, but I can't find another way to get the Ada logo to show up without changing the dimensions of the other subject icons.

sjd210 (approved):
I found a couple of small things in review: "Isaac Physics team" should be "Isaac Science team", and the `centreLast` prop from the initial new style of card list creation was still present.

These have both now been resolved, so I'm happy for this to go in 👍

PR #1556, opened 2025-06-30
Fix issues with multiple registered contexts

axlewin:
On both sites, teachers can select multiple contexts in My Account. However, on Ada, when deciding whether content is relevant to a user, we only consider their first registered context.

For example, try logging in as a teacher on Ada and selecting GCSE and A-Level contexts (in that order). Viewing an A-Level-only page causes a "no relevant content" message to be shown, even though the content is relevant to A-Level. The A-Level accordions are also de-emphasised.

A vaguely related issue is that when adding an additional context, the dropdown label defaults to Core/Ada, but this is actually saved as All Stages/Ada.

This PR aims to address the following:
- Only show the "no relevant content" warning if there is no content on the page relevant to _any_ of the user's registered contexts.
- Only de-emphasise accordion sections if they don't match _any_ of the user's registered contexts.
- Check all of a user's registered contexts when determining which accordion stage label(s) to use.
- When adding an additional context, default the label to All Stages since this is what's saved.


src/app/components/content/IsaacAccordion.tsx
```diff
@@ -77,7 +78,7 @@ export const IsaacAccordion = ({doc}: {doc: ContentDTO}) => {
         // Handle conditional display settings
         .map(section => {
             const sectionDisplay = mergeDisplayOptions(accordionDisplay, section.display);
-            const sectionDisplaySettings = isRelevantPageContextOrIntendedAudience(section.audience, userContext, user, pageContext) 
+            const sectionDisplaySettings = !allUserContexts || allUserContexts.some(uc => isRelevantPageContextOrIntendedAudience(section.audience, uc, user, pageContext))
```
mwtrew, line 81: This breaks things for anonymous users, who have no registered contexts - if I have the user context picker set to `SQA` `N5` and go to `concepts/struct_definitions` (which has no content for that combination), nothing is de-emphasised and the labels say `GCSE`/`A Level.`

The `useUserContext` hook is aware of all the different sources of user context, and it's purpose is to reduce down to one effective context. Perhaps we should be modifying that to be able to arrive at multiple contexts, so it's still the authoritative source?

mwtrew (changes requested):

PR #1558, opened 2025-06-30
Convert Question Deck layout to ALVI list

sjd210:
We've been requested that the question decks should by styled to appear like the question finder. This is mostly a straightforward conversion to ALVIs, but with the following points:
- Content don't want wildcards to be confused with questions, so they have their tags cleared and are portrayed in a neutral context
  - (Also some wildcards are absolute links, so I've had to switch from a `<Link>` to `<a>` in the ALVI itself to accommodate 
- In the sidebar: if topic tags are empty, display field tags. If field tags are ALSO empty, don't display the "Topic" heading at all
- While waiting for markbook changes to bring through the new question states, this includes a temporary function to convert between `GameboardItemState` and `CompletionState` to make sure _some_ state is still displayed

### DIFF
--- a/src/IsaacApiTypes.tsx
+++ b/src/IsaacApiTypes.tsx
@@ -447,6 +447,7 @@ export interface ContentSummaryDTO {
     deprecated?: boolean;
     difficulty?: string;
     audience?: AudienceContext[];
+    className?: string;
 }
 
 export interface QuizSummaryDTO extends ContentSummaryDTO {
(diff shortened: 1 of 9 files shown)


sjd210:
@jacbn Cool! Any thoughts on the wildcard highlighting override? I noticed we have a section for it in `icons.scss`, but since we don't really have access to the icon itself in this context it works slightly differently to those cases and so I didn't have strong opinions on the structure (hopefully you do?)

jacbn:
Also, we've had a bit more discussion about this in the office – most wildcards as they are actually link to notes pages, so the concept icon doesn't make the most sense and we'd rather the notes icon (I believe it's `icon-contact` atm :/ might need a rename).

jacbn (approved):
We probably ought to make the wildcard SVG its own thing at some point so all the lines are the same thickness. But no-one will notice for now 😆

PR #1560, opened 2025-07-01
General improvements 14

jacbn:


src/app/components/pages/QuestionFinder.tsx
```diff
@@ -166,6 +167,10 @@ export const QuestionFinder = withRouter(({location}: RouteComponentProps) => {
                     arrayFromPossibleCsv(params.topics)
                 )
             );
+
+            const solitary = isFullyDefinedContext(pageContext) && isSingleStageContext(pageContext); 
+            setIsSolitaryStage(solitary);
+            setSearchStages(arrayFromPossibleCsv(params.stages).concat(solitary ? pageStageToSearchStage(pageContext?.stage)[0] : []) as STAGE[]);
```
axlewin, line 173:
```suggestion
            if (solitary) setSearchStages(arrayFromPossibleCsv(params.stages).concat(pageStageToSearchStage(pageContext?.stage)) as STAGE[]);
```
This is removing the account stages filters on the generic QF, and defaulting the 11-14 QF to y7&8 only (similarly for A-Level/Further A). Could we use something like this to preserve some of the original behaviour?

PR #1562, opened 2025-07-02
Chevron & arrow iconography 

jacbn:
Attempts to make the meaning of chevrons and arrows consistent across the site.

- A closed chevron pointing right, ">", indicates an expandable region. This includes tabs, collapsible lists, teacher notes, etc. They should be toggleable; they maintain their state until interacted with again.
- A closed chevron pointing down, "v", indicates a dropdown. This includes the buttons in the header. Any interaction anywhere else on the page will close them.
- Arrows, "->", "<-", etc, indicate links. They tend only to exist on buttons.

This PR also attempts to clean up the mess of various SVG icons we have that represent the same thing (chevrons). We now only have one relevant chevron SVG, `chevron-right`, with various rotations via `icon-chevron-{up|down|left|right}`. They can still be rotated correctly in animations, provided the animation is over `transition`, not `rotate`.


axlewin:
Nice to have some consistency here. This all looks good across ~physics~ science, but there's a few regressions on Ada (mostly Firefox issues):

- `/tests` is missing a chevron on the filter dropdown:
image
- Also on `/tests`, the arrows in the table are no longer centered (Firefox):
image
- The vertical alignment's off on `/set_tests` (Firefox):
image
- The topic filters on the question search modal are missing chevrons:
image
- `TopicSummaryLink`s have a vertical line before the chevron:
image

A couple of more general thoughts:
- `TopicSummaryLink`s on Ada now look exactly like concept accordions, which is confusing when they behave differently. Since they're links the chevrons ought to be arrows for consistency? But I'm not sure if that would look right either.
- Cloze dropdowns have a right-chevron when closed - since any interaction elsewhere closes these dropdowns, shouldn't these still be down-chevrons?

jacbn:
I think I've fixed all the Firefox issues. A few points:

> TopicSummaryLinks have a vertical line before the chevron:

This one's intentional. It has always had one. We could remove it, but it is now one of the last indicators that this is a topic view and not an accordion. I suppose the line was supposed to represent a link, as you wouldn't expect a table-styled element to expand.

> TopicSummaryLinks on Ada now look exactly like concept accordions, which is confusing when they behave differently. Since they're links the chevrons ought to be arrows for consistency? But I'm not sure if that would look right either.

The "arrows for links" consistency was just for Physics, as Ada have a few external links too that aren't built with `AffixButtons`. I do agree that it ought to be different, but I think the vertical bar might be enough?

> Cloze dropdowns have a right-chevron when closed - since any interaction elsewhere closes these dropdowns, shouldn't these still be down-chevrons?

Fixed. Nice spot :)

axlewin (approved):

PR #1571, opened 2025-07-08
General improvements 15

jacbn:
Larger changes:
- Make teacher & tutor features pages use styling consistent with the rest of the site
- Add test to ensure mobile navbar entries all close the navbar when clicked
- Change spacing around:
  - Display-mode KaTeX regions to allow more space for cloze regions before they visibly clip
  - Non-question content (usually text) inside accordions on a question (visible on e.g. /q/graph_interpreting_sqw_2e)
- Make all header icons the same size (75px x 75px)


src/app/components/pages/TeacherFeatures.tsx
```diff
@@ -19,7 +20,7 @@ export const TeacherFeatureCard = (props: TeacherFeatureCardProps) => {
     return <Link to={url} className="h-100 w-100" aria-label={title} aria-disabled={disabled} style={{textDecoration: "none"}}>
         <Card className={classNames("p-3 teacher-features h-100", {"disabled": disabled})}>
             <div className="d-flex justify-content-between">
-                <h5 className="mt-3 me-2">{title}</h5>
+                <h4 className="mt-2 mb-0 me-2">{title}</h4>
```
sjd210, line 23: I think this looks _slightly_ more odd than before (probably because most pages formatted like this have sidebars), but for the sake of consistency and that this page ought to have a share button I don't really mind.

Sidenote (perhaps something for later): Do we have a particular reason for TeacherFeatures.tsx and TutorFeatures.tsx being separate? It would be nice to de-duplicate logic like this and repeated cards between the pages.

sjd210 (approved):
Generally lgtm 👍 I'd merge it if we weren't waiting for release, but in the meantime I do have a quick thought about teacher/tutor features.

PR #1574, opened 2025-07-10
Allow for multiple registered contexts when deciding if content is relevant

axlewin:
(Ada): If a teacher has multiple registered contexts set, they should see content highlighted as relevant to them if it is relevant to _any_ of their contexts (we currently just check the first).

### DIFF
--- a/src/app/services/userViewingContext.ts
+++ b/src/app/services/userViewingContext.ts
@@ -43,16 +43,14 @@ import {useContext} from "react";
 import {Immutable} from "immer";
 
 export interface UseUserContextReturnType {
-    stage: STAGE;
+    contexts: UserContext[];
     setStage: (stage: STAGE) => void;
-    examBoard: EXAM_BOARD;
     setExamBoard: (stage: EXAM_BOARD) => void;
     isFixedContext: boolean;
     setFixedContext: (isFixedContext: boolean) => void;
     explanation: {stage: CONTEXT_SOURCE, examBoard: CONTEXT_SOURCE};
     hasDefaultPreferences: boolean;
 }
-
 
 export interface GameboardAndPathInfo {
     boardIdFromDTO?: string;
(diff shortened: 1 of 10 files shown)


src/app/services/userViewingContext.ts
```diff
@@ -43,17 +43,15 @@ import {useContext} from "react";
 import {Immutable} from "immer";
 
 export interface UseUserContextReturnType {
-    stage: STAGE;
+    contexts: UserContext[];
```
axlewin, line 46: This doesn't feel neat, but we can't just convert the existing stage/exam board properties to arrays because we need to keep stages and exam boards paired up correctly (e.g. if a user has AQA GCSE & OCR A-Level set). We also only need one explanation (etc), so having a list of contexts _inside_ `UseUserContextReturnType` instead of having the hook return a list of `UseUserContextReturnType` seems sensible?

src/app/services/userViewingContext.ts
```diff
@@ -88,46 +86,53 @@ export function useUserViewingContext(): UseUserContextReturnType {
...
     } else {
         examBoard = EXAM_BOARD_DEFAULT_OPTION;
         explanation.examBoard = CONTEXT_SOURCE.DEFAULT;
     }
 
+    // If there is no transient context, use registered context(s) if possible
+    if ((isPhy || explanation.examBoard === CONTEXT_SOURCE.DEFAULT) && explanation.stage === CONTEXT_SOURCE.DEFAULT && registeredContexts?.length) {
+        const registeredContextsWithExamBoards = isPhy ? registeredContexts.map(rc => ({stage: rc.stage, examBoard: EXAM_BOARD.ALL})) : registeredContexts;
+        contexts.push(...registeredContextsWithExamBoards);
+        explanation.stage = CONTEXT_SOURCE.REGISTERED;
```
axlewin, line 125: Because we need to keep the stages/exam boards paired up for multiple registered contexts, we can't keep this in the original `else if` and push stages/exam boards as we encounter them.

axlewin:
@mwtrew - I've looked over [your suggestion](https://github.com/isaacphysics/isaac-react-app/tree/uc-refactor-suggestion) and agree it's a much nicer approach. There are a few test failures on that branch right now, but once that's sorted I'm happy for those changes to be merged into this. The only other thing I noticed is that that branch is hardcoding `EXAM_BOARD.ALL` instead of using `EXAM_BOARD_DEFAULT_OPTION`, and on a similar note the default explanation for physics exam boards is `DEFAULT` instead of `NOT_IMPLEMENTED` - I think the tests have also caught the latter, though.

PR #1575, opened 2025-07-10
Empty states for Set Assignments and Markbook

mwtrew:
This PR implements the Ada CS empty state designs (see card) for Set Assignments and the Markbook. 

On Isaac Science, there are also a couple of changes for consistency: 
* Using an alert for the "no groups" empty state on Set Assignments, rather than some text.
* Using an empty state for "no groups" similar to what we currently have for "no assignments" in the Markbook, rather than some text.

Best reviewed with whitespace changes hidden. I ran an automatic reformat of Set Assignments hence the noisy diff (sorry).


sjd210 (changes requested):
The Markbook empty states look great for both Ada and Isaac! I do however have a couple of small nits with the Set Assignments change.

PR #1589, opened 2025-07-24
Test parent reselection on QuestionFinder, Concepts

barna-isaac:
This PR contains a set of changes aimed at making our lives easier.
- It introduces tests for the selection behavior of filter buttons on the QuestionFinder page, and introduces the same tests for the Concepts page, which didn't have any component tests so far.
- the change to `store.tsx` disables automatic redux logging during test execution, as this produced a wall of irrelevant outputs.
- the change to `SidebarLayout.tsx` is this PR's only change to live code. It eliminates ``Warning: Cannot update a component (`HeaderPhy`) while rendering a different component (`MainContent`)`` warnings, which were shown during test execution as well as on the browser console in dev mode.
- by mocking `navigation` in `renderTestEnvironment`, we silence `Not implemented: navigation (except hash changes)` warnings (we were already doing this, in `SessionCookieExpired.test.ts`)


src/mocks/filters.ts
```diff
@@ -0,0 +1,79 @@
+import { isPhy } from "../app/services";
+import { clickOn } from "../test/testUtils";
+import { screen } from "@testing-library/react";
+import _ from "lodash";
```
jacbn, line 4: I don't think it really matters as I'm not sure this file gets bundled, but importing from lodash like this imports the whole (rather large) library. You can import individual functions via 
```suggestion
import zipWith from "lodash/zipWith";
```
and just use that function below. Even if it isn't necessary, it's a useful habit to get into!

src/mocks/filters.ts
```diff
@@ -0,0 +1,79 @@
...
+    return Promise.all(filters.map(filter => toggleFilter(filter)));
+};
+
+const expectClasses = (labels: Array<string | RegExp>, classNames: string[]) => {
+    return _.zipWith(labels, classNames, expectClass);
+};
+
+export const expectClass = (label: string | RegExp, className: string) => {
+    const element = screen.queryByLabelText(label);
+    if (className === 'hidden') {
```
jacbn, line 66: Why does this have different behaviour for one specific class? What if it's hidden via another class like "d-none"? 

Also, should this 'is visible in document' interaction not be a different assertion? If not, why here?

src/test/testUtils.tsx
```diff
@@ -48,6 +49,10 @@ interface RenderTestEnvironmentOptions {
 // When called, the Redux store will be cleaned completely, and other the MSW server handlers will be reset to
 // defaults (those in handlers.ts).
 export const renderTestEnvironment = (options?: RenderTestEnvironmentOptions) => {
+    // mock redirectTo action because jsdom does not support href updates
+    jest.spyOn(Actions, "redirectTo");
+    (redirectTo as jest.Mock).mockImplementation(() => true);
```
jacbn, line 54: Putting this in every test feels like it will cause more problems than it solves; the cookie tests you were using as a basis for this restored the behaviour immediately afterwards, perhaps expecting this to be problematic if used elsewhere. 

The tests run fine for me without this change. What were you trying to fix?

PR #1595, opened 2025-07-28
log federated sign-in attempts

barna-isaac:
I'd like to see the drop off rate for sign-in with Microsoft (as well as sign-in with Google, for comparison). I'm worried too many users might be giving up / are unable to use sign-in with Microsoft because of the lack of verification.

I've decided to pool sign-ins from the log-in page as well as attempts at linking an account as `sign-in attempted`, primarily because I couldn't find a way to differentiate between the two sources for the `sign-in failed` event.

For details on viewing results, see the trello card.

### DIFF
--- a/src/app/state/actions/index.tsx
+++ b/src/app/state/actions/index.tsx
@@ -67,7 +67,7 @@ export function extractMessage(e: Error) {
     return API_REQUEST_FAILURE_MESSAGE;
 }
 
-export function fetchErrorFromParameters(parameters: string): { error?: string, errorDescription?: string } {
+export function fetchErrorFromParameters(parameters: string): { error?: string, errorDescription?: string, parseError?: string } {
     try {
         const parsed = new URLSearchParams(parameters);
         const [error, errorDescription] = [parsed.get('error'), parsed.get('error_description')];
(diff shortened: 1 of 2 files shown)


src/app/state/actions/index.tsx
```diff
@@ -69,6 +67,19 @@ export function extractMessage(e: Error) {
...
+export function fetchErrorFromParameters(parameters: string): { error?: string, errorDescription?: string } {
+    try {
+        const parsed = new URLSearchParams(parameters);
+        const [error, errorDescription] = [parsed.get('error'), parsed.get('error_description')];
+        return {
+            ...(null !== error && { error }),
+            ...(null !== errorDescription && { errorDescription })
+        };
+    } catch {
+        return {};
```
jacbn, line 79: Should we add something to this case so we can differentiate between the case where `error` and `errorDescription` are both null, and the case of us not being able to parse the errors?

jacbn (commented):
Happy with this, one comment about making a distinction between two cases but this may be a non-issue (e.g. if we are certain the errors will parse).

src/app/state/actions/index.tsx
barna-isaac: @jacbn I did it (and also added some simple unit tests), can you please review it again?

jacbn (approved):
Looks good! :)

PR #1600, opened 2025-07-29
Rework "Help" positioning

jacbn:
Several pages have a "Help" to the right of the supertitle, which (at least on Isaac) usually opens a modal with a help video. We would like to move this "Help" down into the ActionButtons in the PageMetadata, without affecting Ada.

In doing so, I have come up with a much nicer-looking approach to displaying the ActionButtons alongside text via `float-right`. I have changed most of the internal (i.e. non-content) pages to use this.


src/app/components/elements/PageTitle.tsx
```diff
@@ -150,7 +150,7 @@ export const PageTitle = ({currentPageTitle, displayTitleOverride, subTitle, dis
             <meta property="og:title" content={currentPageTitle} />
         </Helmet>
         {audienceViews && <AudienceViewer audienceViews={audienceViews} />}
-        {help && !showModal && <React.Fragment>
+        {isAda && help && !showModal && <React.Fragment>
```
sjd210, line 153: Removing `modalId` means we should also remove `showModal`
```suggestion
        {isAda && help && <React.Fragment>
```

We should also remove the original showModal definition (L104), this page's HelpModal (L119-134) and the Fragment below this (L157-159). None of that code has any use here anymore.

sjd210 (changes requested):
Assuming that its intentional to remove the help text that _isn't_ a modal(?) (e.g. on the Question Finder), this looks good.

Before merging this however, there are a few points of now-dead code that we may as well clean up:
- All mention of `modalId` in TitleAndBreadcrumb / PageTitle (it is currently only used as a prop in some defunct book index pages which are now content controlled so we can remove anyway)
- This piece of [css](https://github.com/isaacphysics/isaac-react-app/blob/05da495d6e87ed0aabb518acd1c001d54ec7878f/src/scss/phy/page-title.scss#L74-L76)
- (Not dead code, but the alignment of the "close" button on modals was annoying me, so I've just amended that here too. It was the same on both sites)

jacbn:
Good spots, thanks. And yes, we were intending to remove all the non-modal helps too. Content preferred having the instructions in a page fragment at the top of the page, always visible.

I think I had assumed there were still modals on Ada, but didn't really check this... 🤦

PR #1608, opened 2025-08-04
Mobile sidebar closing unexpectedly

jacbn:
Restructures the `pageLoad` component to only run changes on initial page load and true path changes, not for query params in e.g. QF. 

This logic already existed in `scrollTop`, so I have mostly just moved this up one layer.


barna-isaac (approved):
Looks great, thanks for the fix!

PR #1609, opened 2025-08-04
General improvements 19

axlewin:
Mostly small visual fixes, including some Ada regressions.

The main behavioural change is related to user contexts on sci. Previously, if a user had multiple registered stages, we considered only the first when determining the page context and therefore when highlighting/ordering accordion sections any additional registerered stages were ignored.

This PR changes the context derivation logic to only use the user's registered context(s) to set the page context if there is exactly one match between the registered contexts and the page audience (e.g. user has GCSE & A-Level, page is A-Level & university).

If a user has non-consecutive registered stages (e.g. Y7&8 and university) the intermediate stages are not collapsed into "Additional Learning Stages", but additional stages before/after _all_ registered stages should still collapse as normal.


jacbn (approved):
Looks good!

PR #1614, opened 2025-08-05
Improve support for multiple registered contexts (sci)

axlewin:
(Isaac Science): If a user (tutor+) has multiple registered stages, we consider only the first when determining the page context and therefore when highlighting/ordering accordion sections any additional registered stages are ignored. For example, if a user has GCSE & A-Level contexts set, a GCSE & A-Level page would only have the GCSE accordions highlighted.

This PR changes the context derivation logic to only use the user's registered context(s) to set the page context if there is exactly one match between the registered contexts and the page audience (e.g. user has GCSE & A-Level, page is A-Level & university). Otherwise, don't set the page context, allowing the user contexts to be used directly instead.


sjd210 (changes requested):
(amending review to "Request changes")

axlewin:
> Basically a user on Ada might be able to have multiple of the same stage, may have multiple exam boards but only one of which is considered, and the specific stage/exam board combination sometimes matters. This is far too complicated and I think we should just go back to using the first context as a default on Ada instead rather than put effort into sorting it out. It's worked fine so far 🤷‍♀️

The whole multiple stage/exam board combinations issue does need fixing at some point (there are lots of edge cases and it causes some strange behaviour on the account settings page). I agree that displaying only the first context used is probably less confusing in the interim, though. It does mean that the context picker text doesn't align with which content is actually highlighted, but since this will only affect staff anyway I'm not too worried.

sjd210 (approved):
Hrm, you're right. The Ada context picker situation is worse than I'd even been realising :/ That could definitely do with some tweaking - I think worth making a card.

Anyway for _this_ PR focused on sci, all looks good! 👍

PR #1628, opened 2025-08-11
Ada UX: Improve form feedback

jacbn:
Rework the Set Quiz modal to use the new preferred form format, whereby the submission button is always enabled, but clicking it highlights any invalid fields in red, preventing submission.


sjd210 (approved):
Looks good 👍 Of course there are other modals that should also be due this treatment now we're doing it here (particularly Set Assignments), but we can handle those in time.

PR #1631, opened 2025-08-12
Switch ">25% incorrect" to "<25% correct" in the markbook

sjd210:
On Sci stop any use of attemptedness to calculate correctness (except for an entirely unattempted question) by switching ">25% incorrect" with "<25% correct" in the legend and associated calculation.

 Two other small things to consider here: 
\- For clarity and consistency, "<75% correct" was renamed to "≥25% correct". This doesn't change any calculation.
\- On the attempted view on Sci (but not on Ada), we should show the `not-attempted` state at <25% attempted according to the legend, but we were showing it for only 0% attempted which does not have a distinct state. This was a slight discrepancy I've fixed the calculation for. (I'm not sure I personally agree this _should_ be how the cutoffs work but since that's what it says, we should be consistent)

### DIFF
--- a/src/app/components/pages/AssignmentProgressIndividual.tsx
+++ b/src/app/components/pages/AssignmentProgressIndividual.tsx
@@ -70,7 +70,7 @@ export function markClassesInternal(attemptedOrCorrect: "ATTEMPTED" | "CORRECT",
             return "revoked";
         } else if (status && isQuestionFullyAttempted(status) || correctParts + incorrectParts === totalParts) {
             return "fully-attempted";
-        } else if (status === CompletionState.NOT_ATTEMPTED || (correctParts + incorrectParts) / totalParts < (1 - passMark)) {
+        } else if (status === CompletionState.NOT_ATTEMPTED || siteSpecific((correctParts + incorrectParts) / totalParts < (1 - passMark), correctParts + incorrectParts === 0)) {
             return "not-attempted";
         } else if ((correctParts + incorrectParts) / totalParts >= passMark) {
             return "passed";


jacbn (approved):
> I'm not sure I personally agree this should be how the cutoffs work but since that's what it says, we should be consistent

I think the logic here was if someone has attempted only 10% of the work, as a teacher you would want to treat them the same as someone who has done 0%, which I can understand. It feels like a 0% completion would be nice to have, but on thinking about it it's not actually that useful provided the <25% is around.

---

I'm fine with the technical changes here, the issue I had in the tech meet was less about what we've done here and more about whether this is actually what teachers want. We no longer have _any_ display of incorrectness; I agree that this is better for consistency, but we have just removed a feature that I can see having been useful?

Before this change, 
- a student who attempted everything but got almost all of it wrong would show as red in correct view and yellow in attempted view;
- a student who attempted nothing would show as white in correct view and white in attempted view.

Now, both show as white in correct. This is what we wanted, I know! I just don't see it being very useful. To work out if a student got lots wrong so as to give them help, you need to find them in **both** views of the table and compare the colours; for a big class, or indeed e.g. STEM SMART, this is a pain owing to having to scroll back up to the top etc – now imagine doing that for 5 or 6 students.

The part of me that likes consistency is very happy we split the table out into attempted/correct, and thinks this change is great. But I think what a teacher wants is not having to toggle things and literally _work out_ how their students did, but rather they want one single view that can tell them all 3 of "this student got a lot correct", "this student got a lot wrong", and "this student did not attempt the work" in one go. And, currently, the only single view that could do all of that was the correct view _before this change_, as we still had an incorrectness metric there.

I don't think this is the place to suggest anything (not that I have something in mind that could magically fix all of this), and either way I am really not keen on re-redesigning the markbook. This PR does make the two views we have consistent, and I think it is better for it, so I'm happy to merge. I am just acutely aware that by removing the _in_correct calculation we did have, it muddies the distinction between getting lots wrong and not attempting the work, which seems to go against making teachers' lives easier.

PR #1648, opened 2025-08-14
Update search shortcuts

axlewin:
- Add some new shortcuts based on common searches
- Remove some shortcuts that are deprecated or unnecessary because they're included in normal search results


axlewin:
@barna-isaac 

Good catch on the regex (and the export!) - both changes look good to me 👍 

In terms of your other suggestions:

- I did consider adding some different spellings such as alevel, as well as other ways users might describe stages (year 7, 11-14, AS...) but I think this would overly complicate things. The only two we really need are GCSE/A Level (with no hyphen - I can show you the data for different spellings in-person if you're interested); the others just happen to also work easily with the regex so I included them too.

- I also thought about what we can show for users who search for just a stage. I think you're probably right that all the relevant landing pages for the stage would be the most sensible option, but showing (up to) 4 different pages here feels inefficient when we used to have a single landing page for each stage :/ We do, however, have a nice landing page for each subject which links to all of the stage-specific landing pages for that subject, so the reverse isn't necessary.

PR #1655, opened 2025-08-19
Refactor SetAssignments to improve form feedback

sjd210:
Following the new style from [SetQuizzes](https://github.com/isaacphysics/isaac-react-app/pull/1628) for consistency, the SetAssignments modal now allows the submit button to be pressed at any time, and only gives feedback after submission has been attempted.

It also separates SetAssignmentsModal into its own file for cleanliness, and changes some tests to make them pass and increase clarity.

Finally, if the due date has not been edited in this instance of the modal, changing the start date will set the due date to `StartDate + 6 Days` rather than immediately invalidating that pre-selected due date.


mwtrew (changes requested):
Looks good. I've made some small suggestions for clarity and consistency with other forms. I realise some of these problems are pre-existing, but it would be good to get them fixed anyway.

mwtrew (approved):

PR #1656, opened 2025-08-19
CS icon position adjustment

jacbn:


barna-isaac:
The VRT's look better, and also fixes the issue on the Markbook which triggered this PR.

To me these kind of CSS changes are very difficult to review, and any effort we'd put into trying to control these changes is better spent on extending VRT coverage. I'm a little comforted knowing this change can only affect Ada.

barna-isaac (approved):

PR #1659, opened 2025-08-20
Ada UX: Set assignments modal

jacbn:
Adjusts title of Set Assignments modal to create a clearer user flow when creating then setting an assignment. Also slightly modifies the spacing around ActiveModals (Phy+Ada) to better meet designs and center content.


sjd210 (approved):
Looks good 👍 Just had to update the modal title in tests to align with this edit

PR #1661, opened 2025-08-21
Ada homepage changes

mwtrew:
Ready to review, but the SVGs in use here have large embedded bitmaps. Not ready to merge until I get replacements from design. I'll also hold off updating VRTs until then, but they look fine to me.


src/app/components/site/cs/HomepageCS.tsx
```diff
@@ -1,233 +1,201 @@
...
 
             <section id="what-resources">
                 <Container className={"homepage-padding mw-1600"}>
                     <TextBlock md={8}>
-                        <h2 className={"font-size-1-75 mb-4"}>Teaching outside the UK?</h2>
-                        <p>We&apos;ve organised our learning resources by prior knowledge and age group to make them easy to adapt for curricula around the world.</p>
+                        <h2 className={"font-size-1-75 mb-4"}>Exam board alignment</h2>
+                        <p>We&apos;ve organised our learning resources so they can be easily used, wherever you are.</p>
                     </TextBlock>
                     <ColumnSlice>
```
jacbn, line 147: This changes layout at `md` in the designs:
```suggestion
                    <ColumnSlice breakpoint="md">
```

src/app/components/site/cs/HomepageCS.tsx
```diff
@@ -1,233 +1,201 @@
...
-                            </ul>
-                            <div>
-                                <Button className="mt-3 me-3" tag={Link} to="/students" color="solid">Student Resources</Button>
-                                <Button className="mt-3" tag={Link} to="/teachers" color="solid">Teacher Resources</Button>
-                            </div>
-                        </Col>
-                    </Row>
+                <Container className={"mw-1600 homepage-padding-x"} fluid>
+                    <Card id={"cta-features-card"} className={"icon-card p-5"}>
+                        <Row className={"justify-content-center gy-5"}>
```
jacbn, line 54: These are bold in the designs:
```suggestion
                        <Row className={"justify-content-center gy-5 fw-bold"}>
```

jacbn (commented):
A couple of very minor points I have suggested below, and then:
- This doesn't have the news slice and newsletter slices at the bottom – I feel like I remember this being brought up somewhere, but worth checking: was this intentional?
- The layout of IconCards in the designs is in contradiction with those in the component library, namely the position of the icon when >480px (or maybe >416px, it is inconsistent in the component library too 🙄). I might bring this up with the designers but given it is a reusable component I don't think we need to worry about this for now.

Looks really good overall though!

mwtrew:
> A couple of very minor points I have suggested below, and then:
> 
>     * This doesn't have the news slice and newsletter slices at the bottom – I feel like I remember this being brought up somewhere, but worth checking: was this intentional?
> 
>     * The layout of IconCards in the designs is in contradiction with those in the component library, namely the position of the icon when >480px (or maybe >416px, it is inconsistent in the component library too 🙄). I might bring this up with the designers but given it is a reusable component I don't think we need to worry about this for now.
> 
> 
> Looks really good overall though!

Thanks for the suggestions! 

The news/newsletter slices at the bottom are there, but the former is conditional on content being loaded, and the latter on the user either being logged out or logged in and not having that preference set already. That might be why you didn't see them?

PR #1667, opened 2025-08-21
Allow empty audience views to be displayed on the question builder modal

sjd210:
The main change here is just the first commit, converting the `[]` to a `[{stage: undefined, difficulty: undefined}] AudienceContext[]`.

The rest of this is some mild cleanup, including using `classNames` where appropriate and adding some vertical spacing to the gameboard builder filters - particularly for Ada. These aren't important, so if there are any problems with them I'm happy to abandon them.


jacbn (approved):
LGTM!

PR #1678, opened 2025-08-27
re-eanble VRTs

barna-isaac:
In [7c6d47504d21ba3f7750a20cdcb5f77ce78c2d23], we've disabled two VRTs for the `AssignmentProgress` page. This brings those back, using what I think is the minimal possible adjustment.

I still don't think this is the best way, though, because as we make changes to component interfaces, we'll need to keep adjusting the VRT's. It'd be best if we only had to adjust the VRT's after a functional change (so, if the URL of a page changes, it makes sense that we also need to change the tests, but when only the code changes and the app remains the same, the tests should not need adjusting). I hope to once achieve this by always mounting the `<IsaacApp>` component from the VRTs, but until then, this at least brings back those tests.

### DIFF
--- a/src/mocks/data.ts
+++ b/src/mocks/data.ts
@@ -1,4 +1,4 @@
-import {siteSpecific} from "../app/services";
+import {isDefined, siteSpecific} from "../app/services";
 import {FEATURED_NEWS_TAG} from "../app/services";
 import {DAYS_AGO, SOME_FIXED_FUTURE_DATE} from "../test/dateUtils";
 import {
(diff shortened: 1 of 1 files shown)


jacbn:
That was to fix a bug where the first case can have undefined properties if the user in question has revoked access. It's [already live](https://github.com/isaacphysics/isaac-react-app/commit/8a1aa0a5fc2fdd5ad0603e58a2cdeebdc132c993) so I wouldn't worry about it!

PR #1687, opened 2025-08-28
Re-enable Ada events

jacbn:
Restores `/events` to its former glory on Ada. Reuses the `<AdaCard/>` component for styling, for which I was getting angry at some prior design choices I made so there are some commits to adjust where the spacing between multiple cards is located.


sjd210 (approved):
LGTM 👍 (Although you already said that you weren't accepting criticism anyway, so I suppose it wouldn't matter either way)

PR #1693, opened 2025-09-02
Reunite redundant PhyVersion/AdaVersion splits from redesign

sjd210:
This is mostly pretty straightforwardly taking pieces of shared code that had been split for the redesign and merging them into one `const`. Similar to https://github.com/isaacphysics/isaac-react-app/pull/1689 but various smaller cases instead.

The only things that should have outwardly changed are:
- QuizButton labels e.g. "Continue Test" are now all sentence case on Sci to be consistent with other buttons
- Some other random titles/buttons are now also set to sentence case e.g. "Account Deletion" -> "Account deletion"
- Dates on QuizItems (cards) are now all bold on Ada rather than only the due date. Not sure if it was intentional or not in the first place, but I think this is clearer.

Everything else ought to be code-only cleanup.


jacbn (approved):
Looks great! Switching the themes around has reminded me that another thing that would be nice is to start aligning themes, button styling, etc – it would be great to switch/rename all the $gray-xxx classes to something logical, $primary should probably be the theme colour on sci, ...

For another day!

PR #1700, opened 2025-09-03
2nd edition maths book

jsharkey13:
Add the 2nd Edition of the A-Level maths book to My Progress. Also make book titles consistent on this page.

Remove the 1st Edition from the books modal when adding new boards from books. This is so that the Physics Skills 14 book can also be added to the main books list (so that all book data is in one place and consistent titles are used) but be hidden from view everywhere.


sjd210:
I've just separately checked that the 490 number makes sense, and agree that it does 👍 LGTM

sjd210 (approved):
I've just separately checked that the 490 number makes sense, and I agree that it does 👍 LGTM

PR #1706, opened 2025-09-04
Update news-related copy on Ada

axlewin:
- Update text
- Move news section into component to keep text changes consistent across pages
  - Causes some minor spacing changes - notably the homepage newsletter card is now the same width as the other full-width cards on this page


src/app/components/elements/AdaNewsSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+import classNames from "classnames";
+
+interface AdaNewsSectionProps {
+    news?: IsaacPodDTO[];
+    showNewsletterPrompts: boolean;
+    setLinkedSetting: (targetId: string) => void;
+    isHomepage?: boolean;
+}
+
+export const AdaNewsSection = ({news, showNewsletterPrompts, setLinkedSetting, isHomepage}: AdaNewsSectionProps) => {
```
mwtrew, line 16: I don't think there's much point in these being props, since they're not used in the parent components otherwise. We could just use the relevant hooks in this component now.

src/app/components/elements/AdaNewsSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const deviceSize = useDeviceSize();
+    return <Container className={isHomepage ? "homepage-padding mw-1600" : "overview-padding mw-1600"}>
```
mwtrew, line 18: Perhaps a `containerClassname: string` prop would be more flexible?

axlewin: Hmm, I'm not sure what the neatest approach is here. I've pushed a change to use `containerClassName` for the main container, but I can't see an easy way to remove the other usages of `isHomepage`.

PR #1709, opened 2025-09-05
Fix infinite builder rerender caused by empty context

jacbn:
Fixes an issue caused by context code, where a new 'default context' was being created on each call to `useUserViewingContext()` if there was no context defined in Redux (i.e. no context set in account settings). This was causing `useEffect` loops that were dependent on this context (or parts of it) to fire every render.

This was only really noticeable on the gameboard builder since this makes an API call, but this fix should also prevent infinite rerenders on any `useEffect` on the context.


sjd210 (approved):
Good find! This all makes sense to me and I can confirm that the behaviour is working properly on my machine now (it was quite entertaining to watch the pre-made gameboards being infinitely created in a random order each time pre-fix).

Let's get this in quickly

PR #1710, opened 2025-09-08
Stop displaying old assignments as active in the teacher/student dashboards

sjd210:
A couple of related things here:
- On the teacher dashboard, "View scheduled work" should only show assignments with a due date in the future (as from a teacher perspective, there is no concept of "overdue" for an entire assignment). Very old assignments from before due dates were required were being displayed here, but now are filtered out.
- On the student dashboard, "Complete current work" should show incomplete assignments that are due _or_ overdue however until now this has included assignments very long overdue. This puts a cap of two weeks overdue for assignments to be displayed as "current" in this menu (and adds a link to "View older assignments" when this makes the tab empty)

- In the assignment status logic, we calculated an old `noDueDateButRecent` value for assignments without due dates. This is now completely irrelevant for newly created assignments.
  - This has a niche interaction with old assignments created with a start date ~1+ years in the future (and that start date is now within the last four weeks of the current date) but no due date - these are now being categorised as `inProgressOld`. They were already being treated inconsistently and can still be reached this way, so I think this is a fine classification for these edge-case assignments.
  - Also to note, the page fragment on `/assignments` refers to this value and so is outdated (for both main and this branch). Ignoring the edge-case above, **To do (recent)** only displays non-overdue assignments, and **To do (older)** displays overdue assignments or old assignments without due dates. The assignment being four weeks from start date makes no difference and so might be confusing. This should probably be discussed with content.

### DIFF
--- a/src/test/services/assignments.test.ts
+++ b/src/test/services/assignments.test.ts
@@ -68,6 +68,9 @@ const assignments: AssignmentDTO[] = [assignmentA, assignmentB, assignmentC, ass
 const nextWeek = new Date();
 nextWeek.setDate(nextWeek.getDate() + 7);
 
+const twoWeeksAway = new Date();
+twoWeeksAway.setDate(twoWeeksAway.getDate() + 14);
+
 const tomorrow = new Date();
 tomorrow.setDate(tomorrow.getDate() + 1);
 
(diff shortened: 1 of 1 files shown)


mwtrew (approved):
> The assignment being four weeks from start date makes no difference and so might be confusing. This should probably be discussed with content.

Looks good. Yep, it would be good to correct the fragment to coincide with this.

PR #1717, opened 2025-09-10
Quick fix for Microsoft authentication

barna-isaac:
I've forgotten to put a single instance of the provider's name behind an "if" statement. This fixes that.

Screenshot 2025-09-10 at 09 39 37


mwtrew (approved):
Looks good (pending check results).

PR #1719, opened 2025-09-11
Add warning modal for attempting to archive groups

sjd210:
This is originally an Ada request for a clear explanation of what archiving does and how it can be recovered, but I think it makes sense to put on Sci too.

This could've just been in a `confirm()` for simplicity like are used for many other actions on the page, but I don't think this is clear and is something we should eventually change for these other actions too.


sjd210:
Both foolish mistakes made in a rush 🤦‍♀️ - thanks for spotting them.
If this all compiles and tests correctly I'll merge it now

PR #1727, opened 2025-09-12
Add lint step to CI

barna-isaac:
I'd like to have some kind of automation for running linter rules. This will ensure we don't leave accidental linter violations in our PR's (like, most recently, in https://github.com/isaacphysics/isaac-react-app/pull/1721, where I fixed these as part of the review). While some linter violations are harmless, other violations mask real issues:
- either during automated tests, like in https://github.com/isaacphysics/isaac-react-app/commit/43e048d71b429c08d1cfac954b36f2bced21503e, where not awaiting promise-returning functions has destabilized the test suite (I introduced the mistake),
- or in https://github.com/isaacphysics/isaac-react-app/pull/1577, where ignoring react's linter rules has caused the equation editor to be unusable in Safari on iOS

I'm hoping that automatically running linter rules as part of the build will:
- save reviewers time so they don't need to suggest these trivial changes, and
- save developers time by preventing violations that cause bugs.

For even faster feedback, we could also add a precommit hook for running linter rules. I also support the implementation of a precommit hook, but the execution of this will always depend on a contributor's local setup. I wanted to have an automation in place that will execute for all commits, regardless of local setup.

Because we currently have many linter failures, I added the `eslint-plugin-diff` for an incremental approach. When the lint rules are executed in CI, only the diffs for the most recent commit (on a PR, changes from the PR) are inspected. This prevents the introduction of new violations, without prescribing that all linter rules be fixed at once (we currently have hundreds of violations). It also makes it easy to introduce new linter rules, without needing all violations to be fixed at the time a rule is introduced. In the future, I'd like to use standardised eslint types sets, such as `plugin:@typescript-eslint/recommended-type-checked`, described in https://typescript-eslint.io/getting-started/typed-linting/. This will require the introduction of many new rules, not all of which will be auto-fixable.

I've used commit https://github.com/isaacphysics/isaac-react-app/pull/1727/commits/ea0b8b38c02affa506ad3c9761092f40b24ab167 to introduce a linter violation and test out that this breaks the build.

This setup might need changes so it also works for `master`.

### DIFF
--- a/src/app/state/actions/index.tsx
+++ b/src/app/state/actions/index.tsx
@@ -483,7 +483,7 @@ export const handleProviderLoginRedirect = (provider: AuthenticationProvider, is
         dispatch(showAxiosErrorToastIfNeeded("Login redirect failed", e));
     }
     // TODO MT handle case when user is already logged in
-}
+};
 
 export const handleProviderCallback = (provider: AuthenticationProvider, parameters: string) => async (dispatch: Dispatch<Action>) => {
     dispatch({type: ACTION_TYPE.AUTHENTICATION_HANDLE_CALLBACK});


jacbn:
I like the idea of this, and it's cool that we can ESLint just the lines of the commit! I see only one small downside, which I'll get to in a sec.

I do think, however, that this isn't going to help fix the types of issues you describe. Both "real issues" you described were problematic because _we didn't have the ESLint rule for them in the first place_ – I think in general we are quite good at preventing potentially harmful ESLint warnings if they appear in the IDE (perhaps less so for accidental but harmless console.logs and imports, I concede). The target of this PR then seems to me to be more on the "trivial changes" side than the "preventing violations", which is fine, but worth noting.

I'm happy to use a set of rules made elsewhere, provided we have the ability to disable / change the priority of individual rules within them. 

As for the downside, we're going to have to start including many more `// ts-ignore`s etc around useEffects missing dependencies. I do prefer _not_ having those at the moment, as they tend to hide issues when the useEffect's dependencies change and it seems like everything is fine.

barna-isaac:
thanks @jacbn  and @mwtrew for the reviews. I'll merge this after I've made sure it works on `main`, after I'm done reviewing the search PR (I think by Friday end of day).

PR #1744, opened 2025-09-23
Make LLM-marked questions presentable on Isaac Science

sjd210:
This touches several different areas, and makes some minor improvements to Ada along the way. In brief:

- Overhauls LLM-marked question page to actually have styling on Sci
   - I'm not entirely confident on the layout and colouring, but I mostly just want it an a state that the Sci team can use and give feedback on
  - A question page with multiple question parts in an accordion in which only some are LLM-marked have those specific accordions labelled (This also affects Ada, but only for question pages with multiple parts e.g. intern-written extend pages)
- Adds LLM-marked indicator to the Question Finder ALVI. Happy for feedback on the placement - maybe we make it a little smaller?
- When the account settings are linked to in reference to consent, highlight the relevant setting (This affects Ada too).
- Links exist to the privacy policy, although the page itself has yet to be changed. I've removed FAQ links from text, but we probably should have some information there too and re-link it.

_(I've changed my local dev Sops config, porting over the ada staging configs, to make the questions actually answerable. It'll probably help to do the same when reviewing)_


jacbn:
This works great! The separation of colours into Isaac- and Ada- specific .scss files is a little annoying, and perhaps encourages some thought into shared colours. I tried this to some extent with $color-theme-xxx but the shades of grey in particular are now quite disjointed (there are 3? 4? types of grey) and it is annoying. 

This point aside for now, I have suggested a few styling changes here: https://github.com/isaacphysics/isaac-react-app/pull/1767. Most of it is cleaning up, but I wanted to make sure we were using the new styling where relevant. I also shrunk the `.small` version a bit more, and I think it fits more cleanly in search results. Let me know what you think!

PR #1748, opened 2025-09-26
Switch Exo 2 source to fix issues with ø character

jacbn:
The Google Fonts .ttf source for Exo 2 renders ø without the line passing inside the o. From my research this is not a standardised way of writing it, _with the exception of in extremely bold fonts_. As such, content are keen to keep this letter in the same format learners will have been used to seeing it in textbooks, other sites, etc.

It appears that other variable-weight sources for Exo 2 do not have this issue. I sourced this .woff2 ([caniuse](https://caniuse.com/woff2)) from Adobe, and seems to work as expected. ø does indeed now have a line through when below ~650 font weight.


mwtrew (approved):
Wasn't sure about licensing here, but [the Adobe page](https://fonts.adobe.com/fonts/exo-2-variable) for Exo 2 links to [this license](https://fonts.adobe.com/variations/51937/eula), which is OFL and therefore fine to redistribute.

PR #1750, opened 2025-09-26
Dependency updates

mwtrew:
This addresses all vulnerable packages from `yarn audit`, except for two dev dependencies:

- `undici`, which needs a major version bump that breaks MSW
- Some transitive dependencies of `@frsource/cypress-plugin-visual-regression-diff`, of which we are already on the latest version


jacbn:
I have a feeling we tried to upgrade Cypress before, as I remember seeing the outlines on what's been selected, but we have clearly undone that. I'll keep a lookout for issues with this in case something pops up that we've seen before.

PR #1754, opened 2025-10-01
Collapsible container rework

jacbn:
Restructures the implementation of collapsibles to allow a second method through which height can be recalculated: if a child recalculates its height, its parent should too. This more naturally follows the logical structure of what is happening and avoids reliance on rerenders to update parents.

The old method of requiring a `useLayoutEffect` dependency on `children` is still required, as non-collapsible components updating may reflow the height and thus should cause a recalculation (c.f. Physics QF topic selector). 

The new method specifically permits updates when a child's `data-targetheight` prop changes, which the `useEffect` approach could not deal with as this is not stateful (i.e. updating the attribute did not trigger the effect).

### DIFF
--- a/config/jest/jest.polyfills.js
+++ b/config/jest/jest.polyfills.js
@@ -22,6 +22,7 @@ Object.defineProperties(globalThis, {
 });
 
 globalThis.ResizeObserver = class MockResizeObserver {
+    constructor(_callback) {}
     observe() {}
     disconnect() {}
 }


jacbn:
If we want to fix this, it's apparently [much more efficient](https://stackoverflow.com/questions/48087965/why-was-resizeobserver-introduced-to-listen-to-resize-changes-and-not-a-simple) (for reasons unbeknownst to most people it seems) to use a ResizeObserver over events. Using one of those would also remove the need to check for changes of state on the children, as any relevant state that changes the size will necessarily cause the observer to fire. 

This in theory also means the recursive parent call could be removed too, but since each parent would need to wait until its child has started changing size (i.e. approx. one render) before it starts, there would be a delay chain without it.

It would look like this _in place of_ the current `useLayoutEffect`:
```ts
useEffect(() => {
    const resizeObserver = new ResizeObserver(() => recalculateHeight());
    resizeObserver.observe(divRef.current as Element);
    return () => resizeObserver.disconnect();
}, [recalculateHeight]);
```

The reason I was avoiding this is I've been _told_ that having several resize observers on a page is quite slow. I don't know that we ever can introduce more than 10-15, with the QF filters and markbook being the worst offenders. I suppose with the changes to Ada boards now allowing 30 questions, that would be worse. In this case, maybe the events would be quicker.

Then again, the upside to this approach is we guarantee we have hit all the cases for which we would need to recalculate the height. Using `props.children` only checks state, as was the cause of this entire PR, and there may well be other ways you can modify the children that changes the height that doesn't trigger the useEffect.

I don't know what's best tbh. Any thoughts?

sjd210:
Mentioned in person, but to briefly document here: I've found a problem with our use of ResizeObserver in that it it doesn't work with vertical window resizes since the element being observed is the outer `collapsible-body` div, but the element being resized by the window is the inner div. 

We may need to add a ref to the child so that it may be the element that is observed in order to change e.g. `listRef`'s size - but we don't have direct access to it so this is complicated. Other solutions welcome

PR #1757, opened 2025-10-02
Trigger build to update VRTs

sjd210:
[This commit](https://github.com/isaacphysics/isaac-react-app/commit/19bb1cba242d3fecabf31a2cb360ef7068b11bde) has failed VRTs but due to sequencing(?) couldn't create a branch for the VRTs to be updated from. This PR is for that branch to be created.


sjd210:
Turns out the VRT _was_ broken. We don't show assignments > 1 month old (or 6 months with a button) on the assignment schedule, but the VRT was expecting two visible months

PR #1760, opened 2025-10-02
Update My Assignments filters

axlewin:
- Default to "To do"
- Fix assignment count indicators

### DIFF
--- a/src/app/components/pages/MyAssignments.tsx
+++ b/src/app/components/pages/MyAssignments.tsx
@@ -86,7 +86,7 @@ export const MyAssignments = ({user}: {user: RegisteredUserDTO}) => {
     // that require refetching.
     const assignmentQuery = useGetMyAssignmentsQuery(undefined, {refetchOnMountOrArgChange: true, refetchOnReconnect: true});
 
-    const [statusFilter, setStatusFilter] = useState<AssignmentState[]>([AssignmentState.ALL]);
+    const [statusFilter, setStatusFilter] = useState<AssignmentState[]>([AssignmentState.TODO]);
     const [titleFilter, setTitleFilter] = useState<string>("");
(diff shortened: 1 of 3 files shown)


src/app/components/pages/MyAssignments.tsx
```diff
@@ -86,7 +86,7 @@ export const MyAssignments = ({user}: {user: RegisteredUserDTO}) => {
     // that require refetching.
     const assignmentQuery = useGetMyAssignmentsQuery(undefined, {refetchOnMountOrArgChange: true, refetchOnReconnect: true});
 
-    const [statusFilter, setStatusFilter] = useState<AssignmentState[]>([AssignmentState.ALL]);
+    const [statusFilter, setStatusFilter] = useState<AssignmentState[]>(siteSpecific([AssignmentState.TODO], [AssignmentState.ALL]));
```
sjd210, line 89: Always defaulting to TODO here means it'll be selected even when TODO is empty but other categories aren't. This feels really bad when we have assignments to show and I think in that case we should switch to ALL to ensure we're actually showing them (and in which case the original concern about not seeing current work isn't relevant so it should be fine to do).

Since we're using a query, we can't control that in the definition here, but I added a snippet like this local to the assignments space:
```ts
if (!hasLoaded && filteredAssignments.length === 0) {
  setStatusFilter([AssignmentState.ALL]);
}
setHasLoaded(true);
```

sjd210 (changes requested):
This is a good implementation of what's been asked for, although I'm still not entirely sure I _agree_ with what's been asked for, as per the Slack discussion. Perhaps it'll be fine now that the categories are more recognisably labelled? Either way I'm not too worried about that now - better to get something done sooner and then we can just act if anything needs further changing.

That said, I do still have one small request anyway.

sjd210 (approved):
👍 👍 👍

PR #1776, opened 2025-10-13
Restyle components in hint tabs

sjd210:
- Re-enable automatic linking of **Glossary** and **Concepts** headings, but make them context-specific so as to not be unyieldingly large pages (as is why [they were removed](https://github.com/isaacphysics/isaac-react-app/commit/97211d3cbd26cdeb71f17c08cad6548107306d64))
- Make concept page links unbolded when inside a hints tab (as opposed to embedded in text)
- Reduce percentage-based width between glossary name and definition in hints menu, and replace percentage-based with absolute width for solitary glossary terms not requiring alignment
- On glossary page, adjust spacing at <=sm screenwidths such that names and definitions are aligned

_(+ a little general cleanup in the affected areas)_

### DIFF
--- a/src/app/components/content/IsaacGlossaryTerm.tsx
+++ b/src/app/components/content/IsaacGlossaryTerm.tsx
@@ -3,6 +3,7 @@ import {GlossaryTermDTO} from "../../../IsaacApiTypes";
 import {IsaacContent} from "./IsaacContent";
 import {formatGlossaryTermId} from "../pages/Glossary";
 import { Col, Row } from 'reactstrap';
+import { DeviceSize, useDeviceSize } from '../../services';
 
 interface IsaacGlossaryTermProps {
     doc: GlossaryTermDTO;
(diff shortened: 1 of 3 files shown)


src/app/components/elements/markup/markdownRendering.ts
```diff
@@ -79,10 +80,21 @@ export const renderInlineGlossaryTerms = (markdown: string) => {
...
+    if (isPhy) {
+        let linkContext = "";
+        if (pageContext?.subject && isDefined(pageContext?.stage) && pageContext.stage.length === 1) {
+            linkContext = `/${pageContext.subject}/${pageContext.stage[0]}`;
+        }
+
+        Object.assign(regexRules, {
+            [`[**Glossary**](${linkContext}/glossary)`]: /\*\*Glossary\*\*/g,
+            [`[**Concepts**](${linkContext}/concepts)`]: /\*\*Concepts\*\*/g,
+        });
```
axlewin:
```suggestion
         if (pageContext?.subject && !isSingleStageContext(pageContext)) {
            Object.assign(regexRules, {
                [`[**Glossary**](/glossary?subjects=${pageContext.subject})`]: /\*\*Glossary\*\*/g,
                [`[**Concepts**](/concepts?types=${pageContext.subject})`]: /\*\*Concepts\*\*/g,
            });
        }
        else {
            let linkContext = "";
            
            if (pageContext?.subject && isSingleStageContext(pageContext)) {
                linkContext = `/${pageContext.subject}/${pageContext.stage![0]}`;
            }

            Object.assign(regexRules, {
                [`[**Glossary**](${linkContext}/glossary)`]: /\*\*Glossary\*\*/g,
                [`[**Concepts**](${linkContext}/concepts)`]: /\*\*Concepts\*\*/g,
            });
        }
```
This works nicely for questions with single-stage contexts, but for multi-stage questions (e.g. `/questions/aircraft_balance` ) this links to the unfiltered generic concepts/glossary, which as you pointed out is the reason why these links were removed in the first place. Something like this would at least filter the concepts/terms to the relevant subject?

PR #2065, opened 2026-03-26
Refactor Assignment Progress to logically group components

sjd210:
The file structure of the AssignmentProgress group of files is a mess. e.g. `AssignmentProgressIndividual` is in the `/pages/` folder but is not a page itself, it has the same naming scheme as `AssignmentProgressGroup` despite `SingleAssignmentProgress` being its equivalent in `AssignmentProgressWrapper`, and components like `AssignmentProgressLegend` are not grouped to a logically consistent location within these files.

This PR does not change the logic of these files (except where purposely mentioned) but changes the structure so it's clear where each component is meant to be.

The important changes are:
- Adds a `/pages/assignment_progress` folder to associate these related pages together inside.
- Uses a `AssignmentSummaryCard` component to use shared code for the nearly identical Due/Attempted/Completed summary card shown above both a quiz and assignment summary.
- Uses a `ResultsTableHeader` component to group together the components that are consistently above a result table section (Title/Settings/Key) in each tab/quizzes.
  - **NOTE:** I have decided to make settings appear consistently on both the "Group overview" and "Detailed marks" tabs. On Sci, this means the "Hide/Show settings" button is available and the setting can be dismissed across tabs. On Ada, this means that "Show marks as percentages" checkbox appears on both tabs (it already _affected_ both tabs). This is a change, but I think one for the better.
- Rename and move components in files to the following structure:
  1. AssignmentProgressWrapper _(unchanged)_
  2. AssignmentProgressGroupsListing _(unchanged)_
  3. AssignmentProgressGroup _(unchanged)_
  4. AssignmentProgressIndividual _(this is now the page linked to in AssignmentProgressWrapper for an assignment. it contains the main page structure)_
  5. AssignmentProgressIndividualTabs _(new. this contains the GroupAssignmentTab and DetailedMarksTab code used in AssignmentProgressIndividual)_
  6. components/ResultsTableHeader _(new. all components related to settings or keys are moved here rather than distributed among several files)_
- Fixes visual bug on Sci causing a gap in the top border of the assignment progress tab cards.


jacbn (approved):
This looks good! It's quite likely we end up renaming Assignment Progress to something else come Assignments V2 / core skills so it might be worth renaming these files again to be consistent, but we don't have a working name yet so at least simplifying the structure means it should just be a simple rename job in the future to change it again.

One tiny point about the checkbox change:

PR #2080, opened 2026-03-31
List view expansion

jacbn:
- Expands list views to enable support for builder items (though their implementation is not in this PR for better separation);
- Fixes inconsistencies with where certain props are located in list views;
- Extracts common types into e.g. `ListViewItemBaseProps`, which more obviously dictate where a particular type of list view can be used.

I have currently kept ListView and ListViewCard separate since I couldn't figure out the types, but I see no reason why it shouldn't be possible to specify this with e.g. `layout='card'` in the `<ListView />`. The biggest problem is being able to throw a type error if an instance of ListView had only one layout type (e.g. was typed as `ListViewItemBaseProps<"item",`**`"list"`**`>` rather than a union).


src/app/components/elements/list-groups/AbstractListViewItem.tsx
```diff
@@ -143,48 +145,60 @@ type ALVIType = {
...
     subtitle?: string;
     breadcrumb?: string[];
     tags?: string[];
-    forceFullWidth?: boolean;
     url?: string;
     state?: AbstractListViewItemState;
     className?: string;
-    hasCaret?: boolean;
-} & ALVIType & ALVILayout;
+    Tag?: ListGroupItemProps["tag"];
```
sjd210: Is this needed for the question deck builder? Because it's not being used meaningfully anywhere in this branch.

If it is important, I really don't like it having such a similar name to `tags` while being quite a different prop (nor do I really like the capital T on something other than a type or component). I'd advocate for including all of `ListGroupItemProps` but that has a `[key: string]: any` which feels problematic, so perhaps just a more verbose name like `listGroupTag`?

src/app/components/elements/list-groups/AbstractListViewItem.tsx
```diff
@@ -143,48 +145,60 @@ type ALVIType = {
...
+    audienceViews?: ViewingContext[];
 };
 
 type ALVILayout = {
-    alviLayout: "card"
-    linkTags?: ListViewTagProps[];
+    alviLayout: "card";
 } | {
     alviLayout: "list";
 };
```
sjd210, line 160: Firstly, surely this could now be just `alviLayout: "card" | "list";`?

Secondly, ALVILayout is now only used as an intermediate type for AbstractListViewProps, and it would make sense for this prop to be directly in that context anyway. Could we just change that to:

```ts
export type AbstractListViewProps = {
    // properties of the list as a whole; affect the layout of every item but are not specific to any one item.
    alviLayout: "card" | "list";
    style?: "flat" | "stacked"; // flat = expand horizontally to fit content on one line; stacked = expand vertically, better for mobile or multi-column layouts; undefined = responsive
}
```

src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -329,13 +327,23 @@ export const BookDetailListViewItem = ({item, ...rest}: BookDetailListViewItemPr
...
-    const { cards, showBlanks, ...rest } = props;
-    return <ListGroup {...rest} className={classNames("list-view-card-container link-list list-group-links p-0 m-0 flex-row row-cols-1 row-cols-lg-2 row", rest.className)}>
-        {cards.map((card, index) => card ? <ListViewCardItem {...card} key={index} alviType="item" alviLayout="card"/> : (showBlanks ? <ListGroupItem key={index}/> : null))}
-    </ListGroup>;
+export const CustomListViewItem = (props: CustomListViewItemProps) => {
+    const { item, ...rest } = props;
+    return <AbstractListViewItem
+        {...item}
+        {...rest}
+    />;
```
sjd210, line 340: Here (and in the `Book...ListViewItem`s) is is a nice pattern, but I wish we were using it consistently for all the `...ListViewItem`s in this file for cohesion. e.g. `QuestionListViewItem` is a particularly bad offender with many props redundantly named that are directly from item.

For this specific function, it would also be more consistent to use `{item, ...rest}: CustomListViewItemProps` directly in the component call rather than separately unwrapping `props`, although that much is truly just a nit.

src/app/components/pages/MyProgress.tsx
```diff
@@ -198,11 +198,16 @@ const MyProgress = ({user}: MyProgressProps) => {
                     <Row id="progress-questions">
                         {progress?.mostRecentQuestions && progress?.mostRecentQuestions.length > 0 && <Col md={12} lg={6} className="mt-4">
                             <h4>Most recently answered questions</h4>
-                            <ListView type="item" items={progress.mostRecentQuestions} forceFullWidth={["sm", "lg", "xl"].includes(screenSize)} className="bordered"/>
+                            <ListView 
+                                type="item" 
+                                items={progress.mostRecentQuestions} 
+                                style={["sm", "lg", "xl"].includes(screenSize) ? "stacked" : undefined} 
+                                className="bordered"
+                            />
```
sjd210, line 206: Any reason why this one is expanded onto multiple lines, but the "Oldest unsolved questions" one isn't?

src/app/components/elements/list-groups/AbstractListViewItem.tsx
```diff
@@ -143,48 +145,60 @@ type ALVIType = {
...
     const isCard = typedProps.alviLayout === "card";
     const isDisabled = state && [AbstractListViewItemState.COMING_SOON, AbstractListViewItemState.DISABLED].includes(state);
 
     const isCrossTopic = isAda && tags?.includes("cross_topic");
     const isLLM = tags?.includes("llm_question_page");
 
-    const fullWidth = forceFullWidth || below["sm"](deviceSize) || (isItem && !(typedProps.status || typedProps.audienceViews));
-    const wrapTitleTags = below["xs"](deviceSize) || (isDefined(forceFullWidth) && !forceFullWidth);
+    const flatLayout = style === "flat" && above['sm'](deviceSize);
+    const stackedLayout = style === "stacked" || (below["sm"](deviceSize) || (isItem && !(typedProps.status || typedProps.audienceViews)));
```
sjd210: This isn't currently leading to any issues since style isn't being used very extensively, but surely these should be mutually exclusive? 

If `style === "flat"` and `above['sm'](deviceSize)` and `(isItem && !(typedProps.status || typedProps.audienceViews)))`, then both `flatLayout `and `stackedLayout` are true at the same time, which doesn't seem to make any sense.

sjd210 (changes requested):
This seems much more resilient. Very happy to see it, pending some small points :)

src/app/components/elements/list-groups/AbstractListViewItem.tsx
jacbn: > Is this needed for the question deck builder? 

Yeah, the structure is an outer list (`<ul />`), with each inner ALVI needing to be wrapped in a `<Draggable />` container, which is an element in the DOM and so by the [HTML spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element) must be a `<li />` as it is directly beneath a `<ul />`. This then means the *actual* ALVI underneath shouldn't also be an `<li />` which this does by default, so I'm using this prop to indicate making it a `<div />` instead.

> I really don't like it having such a similar name to tags while being quite a different prop (nor do I really like the capital T on something other than a type or component)

I fully agree with it being awkward. It is a component in the sense that generic elements write `<Tag {...props}>` to make use of it and this _must_ be capitalised (although to make a `div` or an `li` you write that as a string (so `const Tag="div";` before), so it's kind of a `Component | string`); c.f.: https://github.com/isaacphysics/isaac-react-app/blob/07487a9dc8b3c5b8fa3c4e25dd4d64c9110d2d8b/src/app/components/elements/CollapsibleList.tsx#L78
But I'm okay with using a prop like `listGroupTag` here; I think the code is clear enough about what it does.

src/app/components/elements/list-groups/ListView.tsx
jacbn: I'm scared of breaking things so I might do this separately, but aside from specific cases where we _don't_ want props directly from the item (since it's usually some kind of DTO, there may well be unintentionally overlapping props), I agree that the format of 
```tsx
return <AbstractListViewItem
    {...item}
    {overrides}
    {...rest}
/>;
```
ought to be consistent.

Also, valid nit. I changed this too many times before grasping what it was really doing and it was easier to manage things inside the function then.

src/app/components/elements/list-groups/AbstractListViewItem.tsx
jacbn: `(isItem && !(typedProps.status || typedProps.audienceViews)))` is left over from `fullWidth`, I was struggling to tell what it was doing. I might get rid of it entirely (i.e. leave just:
```tsx
const flatLayout = style === "flat" && above['sm'](deviceSize);
const stackedLayout = style === "stacked" || (below["sm"](deviceSize)
```
, which is consistent and makes more sense just looking at it), and manage the weird item layouts through `style={...conditions... && "stacked"}` in the case(s) we need it?

src/app/components/elements/list-groups/AbstractListViewItem.tsx
jacbn: > is _now_ only used

I figured there might be cases where layouts have props specific to them too; we had them previously, although I've moved them now. Maybe there won't be and it'd be fine to collapse it, but I was trying to future-proof this without knowing what might be needed... 🤷 

Can you think of any cases where this might happen? Maybe I'll simplify it and we can always revisit it at a later point if we need to, rather than do it now.

jacbn: > Firstly, surely this could now be just `alviLayout: "card" | "list";`?

On closer look, there is a distinction between your suggestion and what is there. Since `ListViewProps` is generic, TS can use limitations on the generic parameters provided to reduce the scope of the type, so it can infer that alviLayout is *specifically* a `"card"` or a `"list"`. Despite the same information being available for the `"card" | "list"` case, it doesn't actually reduce the scope of the type.

Making it `"card" | "list"` makes the type definition for a ListView's `items` prop result in never, so all `<ListView />` definitions fail to type. Try it and you'll see what I mean; probably easier seen that explained :)

---

FWIW, the `never` comes from `items` in this type:
https://github.com/isaacphysics/isaac-react-app/blob/07487a9dc8b3c5b8fa3c4e25dd4d64c9110d2d8b/src/app/components/elements/list-groups/ListView.tsx#L362-L371

(with e.g. `T = {some DTO}, G = "item", L = "list" | "card"`)

This is trying to say, `items` is a `T[]` iff the provided type `T` is usable within the types of list view that support a type `G` and layout `L`. This doesn't work with `L = "list" | "card"` but does with `L = "list"` or `L = "card"` specifically. Not entirely sure where that separation comes from; perhaps from `ListView` requiring exclusively `"list"`s and `ListViewCards` `"card"`s?

Wherever it comes from, it seems like we have to keep this as is.

sjd210: ~~I think I'd rather simplify it while it's simple, and then _if_ we ever need the separated types again we can easily split it out. And maybe at that point it would make more sense to split it within `AbstractListViewProps` rather than as a separate type, depending on if it interacts with `style`. This feels more presumptuous about the future than leaving it simple and mold-able.~~

src/app/components/elements/list-groups/ListView.tsx
sjd210: Cool 👍 As long as we're agreed, it doesn't have to be done here

src/app/components/elements/list-groups/AbstractListViewItem.tsx
sjd210: I'm still not sure about having the case where neither is true though. When `style === undefined && above['sm'](deviceSize)`, there is no explicit layout. I think that's something that _can_ work, but we'd have to be very careful about.

And actually, currently `flatLayout` is never true since we don't set style other than as `stacked` in `MyProgress`. We're already basically just using `!stackedLayout` anyway for cases where it should be flat.

sjd210: Also `above['sm'](deviceSize)` and `below['sm'](deviceSize)` are both inclusive of 'sm', so this leaves a case of overlap.

jacbn: My intention is that they can both be false! The styles exist as overrides for certain scenarios, e.g. preferring a stacked layout at `xxl` on the My Progress page because there are two columns (and we don't have containerised components yet :( ), or preferring an extra-flat layout to make rows as small as possible in the board builder. Both style "types" being unmet is simply there to mean "use a standard, responsive layout". Sometimes these overlap with the layouts, sometimes they don't.

They should be mutually exclusive, though. The overlap is definitely wrong. I'll fix!

sjd210: Hrm, I feel like that would be clearer with something like `layout: "stacked" | "flat" | undefined`, but that would make the repeated references to `stackedLayout` below clunkier. 
I concede, but I do think we'll have to be careful.

src/app/components/pages/MyProgress.tsx
```diff
@@ -201,13 +201,18 @@ const MyProgress = ({user}: MyProgressProps) => {
                             <ListView 
                                 type="item" 
                                 items={progress.mostRecentQuestions} 
-                                style={["sm", "lg", "xl"].includes(screenSize) ? "stacked" : undefined} 
+                                style={deviceSize !== "md" ? "stacked" : undefined}
```
sjd210, line 204: This now forces "xxl" and "3xl" screens to also be stacked, but they were both borderline requiring it anyway so I think this is fine

sjd210 (approved):

PR #2083, opened 2026-04-02
Cleanup various incongruities and outdated elements in icon folders

sjd210:
This _seems_ like a lot at first glance, but this is generally relatively straightforward. The changes are mostly to bring our icon folders up-to-date, within the following four categories:

1) Delete several icons that are no longer used and we have no reason to use in the future (e.g. mostly cases of non-generic icons used on Isaac Physics before the redesign like `icon-books.png` and `question_types_sprite.svg`, but also specific cases like `chevron-up.svg` which we use a rotated `chevron-right.svg` for now)
2) Groups the hex buttons used in inequality/graph sketcher together in a folder and renames the icons to reflect their use in both inequality and the graph sketcher (e.g. `inequality-confirm-phy.svg` -> `hex-confirm-sci.svg`)
3) Removes outdated icon-related css that is never used 
   - and similarly removes all Sci references from `RelatedContent` since we now have `RelatedContentSidebar` for it instead
4) Standardises all Ada icons in new format to 16x16px. Convert any cases that rely on the old size to use the standard icon format (e.g. "icon icon-sm icon-search" instead of "search-toggler-icon" with its own styling)


src/app/components/elements/cards/GameboardCard.tsx
```diff
@@ -174,7 +174,7 @@ export const GameboardCard = (props: GameboardCardProps) => {
         return <Link {...rest} className={classNames("w-100 d-flex assignments-card mb-3", rest.className)} to={boardLink}>
             {card}
             {onDelete && <Button className="delete-button" color="solid" onClick={(e) => {onDelete(); e.preventDefault();}}>
-                <img src="/assets/phy/icons/bin-black.svg" alt="Delete board"/>
+                <i className="icon icon-bin icon-sm icon-color-white"/>
```
jacbn:
```suggestion
                <i className="icon icon-bin icon-sm icon-color-white" aria-label="Delete board" />
```

sjd210:
From a quick look in the content, `question_types_sprite`, `question_hex`, `key_stage_sprite`, and `teacher_features_sprite` are all used in content and so should not be deleted (although the last two are also used in code and so weren't deleted anyway). I think everything else is good as-is.

jacbn (approved):
Agree with your above list. `cs/callout-icon-puzzlebulb.svg` is also used in Ada content, but it's only in some test pages I made a while back so happy to bin. 

As for the rest, I haven't checked every icon more than that they're all the same size; presumably this is a requirement for the later branch. If something noticeably changes I think we're just going to have to rely on someone spotting it, as this affects basically every corner of the site. It's not perfect, but it's what we did for the previous icons changes so 🤷

sjd210:
I hadn't added the content-used svgs back in yet 🙃 Ah well, I'll do it on main

PR #2086, opened 2026-04-08
Release Ada navigation!

jacbn:
- Removes feature flag and associated old code
- Adds a little icon next to Ada notifications as per request
- Updates tests to use non-flagged site


barna-isaac:
Amazing! Thanks for using a feature flag <3.

I've fixed a bug that I've noticed as I was testing the site locally. The line was introduced 5 months ago, so doesn't strictly pertain to this PR (but still affects the Dashboard, and I think you introduced it 😜).

And a note about feature flags: the next time we release and we're in a hurry, we can actually leave all of the flagging code in there, and just hardcode the "ON" state. That way, it's even less work just before the release.

Overall, though, great work on both the new navigation and the flagging^^. Thanks for making our lives easier!

barna-isaac (approved):

PR #2088, opened 2026-04-08
Ensure question attempt is always loaded into text input for symbolic questions

sjd210:
In the pre-seed world, we had a rudimentary version of this which worked well enough because the text-entry box would always start empty: https://github.com/isaacphysics/isaac-react-app/blob/a095e72743e1f75c3e4bda3fdf06f924094b7b7c/src/app/components/content/IsaacSymbolicChemistryQuestion.tsx#L142-L146

This was removed and replaced with more thorough text initialisation within `useState` to account for seeds, which generally works except that it can cause a race condition in which the `currentAttempt` is not loaded in time to be set.

To fix this, the branch adds a `questionAttemptLoaded` ref. Behaviour does not change if `currentAttempt` has been loaded in time, but the text input is given a second chance to load in a `useEffect` if it has not.

---

Also fixes two other small issues:
- Only update state when the user has interacted on logic questions (so the feedback banner is visible on page reload). This was already the case for other questions - the code just hadn't been ported over
- Stop showing _"The following symbols may be useful:"_ when there is no `symbolList`

### DIFF
--- a/src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
+++ b/src/app/components/content/IsaacSymbolicChemistryQuestion.tsx
@@ -10,6 +10,7 @@ import {
     parsePseudoSymbolicAvailableSymbols,
     initialiseInequality,
     useModalWithScroll,
+    isDefined,
 } from "../../services";
 import _flattenDeep from 'lodash/flattenDeep';
 import {IsaacQuestionProps} from "../../../IsaacAppTypes";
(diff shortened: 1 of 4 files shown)


jacbn (approved):
This is fairly urgent that we get this fixed and this does achieve that, so I'm happy to get this in for now. I'm just not a big fan in the sense that this doesn't fix the issue with the race condition, only provides a second control flow to `setTextInput` if it occurs. I haven't had the time to think this through properly, but I suspect having state that initialises to undefined (rather than using the potentially-unobtained `currentAttempt`) and updates with a `useEffect` provides a more consistent approach. 

This is a unfortunately much bigger change. However, doing this review has given me a number of thoughts about Inequality in general; I think there is much more simplification that can be done, extending the work you did to align the Inequality initialisation previously. We could also apply the above approach to this afterwards, significantly cutting down how big it feels.

Most significantly, across the different symbolic question types, note that:
- *almost all the state* – current attempt stuff, seed stuff, modal state stuff, symbols and other refs – are shared across most symbolic qs;
- *almost all the useEffects* are shared too – updating text input while editing graphically [aside, but why is this not done once on modal _close_?], initialisation, updating symbols 

My thinking is that we could move *all* of this into one e.g. `useInequality("chemistry")` or similar hook, that can hold the required state **and** effects, which returns all the values needed inside the DOM; the idea being to separate the business logic from the display logic, especially given how similar the symbolic questions' business logic is. This tends to be called a headless component / UI pattern, if you haven't seen it before. In theory, then, we'd only need one set of business logic, rather than 4.

I might make this into a card, at least as something to think about. What did you think?

sjd210:
@jacbn I agree. I went with a method analogous to what we had before so that knew it would work and could get this in quicker, but there's definitely room for improvement.

I was already thinking about that kind of thing when I was refactoring the rest of the elements in the symbolic files, but it'd already crept in scope from my original plans so I left it at the time. I think it'd be a huge improvement for the same reason as that original batch of work (i.e. reducing number of places that need to be changed when making edits/making differences clearer), like it'd've caught the reloading issue on logic questions that was fixed on this branch.

I've heard of headless components before as a pattern, but I'm not aware of good practice in terms of implementing them. I might do some reading :)

PR #2095, opened 2026-04-14
Use UTC for today's date

axlewin:
Fixes a bug where assignment start dates appear to default to yesterday, even though the actual value is today.


sjd210 (approved):
I had a quick look at all the places that we use `TODAY()` and couldn't see any differences made except in the one places this was buggy. There are some places like https://github.com/isaacphysics/isaac-react-app/blob/c05dd742d855fce23c28675f90797073302db26e/src/app/components/pages/quizzes/QuizTeacherFeedback.tsx#L148 where we now compare UTC date to BST date, but this being a `>` is the right way around for this not making a difference.

I think for cleanliness we should only compare likewise, but I think this branch is fine functionally speaking without it, so I'll make a separate card.

PR #2098, opened 2026-04-21
Bookmarks: Saving and listing

jacbn:
An initial, feature-flagged set of changes to enable the saving and viewing of bookmarks on the My Bookmarks page (`/bookmarks`), accessible from the My Isaac nav. 

Bookmarked questions currently appear as such on the question finder (there are no plans to allow _filtering_ by bookmarks here, as this is too computationally expensive). A future PR will manage the changes to the deck builder.

We will also soon enable bookmarking of concepts, however we must consider adjustments to the concepts endpoint if we want to allow this in a similar manner to e.g. the question finder. This will take longer, and so is not part of this initial PR.

Requires https://github.com/isaacphysics/isaac-api/pull/777 to run.


axlewin (commented):
This all works nicely on wide-ish screens, I think it's pretty much ready to go (behind a feature flag) once the backend's ready. Just noticed a couple of quirks on mobile:
- The "Search and filter" sidebar button gets duplicated on My bookmarks.
- There's no bookmark button on the question finder on mobile - is this intentional? (I realise there's not really room for it on particularly small screens).

A couple more minor thoughts:
- Could we adapt `icon-bookmark` for use as a hex icon on My bookmarks?
- The subject tag counts on My bookmarks don't update when a stage is selected (and vice-versa). I prefer this approach, but it is inconsistent with the other places where we have tag counts (`/concepts` & `/practice_tests`) - if we decide to go for non-updating counts, we should update the other places to be consistent.

jacbn:
> There's no bookmark button on the question finder on mobile

Yeah, I would like to add this but I can't find a consistent place for _any_ kind of button for it. We could add an entirely new row underneath each question with a on-question-style button, but I think the wasted screen space from doing that when space is already at a premium on mobile isn't worth it. You can still save it from the page itself on mobile, so I'm okay leaving it out.

> The subject tag counts on My bookmarks don't update when a stage is selected (and vice-versa). 

Good spot, I hadn't noticed the inconsistency. I'm hesitant to change other pages without asking content, so maybe we push this through behind the feature flag and ask while they have a chance to play around with it?

Made a new page icon too, hope you like it! 🙈

axlewin:
> Made a new page icon too, hope you like it! 🙈

It's beautiful ;) I agree that the other two points don't need addressing right now, so LGTM!

axlewin (approved):

PR #2101, opened 2026-04-21
Improve spacing above "scheduled work" dashboard card links

axlewin:
Makes this card consistent with the "groups" card above on `xs` screens


barna-isaac (approved):
Indeed looks better, see for example the new

Screenshot 2026-04-21 at 17 20 38
vs the old
Screenshot 2026-04-21 at 17 21 41.

Approved.

PR #2109, opened 2026-04-23
Add initial support for GCSE Biology

sjd210:
This adds support for all of GCSE Biology except specifically for the subject-specific Concepts page (redirecting instead to a filtered main Concepts page, since no such concepts exist yet and an entirely empty page may be confusing).

This also adds hardcoded support for disabling topics in the question finder filter, and for marking them as "Coming soon!". The former _could_ be done more dynamically by checking the list of searched questions in the API and returning which topics are included, but the latter couldn't - and this is just a temporary measure while the stage is new anyway so it doesn't seem worth the effort.


barna-isaac (approved):
Looks great! I especially appreciate the subtle fix you've made so subject-specific question finders now follow the subject theme.

I've checked, and we indeed now have:
- a new subject landing page for GCSE biology
- a new dedicated GCSE biology question finder and glossary
- as well as the remaining requested changes.

Thanks for going the extra mile and accomodating Bryony's request about disabling topics and showing "Coming soon" on the GCSE biology QF. For me, it was at first confusing why some things were simply disabled, and why others were marked coming soon. I then learned from the question decks page that the disabled but not coming soon topics are planned to go live in 2027 -- so those are also coming, just not soon. Not a big deal, and it's exactly what the content team has asked for, so I think it's fine as it is!

PR #2114, opened 2026-04-27
Switch icon colours to css vars for more expressivity

jacbn:
Replaces repeated `background-color: ...` props for each icon type (especially on buttons) with changes to `--icon-foreground` and `--icon-background`. The `::after` element (or parent, if flat) takes the `--icon-foreground` var always, likewise for `::before` and `--icon-background`.

The reason for this change is twofold. 
- First, it significantly reduces the complexity of the CSS related to colour changes. Changing an icon's foreground is as simple as `--icon-foreground: ...`; the old approach required setting the background of both the parent (for flat icons) and the `::after`.
- Second, it enables animated filling of icons, via the new `%icon-fillable` template; this requires two icon svgs, one outlined and one filled. While initially transparent, when `.fill` is applied, the background transitions to `--icon-background: var(--icon-foreground);`, creating a smooth fill animation. This is not used anywhere in this PR, but will be used in upcoming ones.

### DIFF
--- a/src/scss/common/_mixins.scss
+++ b/src/scss/common/_mixins.scss
@@ -276,8 +276,8 @@ $viewport-sizes: (
 @mixin icon-color-override($color) {
   // only do this if masks are supported
   @supports (mask-position: center) {
-    --icon-foreground: #{$color};
-    --icon-background: #{$color};
+    --icon-foreground: #{$color} !important;
+    --icon-background: #{$color} !important;
   }
 }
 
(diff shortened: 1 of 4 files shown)


sjd210 (approved):
Looks good! I've tried making a fillable icon myself and it was quite intuitive.

I'm also imagining that overriding the transition with different animations could be quite nice to add some more dynamism per icon. Obviously that can be done on a case-by-case basis.

jacbn:
Yep, I'd love to see things have a little more flair! My current "save this" animation is to fill a star, spin and pop a little so going along similar lines :)

PR #2116, opened 2026-04-27
Add coordinate q to question types VRT

axlewin:


barna-isaac (approved):
This and the playwright tests work great too! Thanks for writing a thorough test, that also checks what happens when a correct coordinate is missing!

PR #2117, opened 2026-04-27
My Saved Decks

jacbn:
Reworks My Question Decks so that only decks that are **created** or **manually saved** get saved to that page; no longer save decks that were simply attempted. Adds a save button (I used a star because this (will) appear on a different page to bookmarks – please disagree if you don't like it) to all decks to achieve this. Also renames the page to My Saved Decks to better reflect these changes.

Anyone should be able to save and view their saved decks, but teachers should also have the ability to set decks from anywhere without having to redirect to the Set Assignments page. To better enable this, a refactor of GameboardCards was undertaken.

Note that both Isaac and Ada have both requested the changes from automatic saving on attempt to manual saving only, but that the pages in which these are displayed are different; Ada does **not** have a My Saved Decks page or equivalent.

This is the first part of the wider changes to managing assignments; this does not rework Set Assignments in any meaningful way (though the cards do look slightly different).

---

Known issues:
- **Book pages'** gameboard API response does not include the required `savedToCurrentUser`, so they do not appear saved when loading the page;
- As above for **My Assignments**;


src/app/components/elements/cards/GameboardCard.tsx
```diff
@@ -71,66 +116,34 @@ export const GameboardCard = (props: GameboardCardProps) => {
...
+                {showMore ? "Hide details" : "Show details"}
+            </Button>
+            <Spacer />
+            <div className="d-flex gap-3 align-self-stretch align-items-center mb-2 order-0 order-sm-1">
+                {isPhy && gameboard && <SaveBoardButton board={gameboard} color="keyline" size="sm" />}
+                {isPhy && boardLink && <div className="card-share-link">
+                    <ShareLink linkUrl={boardLink} reducedWidthLink clickAwayClose size="sm" buttonProps={{color: "keyline"}} />
+                </div>}
+                {isTutorOrAbove(user) && <Button className="flex-grow-1" color="keyline" onClick={(e) => {e.preventDefault(); openAssignModal?.();}}>
+                    {isSetAssignments ? "Assign / Unassign" : "Assign"}
```
sjd210, line 149: This is missing the previous logic of also only using "Assign" when there are 0 groups that the gameboard is currently set to. It feels clearer to me that we keep this (and also it still has this behaviour on the table view, so even if we did want to change it, this is inconsistent with that)

src/app/components/elements/Gameboards.tsx
```diff
@@ -78,16 +78,16 @@ const CSTable = (props: GameboardsTableProps) => {
         </SortItemHeader>
         {siteSpecific(
             <>
-                <th className="text-center align-middle">Delete</th>
+                <th className="text-center align-middle">Unlink</th>
```
sjd210: Could we please use a consistent term for this? The tooltip on the buttons says "Save"/"Unsave" and obviously the renamed page on Isaac is "My saved decks", but these column headers say "Unlink" and as does the confirmation message for unlinking decks. They're the same action - they should have the same name!

I think "Save" is more user friendly, but "Link" is probably fine(/clearer?) for function names and such in code if we're happy with another pseudonym like that.

src/app/components/pages/SetAssignments.tsx
```diff
@@ -176,7 +165,7 @@ const CSTable = (props: SetAssignmentsTableProps) => {
         </SortItemHeader>
         <th>Manage</th>
         <th>Share</th>
-        <th>Delete</th>
+        <th>Unlink</th>
```
sjd210: Gameboard cards on the Set Assignments page on Isaac now have the save button included, but there is no equivalent in table view. We have this here for Ada (and always did because of the lack of separation between "My question decks"/"Set assignments") and so I think we can certainly add the same column for Isaac without any disconnect.

src/app/components/elements/SaveBoardButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+            user,
+        }));
+    }, [user, board, dispatch]);
+
+    const unlinkBoard = useCallback(() => {
+        if (!user || !board) return;
+        const confirmMessage = board.ownerUserId === user.id && !board.tags?.includes("ISAAC_BOARD")
+            ? `Are you sure you want to unlink your board '${board.title}' from your account? You'll only be able to find it again if you've set it as an assignment.`
+            : `Are you sure you want to unlink '${board.title}' from your account?`;
+        if (confirm(confirmMessage)) {
```
sjd210, line 38: This has strange interactions with the existing deck deletion code at https://github.com/isaacphysics/isaac-react-app/blob/main/src/app/state/slices/api/gameboards.ts#L162. 

- An admin may get two popups in a row when unlinking a deck that has been assigned - it feels like we can combine these into one.
- A non-admin gets an error after they have confirmed unlinking in this same scenario. If we want to maintain requiring a teacher to be linked to a deck while it is assigned, we shouldn't have that confirmation box appear in the scenario and should just fail early.
  - That said, maybe we shouldn't be enforcing it in this new system anyway? It was already inconsistent for groups with multiple managers (since only one manager would need to be linked to and set the assignment for it to apply to that whole group), and since saving is now a much more separate/manually controlled action, it seems more consistent to give them full control over saving.

src/app/components/elements/SaveBoardButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export const SaveBoardButton = (props: SaveBoardButtonProps) => {
+    const { board, size, className, ...rest } = props;
+
+    const dispatch = useAppDispatch();
+    const user = useAppSelector(selectors.user.loggedInOrNull);
+
+    const [justLinked, setJustLinked] = useState(false);
+    const isLinked = useMemo(() => board.savedToCurrentUser || justLinked, [board, justLinked]);
```
sjd210, line 21: This seems to be an API issue, but I'll leave a comment here while it's on my mind and it can be worked on later like the other known issues. This `board.savedToCurrentUser` is not behaving as it should for gameboard pages themselves. The board appears as saved if and only if the user has attempted any questions on it (as would've previously created a link).

Using the button does successfully save/unsave boards if in the right state to do so, but that state does not reflect the actual saved-ness of the board and whether it shows up in "My saved decks".

sjd210 (changes requested):
There's a lot here and most of it still seems to be working as I expect, but there are a few things to look at/discuss.

Love the spinning star animation btw :)

src/app/components/elements/cards/GameboardCard.tsx
jacbn: Ah, sorry, there's additional context here – Set Assignments (soon to be Manage Assignments) will only show **assigned** decks, so there'll never be a case where there are 0 groups assigned and `isSetAssignments` is true. You'll also need to go to Manage Assignments to _un_set an assignment, so Unassign won't be an option outside of it.

I'll update table view to match!

src/app/components/elements/SaveBoardButton.tsx
jacbn: Ah, thanks for spotting this. There's a card for `savedToCurrentUser` not working for My Assignments and book pages, but this seems to be a third case.

src/app/components/elements/Gameboards.tsx
jacbn: Argh, I was being overly pedantic and couldn't pick one. Let's go with save/unsave, it's what the page is called on Sci.

src/app/components/pages/SetAssignments.tsx
jacbn: My only concern with adding it here on Sci is that if we had it on the rightmost side, it's intertwined with the *assignment management* info/settings that the *board* (that you'd be saving) is unrelated to. A user could feasibly believe they are saving/unsaving the *assignment*, which is a very different action to what this button really does.

I'll have a think about where to put it, but I doubt it'll be a quick drop in.

jacbn: I've added in to the side for now. Not completely happy with it, but it aligns with My Saved Decks so there's more consistency now.

src/app/components/elements/SaveBoardButton.tsx
jacbn: Agree with your last point; to me, the spirit of these changes are to make saving a completely separate interaction to creating, assigning, etc. You should be free to unsave a deck if you're done with it, regardless of whether it is assigned, _so long as it is still accessible from somewhere_. Right now, this last point is not met, since I have not yet done the changes to Set Assignments; but when that shows all assigned decks, this should work as expected.

I've removed the whole chunk of code around stopping this; it's obviously a little weird right now with not being able to access assigned ones, but this should be resolved soon.

src/app/components/elements/SaveBoardButton.tsx
```diff
@@ -33,8 +33,8 @@ export const SaveBoardButton = (props: SaveBoardButtonProps) => {
     const unlinkBoard = useCallback(() => {
         if (!user || !board) return;
         const confirmMessage = board.ownerUserId === user.id && !board.tags?.includes("ISAAC_BOARD")
-            ? `Are you sure you want to unlink your board '${board.title}' from your account? You'll only be able to find it again if you've set it as an assignment.`
-            : `Are you sure you want to unlink '${board.title}' from your account?`;
+            ? `Are you sure you want to unsave your board '${board.title}' from your account? You'll only be able to find it again if you've set it as an assignment.`
```
sjd210: Do we want to be using the word `"board"` here at all? Should it not be `siteSpecific("deck", "quiz")`?

src/app/components/elements/SaveBoardButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    if (!isLoggedIn(user)) return null; // anon users should not be able to save boards
+
+    return <IconButton
+        icon={{
+            name: classNames("icon-star", siteSpecific("icon-color-black-hoverable", undefined), { "fill": isLinked, "anim-star-select": justLinked }),
+            color: siteSpecific(undefined, props.color === "solid" ? "white" : "primary")
+        }}
+        className={classNames(className, "w-max-content h-max-content action-button", {"icon-button-sm": size === "sm"})}
+        title={isLinked ? "Unsave board" : "Save board"}
```
sjd210, line 55: As in prior comment:
```suggestion
        title={(isLinked ? "Unsave " : "Save ") + siteSpecific("deck", "quiz")}
```

src/app/components/elements/cards/BoardCard.tsx
```diff
@@ -227,17 +208,20 @@ export const BoardCard = ({user, board, boardView, assignees, toggleAssignModal,
...
+                    </div>
                 </td>
                 {isAda && <td className={"align-middle text-center"}>
                     <div className="table-share-link">
                         <ShareLink linkUrl={boardLink} gameboardId={board.id} innerClassName="btn-keyline" outline clickAwayClose />
                     </div>
                 </td>}
                 {isAda && <td className={"align-middle text-center"}>
-                    <IconButton icon={{name: "icon-bin", size: "sm"}} color="keyline" className="action-button" aria-label="Delete quiz" title="Delete quiz" onClick={confirmDeleteBoard}/>
+                    <SaveBoardButton board={board} color="keyline" size="sm" />
```
sjd210, line 224: Ada now has two of these. One in the Unsave column and one in the Manage column. Can we get rid of one of them?

src/app/components/elements/cards/GameboardCard.tsx
```diff
@@ -77,68 +122,36 @@ export const GameboardCard = (props: GameboardCardProps) => {
...
+            <Button className="my-2 btn-underline order-1 order-sm-0" color="link" onClick={(e) => {e.preventDefault(); setShowMore(!showMore);}}>
+                {showMore ? "Hide details" : "Show details"}
+            </Button>
+            <Spacer />
+            <div className="d-flex gap-3 align-self-stretch align-items-center mb-2 order-0 order-sm-1">
+                {isPhy && gameboard && <SaveBoardButton board={gameboard} color="keyline" size="sm" />}
+                {isPhy && boardLink && <div className="card-share-link">
+                    <ShareLink linkUrl={boardLink} reducedWidthLink clickAwayClose size="sm" buttonProps={{color: "keyline"}} />
+                </div>}
+                {isTutorOrAbove(user) && <Button className="flex-grow-1" color="keyline" onClick={(e) => {e.preventDefault(); openAssignModal?.();}}>
```
sjd210, line 148: This is showing up on the My Assignments page even though `openAssignModal` is Set Assignments only so it doesn't do anything. Shouldn't this be `isSetAssignments`-gated (or by openAssignModal truthiness itself)?

src/app/components/elements/cards/GameboardCard.tsx
```diff
@@ -77,68 +122,36 @@ export const GameboardCard = (props: GameboardCardProps) => {
...
+        {below['xs'](deviceSize) && <CardUsageInfo className="d-flex w-100 justify-content-around" gameboard={gameboard} groupCount={groupCount} isSetAssignments={isSetAssignments} />}
+
+        <div className="d-flex flex-column flex-sm-row align-items-start mt-2">
+            <Button className="my-2 btn-underline order-1 order-sm-0" color="link" onClick={(e) => {e.preventDefault(); setShowMore(!showMore);}}>
+                {showMore ? "Hide details" : "Show details"}
+            </Button>
+            <Spacer />
+            <div className="d-flex gap-3 align-self-stretch align-items-center mb-2 order-0 order-sm-1">
+                {isPhy && gameboard && <SaveBoardButton board={gameboard} color="keyline" size="sm" />}
+                {isPhy && boardLink && <div className="card-share-link">
```
sjd210, line 145: `GameboardCard` is already a Sci-only element, so we don't really need `isPhy` here

---

While on this point, the Save Board button gets added to the Sci "My assignments" page but not to the equivalent Ada "Assigned to me" page since it uses difference components. It would be teacher-only, but it probably still makes sense for it to be there too to save the user needing to click through to the assignment page itself for it?

I don't think it's a big deal and think we can choose to forgo it for styling simplicity, but thought I'd raise it while spotted.

sjd210 (changes requested):

sjd210:
@jacbn Is there a reason this is still a draft?

src/app/components/elements/SaveBoardButton.tsx
jacbn: Apologies, this was fixed in a later branch but not backported (as with most issues you've raised); I've cherry-picked it now.

src/app/components/elements/SaveBoardButton.tsx
jacbn: As above

src/app/components/elements/cards/BoardCard.tsx
jacbn: As above

src/app/components/elements/cards/GameboardCard.tsx
jacbn: This one was a bit tricky to backport, I've left a comment to remove a quick definedness hack when merging later that was sorted properly in future commits.

src/app/components/elements/cards/GameboardCard.tsx
jacbn: Happy to do, but I'll do them in `set-manage-all-work-types` to avoid rebasing (and to get this in asap).

I haven't discussed whether Ada like the "assign button on all board instances" feature – they may prefer assignment to happen in one place on the site. I've tried to keep Ada changes to a minimum (beyond cleanup), but if you think it's a good feature we can always just add it and see.

sjd210 (approved):

PR #2122, opened 2026-04-28
Release dark mode behind feature flag

jacbn:
Enables a staging-only dark mode to locate areas that need fixing before a wider release.

Almost all of the changes here are CSS only. The `.tsx` changes are mainly to support the new setting in `/account`, and adjusting the position of `data-xxx` attributes in the DOM so that they apply to all non-`#root`-enclosed elements like popups and modals.

---

Owing to the overhaul in how colours are generated / used, most Sci [VRTs](https://github.com/isaacphysics/isaac-react-app/pull/2124) have failed. This is usually because I changed a hard-coded colour to one of the `color-xxx-100` variables that can adjust to the lightness mode, and so there is a subtle shift in the displayed colour. In some instances (most noticeably the hexagon-themed banner behind the page supertitle), the colours have changed slightly more dramatically to suit both dark and light modes when flipping – I hope that none of these changes are too egregious.

As for reviewing, many of the CSS files that have changed are simply migrating from the old hard-coded `$color-xxx-100` format to `var(--color-xxx-100)`, which can be adjusted by the lightness theme. The most important ones to understand are `src/scss/phy/isaac-primitives.scss` and `src/scss/phy/color-theme.scss`; `isaac-primitives` defines the hard-coded constants and the different lightness mode maps, and `color-theme` creates CSS vars for all of these, as well as defining e.g. a lightness-dependent `color-theme-200`, as it did before.

I'd recommend going commit-by-commit, as many of the first ones (particularly https://github.com/isaacphysics/isaac-react-app/pull/2122/commits/763d50c407fc64133f1ad62839cc0dc57d0c2bc1 and https://github.com/isaacphysics/isaac-react-app/pull/2122/commits/c95e765c886035c2c8d08f29d72cf100769bb940) covered many individual files but made the same change to each, so can be reviewed together quickly.

If there are _additional areas that need dark mode support_, rather than bugs with any of the areas in this PR, I'd rather we add them to a list that content have access to, as this PR is intended only to support the feature, not to make it complete.


src/scss/phy/isaac.scss
```diff
@@ -317,13 +173,13 @@ $card-inner-border-radius: 1rem;
...
-$dropdown-link-hover-bg: $color-neutral-50;
-$dropdown-link-active-bg: var(--subject-color-200, $color-neutral-200);
+$dropdown-link-hover-bg: var(--color-neutral-50);
+$dropdown-link-active-bg: var(--subject-color-200, var(--color-neutral-200));
 $dropdown-link-active-color: inherit;
 
 $input-border-radius: 0.5rem;
 
-$tooltip-bg: rgba($color-neutral-700, 0.98);
+$tooltip-bg: rgba(var(--color-neutral-700), 0.98);
```
sjd210: This has broken tooltips, with or without dark mode enabled.
I _think_ what's going on here is that `rgba` is expecting 3 separate rgb values in this position, and sass is able to convert a `$color` hexcode into those three values at compile time, but the same can't be done for a `var(--color)`.

We're only very barely using the opacity anyway so I suppose it wouldn't be a huge deal to remove it, but it is at least slightly nice to have. I think the only alternative is some variant on hardcoding this case.

src/scss/phy/isaac.scss
```diff
@@ -305,7 +161,7 @@ $container-max-widths: (
 
 $enable-negative-margins: true;
 
-$border-color: $color-neutral-200;
+$border-color: var(--color-neutral-200);
```
sjd210, line 164: I'm not sure exactly what's going wrong here, but this is breaking the bootstrap-based `form-check-input` border specifically on light mode.

We can override this by defining our own `border-color: var(--color-neutral-200)` in our own `.form-check-input` code in `phy/forms.scss`. If there aren't any other cases of issues caused by this, this is probably fine by itself (and I can't fine any other issues!) - but I do worry that other cases could come up.

src/scss/phy/typography.scss
```diff
@@ -243,12 +247,12 @@ h5 {
...
-  color: $gray-194;
+  color: var(--color-neutral-800);
   display: inline-block;
   text-decoration: underline;
 
   &:hover {
-    color: $gray-194;
+    color: var(--color-neutral-600);
     text-decoration: none;
   }
```
sjd210, line 257: I think this is the only super noticeable change from our existing light-mode colour scheme. When these `<a>`s are embedded in text, they're now notably different from the rest of the text. I'm not a fan.

I vote switching these over to `color-neutral-900` and `color-neutral-700` instead (or not changing the colour at all on-hover, like we did before)

src/app/components/elements/panels/QuestionFinderFilterPanel.tsx
```diff
@@ -114,8 +114,10 @@ export function QuestionFinderFilterPanel(props: QuestionFinderFilterPanelProps)
         }
     };
 
+    const HeaderTag = siteSpecific("div", CardHeader);
+
     return <div className={classNames({"card": isAda})} data-testid="question-finder-filters">
-        <CardHeader className="finder-header" onClick={(e) => {
+        <HeaderTag className={classNames("finder-header")} onClick={(e) => {
```
sjd210:
```suggestion
        <HeaderTag className="finder-header" onClick={(e) => {
```

sjd210 (changes requested):
I checked through the site quite extensively on both light and dark mode, and the only blocking changes (in light mode) that I could find are the small ones I commented on here.

I've also got some brief notes about issues I found while trawling in dark mode itself, but I'm recording these in a doc elsewhere. They're just bullet points right now, but I'll expand on them/attach pictures by the time we share it with the content team - and then they can follow suit.

src/scss/phy/isaac.scss
jacbn: Ah, that's a good explanation for it; sass perhaps has its own `rgba` function that overrides the css one.

For constants scss recommends `color.scale` or `opacify`, but since this is a variable we can use a pure-css alternative with `color-mix(var(--color-neutral-700) 98%, transparent)`. It doesn't exactly match (I *assume* because `transparent` is `#00000000` rather than `#______00`, so the hue ends up changing very slightly too), but it's pretty unnoticeable. This works on modern browsers and is something we've used elsewhere, so good for consistency.

src/scss/phy/isaac.scss
jacbn: Bootstrap defines `--bs-border-color` at `:root`, which other colour props use. This `$border-color` changes what that colour is. I thought at first that this was being used in SCSS *functions* somewhere, since you can't use CSS variables inside these; on closer look, however, it was compiling fine but `--color-neutral-200` genuinely wasn't defined. The issue is that, while it is defined when you use the prop, it isn't defined **at :root**; it was being defined one layer down. Bootstrap couldn't use the definition from a DOM element lower down than the root. 

As you rightly say, adding our own `border-color: var(--color-neutral-200)` fixes this because the colour is defined at that stage, but this avoids fixing the issue in general. To fix this more widely, I've just moved the `data-...` attributes up one more layer to the `:root`. I don't think this hurts anything, and all the necessary colour props are always defined at the topmost level now, where Bootstrap can see them.

jacbn:
To give a bit more context to https://github.com/isaacphysics/isaac-react-app/pull/2122/commits/b1b8678ae617c78a6811fe1eb9cda312e6360f4d than the comment:
 
I've noticed another checkbox issue, this time seemingly caused by the `<meta name="color-scheme" content="light dark">` HTML tag I added. I think what's happening is the browser sees this and assumes the page (in light theme) is able to be rendered *there and then* in dark mode, so if your *system* is in dark mode, it sets some initial properties to dark variants. The reality is that while we can render the page in both light and dark mode, which we show is **not** defined by the browser / OS theme right now (I would like to have a "system default" mode which respects this, but we can't with our current SOPS setup). 

I've removed the light dark HTML meta tag for now, and replaced it with a CSS `:root {color-scheme: light}` when in light mode and `:root {color-scheme: dark}` when in dark mode. This CSS is equivalent to the HTML tag; the only reason HTML is slightly preferred is that browsers load it before downloading the CSS.

src/scss/phy/typography.scss
jacbn: Not sure why I did this. The default should definitely be 900; removing the underline is enough on its own, it doesn't need a colour change. A quick search suggested that if we change the colour on hover we ought to change it on keyboard focus as well; I think I'd rather just leave this as close to how it was and we can decide later if we want to change accessibility standards and do it formally and properly then.

PR #2127, opened 2026-04-30
Add Safari-only dropdown scrollbar

jacbn:


barna-isaac (approved):

barna-isaac:
Fixes the issue for this question :). If the issue is ever raised again, I've got an idea for a solution that would work on any number of cloze items. When the dropdown opens, we could fire an upwards scroll event. This would make iOS show the scroll bar, without actually doing any scroll. The scroll bar would disappear after a while, but I still think this would be a good subliminal signal that most users pick up.

PR #2128, opened 2026-04-30
Begin migrating to Assignments v2 link format

jacbn:
All changes should be behind a feature flag.

---

Creates "mirror" pages that use the new Assignments v2 link format – `/assignment/{a_id}/view` and `/assignment/{a_id}/question/{q_id}` – and some components to support them. Links from e.g. on the `/assignment/{a_id}/view` page are from the API, so haven't yet been updated; however, on going to an assigned question page, a new warning will popup that does contain the correct link. 

Current link changes:
- Student dashboard
- Assignment schedule
- Assignment progress

I'd recommend going commit-by-commit to review here; https://github.com/isaacphysics/isaac-react-app/pull/2128/commits/eafd28351f7ffca07425a9b53e9038ba0537636e is quite large but entirely moving files around, no changes were made; Git can't easily understand that a two files were renamed, one into the old name of the other :/


barna-isaac:
When the feature flag is disabled, the "My assignments" page generates broken gameboard links, which still use the new `/assignments/...` format.

barna-isaac:
When the feature flag is disabled, an extra `;` is shown on the `/question_decks#{id}` page, see screenshot

Screenshot 2026-05-05 at 13 34 28

src/app/components/navigation/AssignedQuestionWarningBanner.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const AccessingAssignedQuestionOutsideAssignmentWarning = ({question, assignment}: AccessingAssignedQuestionOutsideAssignmentWarningProps) => {
+
+    const isAssignmentsV2 = useFeatureFlag(FeatureFlag.ASSIGNMENTS_V2);
+
+    // don't bother loading if we're in an assignment – this warning won't show
+    const {data: assignments} = useGetMyAssignmentsQuery(!isAssignmentsV2 || assignment?.id ? skipToken : undefined);
+    const assignmentsToDo = getAssignmentsToDo(assignments);
+
+    if (!isAssignmentsV2) return null;
```
barna-isaac, line 23: Why don't we return immediately after line 17? Wouldn't that be safer?

barna-isaac (changes requested):
The important bits to fix are the broken My Assignments links and the unnecessary `;` character on the Gameboard page. After that, I think it's good to merge, although @jacbn please also resolve the merge conflict in `featureFlag.tsx`.

As far as I understand, this PR did not intend to make any changes when the `ASSIGNMENTS_V2` feature flag is disabled, and I haven't found any changes besides the two bugs.

Just a few more notes, none of which blocks merging this PR.

When the flag is enabled, this PR introduces:
- new, assignment-specific URL's for gameboards and questions,
- and shows a warning when an assigned question is being attempted outside of an assignment.

@jacbn explained that the warning will show even when the question is being attempted for an assignment. I'm fine with this, since this is only when the flag is enabled. @jacbn also says this can't be fixed without a backend change. I think it could be, but this would involve creating the URL on the client, which @jacbn says now happens on the backend, and we might not want to change this.

I also wonder what our long-term plans are for these changes. I imagine we should only start showing the warning once the new V2 assignment behaviour has been introduced. However, maybe the new URL scheme could be released to live even before any of the new Assignments V2 changes are implemented? That way, only the warning would remain behind the feature flag.

src/app/components/navigation/AssignedQuestionWarningBanner.tsx
jacbn: Line 20 uses a hook; returning early would break React's [rules of hooks](https://react.dev/reference/rules/rules-of-hooks):
> Instead, always use Hooks at the top level of your React function, **before any early returns**.

We could return after L20, sure, since `getAssignmentsToDo` isn't a hook – but then the code becomes poorly segregated. There also wouldn't really be any performance gain, since if we're going to return,`assignments` is `undefined` so `getAssignmentsToDo` does no work in the first place.

jacbn:
> @jacbn also says this can't be fixed without a backend change. I think it could be, ...

Okay, yes, most things are possible. That doesn't mean we should do them...

The logic exists in the backend for a reason. If we duplicate the logic in the frontend to _sometimes_ modify a URL, there are now two places in entirely separate codebases that calculate question URLs that **we need to keep in sync**. Let's not make work for ourselves.

barna-isaac (approved):
Thanks for the fixes, looks good now!

PR #2131, opened 2026-05-01
Mild event details layout fixes

sjd210:
1) overflow-scroll -> overflow-auto (wrong way around in commit oops) so the scroll bar isn't always visible in the box when it isn't needed for applicable browsers
2) Change DOM structure so info box is always strictly below action buttons for Ada
3) Remove `<Row>` and `<Col>` from DOM which are no longer having an effect


axlewin (approved):
LGTM, this looks a lot better for Chrome especially!

PR #2133, opened 2026-05-01
Add spacing between news/event cards

axlewin:
Affects events & news on the homepage & subject landing pages on `xs` screens, and news on subject landing pages at all screen sizes.


src/app/components/pages/SubjectLandingPage.tsx
```diff
@@ -150,7 +150,7 @@ const FooterRow = ({context, books, news, events}: FooterRowProps) => {
             </div>
             <Row className="h-100 item-list-container">
                 {relevantEvents.map((event, i) =>
-                    <Col xs={12} key={i}>
+                    <Col xs={12} key={i} className="mb-3 mb-sm-0">
```
sjd210, line 153: This adds the margin at xs, but since we use the same format for both xs and md (per the line below), we need it at md too.
```suggestion
                    <Col xs={12} key={i} className={classNames({"mb-3": ['xs', 'md'].includes(deviceSize)})}>
```

(I've already committed this directly. This is moreso to explain why)

sjd210 (approved):

PR #2136, opened 2026-05-05
Simplify Graph Sketcher modal opening using new logic

sjd210:
The symbolic question refactor: https://github.com/isaacphysics/isaac-react-app/pull/2022 introduced a new hook for restoring scroll position when opening and closing a fullscreen modal. This idea is used for both Inequality and the Graph Sketcher, so the logic can be copied over rather simply.

Also: 
- Moves the `readonly` check that was being applied manually outside each `openModal()` to part of the hook
- Some mild cleanup of the` GraphSketcher.tsx` demo page
  - Remove now unuseful `<Row>` `<Col>` structure
  - Clean up `useEffect` dependencies related to opening/closing the modal

---

Further note: The graph sketcher demo page appears to be rendering two canvases locally? This is also happening outside of this branch for me, but not on the live sites. The page could do with some cleanup in general, so I'll make a card and we can get to it *eventually*.


jacbn (approved):
This looks great, thanks for this.

The duplication of the sketcher canvas locally is because there's no teardown in the `useEffect` that runs `makeGraphSketcher`, and strict mode runs UEs twice. There is a teardown in the sketcher question type itself, and while that does work, I'm not sure all the code there is necessary.

On trying to understand the sketcher page itself too, I see your point in that it really needs some work – it doesn't even load the sketch you've made in the preview... I've made some changes for how I'd modify things if you'd like to see if it's the same kinds of things you were considering, but no rush on the review; very low priority. I want to leave it in draft anyway until I can figure out why the main question type preview _isn't_ completely broken, like the /sketcher page currently is. 

Regardless, these changes are completely fine. :+1:

PR #2140, opened 2026-05-07
Refactor /sketcher page

jacbn:
Refactors the /sketcher page to make better use of React objects.

- Things that ought only to exist once (e.g. `previewSketch`) and that were not updating correctly and used in UE dependencies wrong, are now refs.
- Most interaction now occurs in `closeModal`. Good React style _should_ (although we are fairly bad at this) [avoid chained `useEffects`](https://react.dev/learn/you-might-not-need-an-effect#chains-of-computations) (i.e. one runs a `setX`, and `X` is a dependency in a second UE). I've tried to push everything into happening once, when the modal shuts.
- Adds a teardown to the `makeGraphSketcher` function, so the canvas is no longer duplicated locally.

I don't really want to take this out of draft until I understand the comment below :pensive:

### DIFF
--- a/src/app/components/pages/GraphSketcher.tsx
+++ b/src/app/components/pages/GraphSketcher.tsx
@@ -16,27 +16,44 @@ const GraphSketcherPage = () => {
     const user = useAppSelector(selectors.user.orNull);
     const [modalVisible, setModalVisible] = useState(false);
     const {openModal, closeModalAndReturnToScrollPosition} = useModalWithScroll({setModalVisible});
-    const [currentAttempt, setCurrentAttempt] = useState<GraphChoiceDTO | undefined>();
-    const [previewSketch, setPreviewSketch] = useState<GraphSketcher>();
-    const [initialState, setInitialState] = useState<GraphSketcherState>();
+    const [currentAttempt, setCurrentAttempt] = useState<GraphChoiceDTO | undefined>(undefined);
+    const previewSketch = useRef<GraphSketcher | undefined>(undefined);
     const previewRef = useRef(null);
     const [generateGraphSpec, {data: graphSpec}] = useGenerateAnswerSpecificationMutation();
 
-    const onGraphSketcherStateChange = (newState: GraphSketcherState) => {
-        setInitialState(newState);
-        setCurrentAttempt({type: 'graphChoice', value: JSON.stringify(GraphSketcher.toExternalState(newState))});
-        if (previewSketch) {
-            previewSketch.state = newState;
-            previewSketch.state.curves = previewSketch.state.curves || [];
+    const initialModalState: GraphSketcherState | undefined = currentAttempt?.value ? GraphSketcher.toInternalState(JSON.parse(currentAttempt.value)) : undefined;
+    
+    // loads with previous attempt data in real use
+    const [pendingAttemptState, setPendingAttemptState] = useState<GraphSketcherState | undefined>(undefined);
+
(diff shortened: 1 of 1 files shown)


src/app/components/pages/GraphSketcher.tsx
```diff
@@ -50,29 +67,21 @@ const GraphSketcherPage = () => {
...
-            data.curves = data.curves || [];
-            previewSketch.state = data;
-        }
-    }, [currentAttempt]);
+        return () => {
+            // teardown sketcher instance on unmount
+            previewSketch.current?.teardown();
+            p5?.remove();
+        };
+    }, [initialModalState?.curves]);
```
jacbn: This UE dependency is the one thing I am completely lost on. It seems to me that the intention was for setting `sketch.state` (i.e. `previewSketch.current.state` here) to automatically update the `previewRef` canvas and everything to just work; with this approach, we'd only ever need to run this UE **once**, and only adjust `sketch.state` to edit the preview. 

However. This simply doesn't seem to be working. The only way I can get the preview to update is if I insert `initialCurves: initialModalState?.curves` into the `makeGraphSketcher` (which _shouldn't need to be there_ because the graph always initialises from empty on this page), and rerun this UE after the modal closes with this extra dependency. I really don't understand how the sketcher question page works without doing this...

src/app/components/pages/GraphSketcher.tsx
```diff
@@ -16,27 +16,44 @@ const GraphSketcherPage = () => {
...
+    const initialModalState: GraphSketcherState | undefined = currentAttempt?.value ? GraphSketcher.toInternalState(JSON.parse(currentAttempt.value)) : undefined;
+    
+    // loads with previous attempt data in real use
+    const [pendingAttemptState, setPendingAttemptState] = useState<GraphSketcherState | undefined>(undefined);
+
+    const updatePreviewState = useCallback((attempt: GraphChoiceDTO | undefined) => {
+        // Set the state of the preview box whenever currentAttempt changes
+        if (previewSketch.current && attempt?.value) {
+            console.log(previewSketch.current);
+            const data: GraphSketcherState = JSON.parse(attempt.value);
```
sjd210: re: Your other comment

I've found the issue! The preview is meant to update when `previewSketch` updates, and that happens in this callback - however the data being fed into `previewSketch` is in the wrong format. We need to convert it into the internal state that the graph sketcher can read first:

```suggestion
            const data: GraphSketcherState = GraphSketcher.toInternalState(JSON.parse(attempt.value));
```

The reason it *was* working with `initialModalState` is that it does make this conversion (up on L24) and so the graph sketcher could be re-initialised with a correct data format and that would allow a rendered curve.

I'm pretty sure this change lets you remove the `initialCurves` parameter and `initialModalState?.curves` dependency that has been troubling you.

sjd210 (changes requested):
Everything apart from the UE that you've already commented about looks good, and I think(!) I've found a fix for that issue too :)

src/app/components/pages/GraphSketcher.tsx
jacbn: I see! Nice find. The sketcher really ought to error somewhere if given invalid data...

https://github.com/isaacphysics/isaac-graph-sketcher/pull/24

This does now work as expected! Happy to merge if you are.

sjd210 (approved):
Looks good. This doesn't require the graph sketcher update to work, so I'll merge this now and leave you to update the package version on main once it's released

PR #2141, opened 2026-05-07
Add charge to electron menu item for chemistry

sjd210:
Changes the `electron` menu item so that it has a minus subscript child, and a label reflecting that (All electrons are treated as if they have a negative charge in terms of marking, and we already translate e to e- from text to symbol, so this is just to make things more consistent).

Introduces a new `electron_nuclear` menu item that has the same behaviour as the old electron menu item (without an inbuilt subscript), and for which electron is always translated to for nuclear questions. This is useful because nuclear physics includes positrons as well as electrons so we cannot assume charge.

---

Also runs `lint --suppress-all` and `lint --prune-suppressions` (as should be done periodically anyway as we fix violations) to remove suppressions that were causing an unrelated eslint test failure in `InequalityModal`.

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -1,7 +1,7 @@
 {
   "src/IsaacApiTypes.tsx": {
     "@typescript-eslint/no-empty-object-type": {
-      "count": 12
+      "count": 11
     }
   },
   "src/IsaacAppTypes.tsx": {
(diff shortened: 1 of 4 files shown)


jacbn (approved):
I've added an extra commit to add the `editorMode` prop to one more fn call, but otherwise seems good. Feel free to merge after you've checked that's okay!

PR #2143, opened 2026-05-08
Don't unnecessarily show scrollbar on event cards

axlewin:
If there's only one event card shown it doesn't need a scrollbar, but at certain screen sizes one was shown anyway.

### DIFF
--- a/src/app/components/pages/SubjectLandingPage.tsx
+++ b/src/app/components/pages/SubjectLandingPage.tsx
@@ -148,7 +148,7 @@ const FooterRow = ({context, books, news, events}: FooterRowProps) => {
                 <h4 className="m-0">Events <span className="text-theme">({relevantEvents.length})</span></h4>
                 <div className="section-divider-bold flex-grow-1"/>
             </div>
-            <Row className="h-100 item-list-container">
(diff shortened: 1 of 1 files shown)


sjd210 (changes requested):
Jaycie and I briefly spoke about this last week, and unfortunately I think this just needs a more complicated solution than this to solve.

With this change, the card just abruptly cuts off without it being clear that it had no more useful info. We know this is it, but it may not be known to the user:

Image

This would of course be a bigger issue with e.g. event or location name being long enough to cover two lines, since at that point the "Available to book" would just be clipped off entirely.

I think the best way to handle this is to decouple this from screensize and instead have the card change size depending on container width (we have a [similar system for question metadata](https://github.com/isaacphysics/isaac-react-app/blob/c0062a25d6c1c1053cb2fb36791a7579bfc1653b/src/scss/phy/metadata-container.scss#L4) for reference). I don't expect you to do this now - I think my recommended next step is that we close this PR, and then make a new card to do the more complicated solution later.

axlewin:
> I think my recommended next step is that we close this PR, and then make a new card to do the more complicated solution later.

Agreed, given the priority I don't think this is worth pursuing right now 👍

PR #2144, opened 2026-05-08
Copies superseded/deprecated warning to My QDecks

jacbn:
The warning remains teacher-only.

### DIFF
--- a/src/app/components/elements/cards/BoardCard.tsx
+++ b/src/app/components/elements/cards/BoardCard.tsx
@@ -221,7 +221,7 @@ export const BoardCard = ({user, board, boardView, assignees, toggleAssignModal,
                     <a href={boardLink} className={isAda ? "fw--semi-bold" : ""}>{board.title}</a>
                     {isPhy && <span className="text-muted"><br/>Created by {<span data-testid={"owner"}>{formatBoardOwner(user, board)}</span>}</span>}
                     <br/>
-                    {isSetAssignments && <SupersededDeprecatedBoardContentWarning gameboard={board} />}
+                    <SupersededDeprecatedBoardContentWarning gameboard={board} />
                 </td>
                 {stagesAndDifficultiesTD}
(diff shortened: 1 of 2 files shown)


src/app/components/elements/cards/BoardCard.tsx
sjd210, line 250: On Set Assignments, this warning exists on both table and card view. On My Question Decks, you've only added it to card view. It looks a bit clunky, but I'd rather these be consistent.

```suggestion
                    <SupersededDeprecatedBoardContentWarning gameboard={board} />
                    </td>
```

We may also want to center the Attempted/Correct column hexes vertically for the sake of this case.

sjd210 (changes requested):

src/app/components/elements/cards/BoardCard.tsx
sjd210: On a similar note, on Ada this superseded warning currently appears on an individual Quiz page, but not on (edit: the Card View of) the Quizzes page. Should we be displaying this? I feel like it also has merit for Ada, but I'm not sure if they've seen it enough to approve.

jacbn: I've added the "staff-only" version without the description here instead; I think this looks much more natural and doesn't need to centre things. The only downside is of course that new users might be a little confused by what it means, but I'm okay with this in this specific scenario because they can **always** click through to the board to find out what that warning means in full.

sjd210: Okay, that seems suitable - but this change further raises the Ada question.

This doesn't really change the behaviour over there since their My Quizzes page is treated like Isaac's Set Assignments (at least for the sake of the `isSetAssignments` check) so I'm fine with this being merged as-is post-release without any changes on Ada...

...but at least for the sake of discussion: It's weird that this warning is given on the Quiz itself and Table View on Ada, but not Card View - and that we're now using this shortened version on Table View on Isaac's My Question Decks but the full version on (Isaac's) Set Assignments or (Ada's) My Quizzes Table Views. It's not strictly wrong, but it's weird.

PR #2146, opened 2026-05-08
Ada action buttons focus-visible interaction

sjd210:
- The share button did get an outline, but its background disappeared when highlighted. This is because it was `btn-tint` rather than `btn-primary` like the other two action buttons (this is site-specific). This has been fixed.

- The print and report buttons did not get an outline. This is because the `$outlined` property was false in its `ada-icon-button-base` mixin, and of the same specificity as the general `focus-outline` mixin application. A more specific `focus-outline` is added in the same place to override this.

### DIFF
--- a/src/scss/cs/button.scss
+++ b/src/scss/cs/button.scss
@@ -139,6 +139,9 @@
 .action-button {
   &:not(.outline) {
     @include ada-icon-button-base(48px, 48px, false);
+    &:focus-visible {
+      @include focus-outline;
+    }
   }
   &.outline {
     @include ada-icon-button-base(48px, 48px, true);
--- a/src/scss/cs/button.scss
+++ b/src/scss/cs/button.scss
@@ -139,9 +139,6 @@
 .action-button {
   &:not(.outline) {
     @include ada-icon-button-base(48px, 48px, false);
-    &:focus-visible {
-      @include focus-outline;
-    }
   }
   &.outline {
     @include ada-icon-button-base(48px, 48px, true);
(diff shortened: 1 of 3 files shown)


src/scss/cs/button.scss
```diff
@@ -139,6 +139,9 @@
 .action-button {
   &:not(.outline) {
     @include ada-icon-button-base(48px, 48px, false);
+    &:focus-visible {
+      @include focus-outline;
+    }
```
jacbn: Rather than adding more rules with higher specificity, is there any downside to either of:
1. removing the 4 `box-shadow: none;` rules in `cs/buttons.scss` that are preventing the accessibility rules showing, or
2. raising the specificity of the existing accessibility rules?

I just think that generally leaning towards having minimal CSS is cleaner. I can't see any downside to the former, since none of the Ada buttons seem to set `box-shadow` anyway? I might be missing something obvious though.

jacbn (commented):
Works great, much more consistent and legible outline now. I have a small style point over adding a new rule for this, but if none of the alternatives work I'd be okay with this as is.

src/scss/cs/button.scss
sjd210: It seems like the `box-shadow`s in 1 are in place because of bootstrap interactions adding their own `box-shadow` (I'm particularly noticing one when a button is both `:focus-visible` and `:active` without these rules) so I'm not inclined to remove these.

I was already considering raising the specificity of the existing rule, but wanted to make sure that wouldn't have any unwanted effects from being further reaching. I'll give that a go and have a quick look around to make sure it's good.

jacbn: Ah, okay. The only problem I can think of from increasing specificity is it being more noticeable if one of the two rules from the `focus-outline` mixin gets overridden by e.g. an `!important` and one doesn't. But it seems like all the `box-shadow: none !important;` and `outline: none !important` rules we have don't break things.

PR #2147, opened 2026-05-11
Fix various cut-off focus outlines

sjd210:
### Quite small:

- Add equal padding and negative margin to the start of ALVI card text region so it is visually equivalent but has space for full outline around tags
- Add margin to top of "See all books" link so that there is a minimum vertical distance between it and the above scrollbar, so the focus outline can't overlap with said scrollbar
  - (When there is already enough space, of course the margin has no effect)
- Add small margin to top of event card column on subject landing pages to stop top of outline being cut off

### Slightly larger (but still not very big):

Restructure book cards and subject landing page book container so that a) focus outlines are always visible (when book is focused), b) the spacing of books relative to the rest of the page is consistent, and c) the "View more books" button is properly centered when in a two-column layout - although I believe this is currently impossible.

Notably - in a full-width book layout, the left side of the book container was bleeding into the left margin of the page. It no longer does this, and instead respects the margin like in a book/event half-width layout.


jacbn (approved):
These are all improved. Both the book card and the ALVI are situations where we cannot use `overflow` to solve this owing to these elements relying on specific overflow values for other reasons, so the margin (or negative margin) + padding approach is the next most sensible.

LGTM!

PR #2149, opened 2026-05-11
Deck builder rework + bookmarks integration

jacbn:
~~Something is wrong with VRTs as the moment – still working out whether it's caused by this branch or not. The code is perfectly reviewable however :)~~

I have investigated the VRTs and most changes seem to be caused by the WCAG 2.2 change to make all click targets at least 24px x 24px. Some (notably the "go to deck" button on deck cards) were previously just shy of this and so have expanded _very_ slightly.

---

Reworks the deck builder to:
- use ALVIs in the main dnd interface
- support non-drag pointer operations for reordering qs in a board (WCAG 2.2)
- clean up the filters in the modal for ease of use
- support bookmarks under the feature flag; enable a bookmarked-only option in the search.

~~TODO: check `@hello-pangea/dnd` is not in the main bundle!~~ all good!


src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -327,7 +329,66 @@ export const BookDetailListViewItem = ({item, ...rest}: BookDetailListViewItemPr
...
+    const icon: TitleIconProps = { type: "icon", label: "Question",
+        icon: isPhy
+            ? {name: "icon-question", size: "md"}
+            : {name: QUESTION_STATUS_TO_ICON[CompletionState.NOT_ATTEMPTED], size: "md", altText: classNames(HUMAN_STATUS[state], "question icon"), color: "tertiary", raw: true}
+    };
+
+    return <DraggableListViewWrapper id={item.id ?? ""} index={index ?? -1}>
+        <div className="d-flex vertical-center bg-white rounded-2">
+            <div className="d-flex flex-column align-items-center">
+                <button type="button" title="Move question up" className="btn btn-blank p-0 m-0" onClick={() => onMove?.(item.id ?? "", -1)}>
```
sjd210: Could we disable this button and make it unfocusable when at the top of the list (and respectively for "Move question down" for the bottom of the list)? Or even remove the button entirely in this case ((Although this does look odd unless we find something else to fill the gap))?

It doesn't do anything in this state, and we don't want to confuse people in thinking that it might (e.g. thinking the list is cyclical, and that clicking the button would move the top element of the list to the bottom).

src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -327,7 +329,66 @@ export const BookDetailListViewItem = ({item, ...rest}: BookDetailListViewItemPr
...
 
-export type CustomListViewItemProps = ListViewItemBaseProps<"item", "list" | "card"> & {
+type BuilderListViewItemProps = ListViewItemBaseProps<"builder", "list" | "card"> & {
+    item: ContentSummaryDTO;
+    index?: number;
+    onMove?: (id: string, adjustment: number) => void;
+    onDelete?: (id: string) => void;
+}
+
+export const BuilderListViewItem = (props: BuilderListViewItemProps) => {
```
sjd210, line 340: The `box-shadow` for the focus outline of this entire structure is barely visible when just navigating. It renders perfectly fine when the item is being moved/dragged in the list (since the z-index is then above everything else), but not when static.

This also applies to the arrow and bin buttons on the item, although less heinously for the arrows.

src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -193,133 +193,171 @@ export const QuestionSearchModal = (
...
-                            <Label check><input type="checkbox" checked={searchFastTrack} onChange={e => setSearchFastTrack(e.target.checked)} />{' '}Show FastTrack questions</Label>
-                        </Form>}
-                </Col>
-                {isPhy && !isBookSearch && deviceSize === "lg" && <Col className="col-6">
+                </div>}
+            </Col>
+        </Row>
+
+        <Row>
+            <Col className={classNames("mb-2", {"d-none": isBookSearch})}>
```
sjd210: This is a pre-existing problem that I've just realised, but we absolutely should address it anyway.

Selecting a book disables the other search options in the modal from being selected, but doesn't restrict the search to not include those options as parameters (e.g. Set Stage to `University`, then set Book to `Step into Physics` - no results come up, since no questions in that book apply to University)

We should modify the request to stop including these parameters and fix this strange behaviour.

src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -327,7 +329,66 @@ export const BookDetailListViewItem = ({item, ...rest}: BookDetailListViewItemPr
...
+                <button type="button" title="Move question up" className="btn btn-blank p-0 m-0" onClick={() => onMove?.(item.id ?? "", -1)}>
+                    <i className="icon icon-chevron-up icon-color-muted-hoverable icon-color-theme-on-hover" />
+                </button>
+                <img src="/assets/common/icons/drag_indicator.svg" alt="Drag to reorder" className="mx-1 grab-cursor" />
+                <button type="button" title="Move question down" className="btn btn-blank p-0 m-0" onClick={() => onMove?.(item.id ?? "", 1)}>
+                    <i className="icon icon-chevron-down icon-color-muted-hoverable icon-color-theme-on-hover" />
+                </button>
+            </div>
+        </div>
+        <AbstractListViewItem
```
sjd210, line 369: Has this been tested on Ada at all? Because this new ALVI builder is definitely being used on the Quiz Builder and it looks very strange (no borders, topics not centered, bin button is huge and has bad contrast).

It certainly SHOULD be used on Ada, since the WCAG deadline is more relevant to them and we need the dragging change to meet that. Am I missing something?

src/app/components/elements/modals/QuestionSearchModal.tsx
sjd210: In a similar area to my last comment, and following the _"consider the possibility of having book sections appear in the sidebar, in place of the topic selection"_ point on the original card, I still find the existing system of hiding other options when a book is selected incredibly odd.

The original reason we did this (I believe?) was so we could provide a list of all questions in book order, when titles were used programmatically to accommodate this. We don't structure titles like this anymore, so this reason is void. 

The reason we _might_ want to do this now is because not all topics/stages are able to apply to a given book, but this doesn't accomodate books that DO cover many topics/stages, or for searching by difficulty, and it's also just inconsistent with the style of searching that we already use on the question finder - I think this is actively confusing. For example as a teacher, it would frustrate me if I wanted to look for Further A appropriate content in Pre-Uni Maths, but the option is hidden from me.

The API is suited just fine to searching with both books and other terms (hence the search already including those other terms as per my last comment), and I really do think that the simplest thing for now would be enabling all these options. (I think allowing searching by book section IS a nice idea, but there's no reason we need to keep these other restrictions while waiting for that to be implemented).

src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -193,133 +193,171 @@ export const QuestionSearchModal = (
...
-                    <Label htmlFor="question-search-title">Search</Label>
-                    <Input id="question-search-title" className="mb-3"
-                        type="text"
-                        placeholder={siteSpecific("e.g. Man vs. Horse", "e.g. Creating an AST")}
-                        onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
-                            setSearchQuestionName(e.target.value);
-                        }}
+    return <>
+        <Row>
+            <Col xs={9}>
```
sjd210: There is no book option taking up the last 3 units on Ada, so I think this should just be full width over there:


```suggestion
            <Col xs={12}>
```

src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -327,7 +329,66 @@ export const BookDetailListViewItem = ({item, ...rest}: BookDetailListViewItemPr
...
+            supersededByPath={item.supersededBy ? `/questions/${item.supersededBy}` : undefined}
+            style="flat"
+            subtitle={topic}
+            // subtitle={item.subtitle}
+            // breadcrumb={breadcrumb}
+            audienceViews={audienceViews}
+            className="flex-grow-1 align-content-center"
+            disableRedirect
+        />
+        <Button className="delete-button" color="solid" onClick={(e) => {if (item.id && onDelete) onDelete(item.id); e.preventDefault();}}>
```
sjd210: At particularly small screen sizes (<~420px width) the right end of this button is being cut off, even within the scrollable context. I haven't check precisely why this is happening, but I don't think this button has any reason to get smaller anyway so I think we can set a `min-width` on it to solve this.

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -446,60 +449,31 @@ const GameboardBuilder = ({user}: {user: RegisteredUserDTO}) => {
...
-                                                    <td colSpan={20}>
-                                                    </td>
-                                                </tr>}
-                                            </tbody>
-                                        </Table>
-
-                                    );
+                        <DraggableListViewContainer reorder={reorder}>
+                            {/* dragging here can be a little choppy on local development if browser cache is disabled! */}
+                            <ListView
```
sjd210, line 458: These new ALVI items no longer have the button to preview the question. I think this preview is helpful - are we definitely okay with losing it?

sjd210 (changes requested):
I haven't looked at everything in depth, but I've already found quite a lot that I think is unusual so I'm going to hand it back with these comments and look more closely at the code itself once things have changed for a re-review.

I think the general dragging structure and ALVI layout is looking good, but some edge cases were missed - and I'd like to have more of a discussion about how filtering is going to work when searching with books, because I don't like our old system and think we can use the opportunity to improve upon it.

src/app/components/elements/modals/QuestionSearchModal.tsx
jacbn: > The original reason we did this (I believe?) was...

It's more complicated than this, as always. We did this because the backend `questionSearch` function was a big mess.

We used map all kinds of tags together into one big clump – topics, books, subject etc were all just "tags". These tags, when thrown into the content manager search, were combined in a single "OR" search. Books being included in this OR was a pain; if you added other tags like things inside maths, you'd end up with a slew of maths results mixed in with the book results. To fix this, we just disabled searches while books were set, so you only saw that book.

Anyway, some work was done on the QF a while back to distinguish different tag types by expanding the `/questions` endpoint. Each tag [_type_](https://github.com/isaacphysics/isaac-api/blob/301e461ace52a5d98ecea46163bcb594541e87a0/src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java#L495) is now ORed amongst itself, but ANDed across each type (i.e. a search for `book1`, `book2`, `topic1`, `topic2` results in questions that match `(book1 || book2) && (topic1 || topic2)`

We've had this functionality on the QF for a while, but apparently it didn't migrate to the builder. We need to move the books from the `tags` field to a new `books` field to achieve this.

---

Having done this, I will say that 95% of topics not applying to a given book feels quite annoying from a UX perspective. Grouping by chapter is the only request we've had from teachers about this filtering and would be _so_ much better than this. I've copied over the filter list from the top of the QF to hopefully make it more visible when there are filters "blocking" questions from appearing to make it less egregious.

src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -193,133 +201,213 @@ export const QuestionSearchModal = (
...
-                            <Label check><input type="checkbox" checked={searchFastTrack} onChange={e => setSearchFastTrack(e.target.checked)} />{' '}Show FastTrack questions</Label>
-                        </Form>}
-                </Col>
-                {isPhy && !isBookSearch && deviceSize === "lg" && <Col className="col-6">
+                </div>
+            </Col>}
+        </Row>
+
+        <Row>
+            <Col xs={6} lg={4} className={classNames("mb-2")}>
```
sjd210: The Ada-specific Exam Board select is leaving a gap in this row for Isaac

```suggestion
            <Col xs={6} lg={siteSpecific(6,4)} className={classNames("mb-2")}>
```

src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -193,133 +201,213 @@ export const QuestionSearchModal = (
...
+                <Label htmlFor="question-search-stage">Stage</Label>
+                <StyledSelect
+                    value={searchStages.map(s => itemise(s, stageLabelMap[s]))}
+                    inputId="question-search-stage" isClearable isMulti placeholder="Any" {...selectStyle}
+                    options={getFilteredStageOptions()} onChange={selectOnChange(setSearchStages, true)}
+                />
+            </Col>
+            <Col xs={6} lg={4} className={classNames("mb-2")}>
```
sjd210: See above
```suggestion
            <Col xs={6} lg={siteSpecific(6,4)} className={classNames("mb-2")}>
```

src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -193,133 +201,213 @@ export const QuestionSearchModal = (
...
+                {/* required for sticky */}
+                <div className="flex-grow-1" />
+                {addSelectionsRow}
+
+            </Col>
+
+
+            <Col className="col-12 col-xl-9">
+
+                <FilterSummary filterTags={filterTags} removeFilterTag={removeFilterTag} clearFilters={clearFilters} />
```
sjd210, line 355: There being filter options in the sidebar for FastTrack and bookmarked questions, but them not showing up in this summary is odd (especially since I would expect "Clear all filters" to actually clear every filter). Can we add them as filter tags?

src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -327,7 +329,70 @@ export const BookDetailListViewItem = ({item, ...rest}: BookDetailListViewItemPr
...
+            title={item.title ?? ""}
+            subject={itemSubject !== "neutral" ? itemSubject : undefined}
+            url={item.url}
+            tags={item.tags}
+            deprecated={item.deprecated}
+            supersededByPath={item.supersededBy ? `/questions/${item.supersededBy}` : undefined}
+            style="flat"
+            subtitle={topic}
+            // subtitle={item.subtitle}
+            // breadcrumb={breadcrumb}
```
sjd210: Do we need these commented lines around? I think these look fine without the subtitle/breadcrumb fwiw.

src/scss/common/modals.scss
```diff
@@ -33,10 +33,6 @@
     }
   }
 
-  .modal-body {
-    overflow: auto;
```
sjd210, line 37: Without this `overflow: auto`, the title graphics on a question preview displays beyond the bounds of the modal.

Image

src/scss/phy/gameboard.scss
```diff
@@ -38,26 +38,12 @@
...
-    }
-  }
-}
-
 button.btn.delete-button {
   border-radius: inherit;
   border-top-left-radius: 0;
   border-bottom-left-radius: 0;
   padding: 0.625rem;
+  min-width: 36px;
```
sjd210, line 46: The Ada version of this button now has a similar issue that we solved here for Isaac (the button is shrinking from the right from ~400px). Could we add a similar `min-width` for it too?

sjd210 (changes requested):
A few more things to fix, but nothing major.

On the VRT front, the slight pixel chanes do appear to also be happening for me locally in this branch but not elsewhere - I'm not sure what's causing it.

src/app/components/elements/list-groups/ListView.tsx
sjd210: @jacbn There is still this issue of the focus being mostly obscured, particularly for the top and bottom elements of the list.

jacbn:
See change to description; I believe a majority of the VRT changes are owing to the WCAG 2.2 button target min sizing. Leaving them unmerged for reviewer (of particular note are of course the deck builder ones), but feel free to merge in.

src/app/components/pages/GameboardBuilder.tsx
sjd210: Minor, but losing this `mb-2` when at `xs` screens means there's no vertical gap between the icon buttons and question list.

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -445,61 +448,33 @@ const GameboardBuilder = ({user}: {user: RegisteredUserDTO}) => {
...
-                                                </tr>}
-                                            </tbody>
-                                        </Table>
-
-                                    );
+                    <div className={classNames({"is-invalid": submissionAttempted && !questionSetIsValid}, "p-1 m-n1 mt-2 gameboard-builder-container vertical-scroll-shadow-subtle")}>
+                        <DraggableListViewContainer reorder={reorder}>
+                            {/* dragging here can be a little choppy on local development if browser cache is disabled! */}
+                            <ListView 
+                                id="gameboard-builder-questions"
```
sjd210, line 459: Two related things when the screen size is small enough that these questions become scrollable,
1) The bin buttons on those questions will become un-aligned depending on the width of the rest of the container on Isaac, and will overlap with the other contents on Ada
2) The border around the list-group-item part of the container is visually set to only the width of the parent div above this

It only starts having an effect at <~420px, but does look quite strange from then on.

These can both be solved by allowing the content to actually fill the space, like with:

```suggestion
                                id="gameboard-builder-questions"
                                className="mw-max-content"
```

src/app/components/elements/list-groups/AbstractListViewItem.tsx
```diff
@@ -281,10 +300,16 @@ export const AbstractListViewItem = ({title, icon, subject, subtitle, breadcrumb
                     {isQuiz && <Col md={6} className="d-none d-md-flex align-items-center justify-content-end">
                         <QuizLinks previewQuizUrl={typedProps.previewQuizUrl} quizButton={typedProps.quizButton}/> 
                     </Col>}
-                    {isItem && contentId && typedProps.allowBookmarking && isLoggedIn(user) && bookmarksFeatureFlag && <button 
-                        className={classNames("alvi-bookmark", {"saved": isBookmarked(contentId)})} 
-                        onClick={() => bookmarkItem(contentId)}
-                    /> }
+                    {(isItem || isBuilder) && contentId && typedProps.allowBookmarking && isLoggedIn(user) && bookmarksFeatureFlag && <div
```
sjd210: As we discussed, bookmarks shouldn't exist on Ada (at least yet), but they are here on the ALVIs. I noticed it because of them appearing on the Quiz Builder, but this also affects e.g. the Question Finder (it already did, and is still behind a feature flag so 🤷‍♀️, but we should probably still not include them)

src/app/components/elements/modals/QuestionSearchModal.tsx
```diff
@@ -193,133 +201,221 @@ export const QuestionSearchModal = (
...
+                    value={getFilteredExamBoardOptions({byStages: searchStages}).filter(o => searchExamBoards.includes(o.value))}
+                    options={getFilteredExamBoardOptions({byStages: searchStages})}
+                    onChange={(s: MultiValue<Item<ExamBoard>>) => selectOnChange(setSearchExamBoards, true)(s)}
+                />
+            </Col>}
+        </Row>
+        <Row>
+            <Col className="d-flex flex-column col-12 col-xl-3 mt-2">
+
+                <FeatureFlagWrapper flag={FeatureFlag.ENABLE_SCI_BOOKMARKS}>
```
sjd210: Shouldn't this be sci-only?

src/scss/common/list-groups.scss
```diff
@@ -46,3 +58,19 @@
     }
   }
 }
+
+.list-view-border {
```
sjd210, line 62: Moving this to `/common` has added border lines to the main Question Finder flat layout on Ada. The dividers look good on the Quiz Builder, but we should probably make them "builder"-`alviType` specific to avoid them spilling over.

sjd210 (changes requested):
VRTs look good except for the missing vertical gap that now exists between questions/icon buttons that I mention below - but with this now having a unique ALVI type I think it would be good to see some questions actually in the list for us to test regressions against (perhaps even with a wider VRT than standard - to show off the full ALVI with the draggable item interface?)

src/app/components/pages/GameboardBuilder.tsx
jacbn: I'm okay adding this as it's a slight improvement, but the reason I didn't spend more time cleaning this up is that at <=370px you can't tell there's a delete option at all, regardless of the approach. If we were going to fix this for `xs`, it'd subsume this change. The issue is that the delete button probably ought to go underneath and not be vertical in this case; it's just a lot of work for a page that is already hard enough to use in mobile.

src/app/components/elements/list-groups/AbstractListViewItem.tsx
jacbn: I developed these under the assumption that `bookmarksFeatureFlag` could only be enabled for Sci; I think you've been testing these under the assumption that it will be replaced with `true` once merged. I'll add some `isPhy` checks to the bookmarks usages now so that we're on the same page.

src/app/components/elements/modals/QuestionSearchModal.tsx
jacbn: As above

src/app/components/elements/list-groups/AbstractListViewItem.tsx
sjd210: Mm, your assumed way doesn't make much sense to me. We have feature flags so that we can develop things in stages and test how other things interact when doing so. Which site the feature is present on is *part* of that interaction, even if a mild one (e.g. What if enabling this feature flag *unintentionally* broke something on Ada? We'd still like to know about that, and so should keep Ada in the correct state to see if that would happen)

sjd210 (approved):
I made one more slight change so that the new mw-max-content only applies at xs screens where the ALVI width issues were appearing.

+ I'm making a separate card to improve the VRTs here to actually show list items

Otherwise looks good, and we've decided this should be safe enough to release right now, so this is finally getting merged 🫡

PR #2151, opened 2026-05-12
Restores proper bundle split for inequality (notably p5)

jacbn:
After some recent inequality improvements, inequality / p5 were being imported in the main bundle. This introduces `inequalityUtils.ts`, a utils file for inequality components, that imports all the required pieces but importantly is only imported by the lazily-loaded inequality components.

Before (note big green blob on left):
image

After:
image


sjd210 (approved):
I agree that the bundles are separated, and the symbolic/graph sketcher questions all seem to work fine 👍 LGTM

PR #2159, opened 2026-05-15
Allow dropdown inputs to be manually enabled for DnD questions for accessibility 

sjd210:
Adds an accessibility setting to _"Enable non-dragging alternative inputs"_. For now this only affects cloze/drag-and-drop questions, but is defined broadly enough that we can use the same setting for other inputs in the future if they apply (e.g. depending on how we implement reorder question accessibility).

Also adds a `Dropdown`/`Drag and drop` override toggle to all question and quiz pages containing these input types in the top right, equivalent to the context switcher on concept pages. The toggle is defaulted to `Dropdown` if the user has enabled the accessibility setting or is on a mobile device, and `Drag and drop` otherwise. If manually toggled, this overrides all other settings and the toggled input mode is enabled instead (e.g. allowing drag and drop at xs screens, if wanted).

I'm not 100% confident on things like:
- if we should hide the toggle for logged-in users with the new setting disabled (so they'd see no external change)
- the exact placement of these toggles - there could be an option to hide it? or we could add it to the sidebar instead? 🤷‍♀️

Thoughts welcome!

~~Also requires https://github.com/isaacphysics/isaac-sops-config/pull/22 for new accessibility setting to work.~~ Merged!

---

~~**Known problem:** If a question is using dropdowns on first page load, then the first time drag and drop is switched to the dropzones will not load properly. This is a pre-existing issue (try some incremental page width changes on a live question) but is surfaced a lot more heavily here. This is sure to be some portalling shenanigans, and I don't think we should merge this without fixing it, but the rest is ready to review now.~~


src/app/components/content/IsaacDragAndDropQuestion.tsx
```diff
@@ -134,8 +136,24 @@ const useAutoScroll = ({active, acceleration, interval}: {active: boolean; accel
     }, [active]);
 };
 
-const IsaacDragAndDropQuestion = ({doc, questionId, readonly, validationResponse}: IsaacQuestionProps<IsaacDragAndDropQuestionDTO, DndValidationResponseDTO>) => {
+export function useDragAndDropAccessibility() {
```
jacbn: Since we use this on Cloze qs too, I'd move this hook out into `/services/accessibility.ts` (or a new dnd-specific one if you prefer) – this is where generally useful hooks or those that use actions/reducers/slices etc are _supposed_ to live.

The 'why' is that we shouldn't _really_ ever be importing question types inside other question types, for two reasons – one, all questions are lazily loaded, so it increases the page load time slightly if we need to import from another lazy file, and two, it avoids the possibility of circular dependencies.

Of course, none of this is to say that these things don't happen in the codebase. But we ought to stick to good style where we can!

jacbn (commented):
This is looking really good! I have one comment here about keeping hooks inside `/services` but beyond that my only consideration here is where and how we display the toggle.

As we've discussed, perhaps the most important thing for accessibility across the site is that we are consistent in how we surface accessibility features, so that using a new question type does not feel completely alien to e.g. a screenreader user.

Having a toggle at the top (which may be skipped if the user jumps to a specific section via a hash anchor) for just this question type does feel a little specific, though I can see it having value, particularly for tablets. To this end, I'm very happy for this feature to exist in the accessibility settings (and as you mention, it can remain on the page for non-logged in users), but _if our plans for accessibility on other question types do not include such a toggle_, I think I'd prefer we improve the accessibility on Cloze by allowing click-to-select and click-to-drop, so that we remain consistent in that all questions can be answered accessibly without requiring a toggle elsewhere on the page.

I'd love to be able to ask an accessibility panel about how they feel about which method is most accessible, as I can only base this off being consistent with other questions... :/

sjd210:
The main changes since last review:
- Move DnD accessibility hook inside `/services`
- Move _"dropdown/drag and drop"_ toggle inside InaccessibleContentWarningAlert, and colour it to match
  - This allows the flexibility of users being able to both control the default mode and whether they see the toggle separately
  - (I don't think I agree that this is the _best_ choice, but I think this is an acceptable one and that we're not going to come to an easy agreement otherwise)
- Fix portalling issue (🤞) on switching between input modes by putting the main `setPortalFuncs()` inside a `useEffect`, giving it the potential to update on every render based on the contents of `newPortalFuncs`.
- Enable accessibility tags for quizzes
  - (Note: The ViewQuizzes page uses ALVIs but the MyQuizzes page doesn't. The tags will only show externally on ViewQuizzes - perhaps we should amend this)

See commit messages for other small things.

src/app/components/elements/alerts/InaccessibleContentWarningAlert.tsx
```diff
@@ -1,11 +1,37 @@
...
+
+    return <div className="input-mode-toggle">
+        <span>Question input mode</span>
+        <Spacer />
+        <StyledToggle
+            checked={dragAndDropEnabled}
+            falseLabel="Dropdown"
+            trueLabel="Drag and drop"
+            onChange={(e) => {toggleDragAndDropEnabled(); e.stopPropagation();}}
+        />
```
jacbn, line 20: We ought to consider improving the styling for this in dark mode – both are accessibility features and having both on is currently extremely illegible!

1: Input mode toggle
Image

2: Drop zone button replacement
Image

src/app/components/elements/markup/portals/utils.ts
```diff
@@ -41,27 +41,32 @@ const portalsInHtmlHookBuilder = (hookList?: PortalInHtmlHook[]): PortalInHtmlHo
...
     });
 
-    if (!isEqual(newPortalFuncs, portalFuncs)) { // ignore reference inequality (guaranteed by construction – infinite re-render without) so long as the contents are the same
-        setPortalFuncs(newPortalFuncs);
-    }
+    useEffect(() => {
+        if (!isEqual(portalFuncs, newPortalFuncs)) {
+            setPortalFuncs(newPortalFuncs);
+        }
+    }, [newPortalFuncs, portalFuncs]);
```
jacbn, line 69: This doesn't seem to have fixed the portals issue – I still get missing inputs after changing from Dropdown => DnD for the first time since page load:
Image

On Ada, it seems to keep happening until I toggle _twice_...

src/app/components/elements/markup/portals/InlineDropZones.tsx
jacbn, line 105: A subtle issue, but switching from DnD <=> drop zones clears the `nonSelectedItemIds`, meaning that items that have been used appear to be selectable again despite not doing anything when clicked. For example, in the following image, 3N should be greyed out, as it is used prior:

Image

src/app/components/elements/panels/UserAccessibilitySettings.tsx
```diff
@@ -49,6 +48,20 @@ export const UserAccessibilitySettings = ({ accessibilitySettings, setAccessibil
...
+                    onChange={e => {
+                        setAccessibilitySettings((oldDs) => ({...oldDs, NON_DRAGGING_INPUTS: e.target.checked}));
+                    }}
+                    color={siteSpecific("primary", "")}
+                    label={<p>Enable non-dragging alternative inputs</p>}
+                    id={"non-dragging-movement"}
+                    aria-describedby="non-dragging-movement-helptext"
+                    removeVerticalOffset
+                /></b>
+                <p id="non-dragging-helptext">{`Enabling this will allow you to use alternative input methods that don't require dragging for certain question types (e.g. drag-and-drop).`}</p>
```
jacbn, line 62: I do find this a little confusing as is.

"_Enabling this will **allow** you to use..._" makes it sound like you will be given the _option_ to use alternative input methods for DnD questions, presumably somewhere on the question page. But if you haven't selected "Warn about inaccessible content" too then you aren't given any options, and the questions are just forcibly set to using dropdowns with no indication that anything has changed. I was expecting the toggle to exist in all situations, but accessible from within the banner if said banner exists (such that all accessibility-related content is in the same place), or outside above the (entire) question otherwise. It could even still have a warning banner in the case where "Warn about inaccessible content" is off, for consistency's sake, but I'm not sure what the text would say in that case.

Speaking of things we disagree with in your latest comment, I disagree with the solution to use an in-question toggle _at all_. I truly believe the chances of needing to switch an accessibility setting once you have set them initially is incredibly negligible. Why do content insist on the option there on _every_ question?? It's messy for us having to consider situations like this and just clutters the accessibility tree... ugh.

jacbn (changes requested):
I think this still needs some further refinement – I've listed a few points below. For the last one, as we both disagree with content's decision(s) on how the toggle should work / where it should be, maybe we need to better convince them otherwise? :/

src/app/components/elements/panels/UserAccessibilitySettings.tsx
sjd210: On the latter point, I think they see this as both an accessibility setting AND a convenience one. Like I remember them raising that for DnD questions taller than the user's screen, it would be helpful to be able to enable dropdowns for that question in particular - but that they might still want dragging for a more reasonably-sized question. That does sound more like a problem with the question content being too tall, but I can at least see where the idea is coming from.

src/app/components/elements/markup/portals/utils.ts
sjd210: This is a problem with me poorly cleaning up the messy code I had after my load of portal testing :facepalm:

The other half of this should be reconstructing the `portalFunc` callback when the toggle is used by adding `dragAndDropEnabled` as a dependency - which is what actually triggers the `useEffect`. 

NOW it should hopefully be fixed...

src/app/components/elements/alerts/InaccessibleContentWarningAlert.tsx
sjd210: For 1, we have an (incomplete) spectrum of misc orange colours that I had already been thinking of adding this to - but it isn't inverted as part of the dark theme. The colours are barely used, but for future cases like this where they _would_ be used we probably should incorporate them into the colour shifting so we don't get problems like this. I'll work on that.

For 2, giving them backgrounds at all was intended to fix an issue with the dropdown items overlapping with images. Of course this does still work with a darker background rather than white, but it no longer blends into the background of most images. This is _probably_ fine, but just to note.

image   image

src/app/components/elements/markup/portals/InlineDropZones.tsx
sjd210: I'm not entirely sure why, but I think this is effectively the same issue as the portalling one. I can reproduce it without the portal changes implemented, but not with them. I'm tentatively resolving it on those grounds, but let me know if you can still reproduce it!

src/app/components/elements/panels/UserAccessibilitySettings.tsx
sjd210: There was no objection to me saying that we've decided we need to make the setting universal-only for logged-in users, so I've just done that. It's much nicer for our accessibility tree.

I've kept accessibility warnings for logged-out users because I think these are suitable to have in general anyway, and am keeping `access:motor` on as a tag for all DnD/cloze questions to highlight the fact that they need a setting in order to be accessible (I've made a card to add a content editor warning when the tags aren't added for certain question types to go along with this). We could _maybe_ also change the wording of the warning to link to the setting for those with it disabled, to not entirely lose the idea of the per-page toggle?

src/app/components/elements/alerts/InaccessibleContentWarningAlert.tsx
jacbn: It might be possible to get around this by defining certain regions (figure containers, image containers) that we purposefully don't do any dark mode styling for to _explicitly_ have `[data-ld-theme="light"]`. Then the buttons can use the regular lightness theme mixin; they'll have a white background here.

jacbn (approved):
Happy to get this in now. There's a small point in a comment above, but I'd rather turn this into a card and do it for next week just to get this in.

PR #2161, opened 2026-05-18
Change default incorrect feedback for inline regions

machacekj0:
The content team requested that particularly for string match inline questions, the word "working" wasn't appropriate.


axlewin (approved):
Looks good to me! 👍

PR #2168, opened 2026-05-22
Enable accessibility warnings on Ada

sjd210:
- Adds icon to accessibility warning banner on Ada (since there is no question metadata box to otherwise display the icon in)
- Enables accessibility icons as a content property tag for Ada question ALVIs
~~- Fixes bug with `useDeviceSize` on Ada, which is currently calculating size as if ALL pages have sidebars. This now only applies to the pages under the My Ada branch (relevant here because this was causing the ALVI titles to wrap too early)~~


barna-isaac:
It seems to me that `useDeviceSize` is still buggy. For example, on the Ada My Progress page, on a viewport whose actual size is 786 px and does not get the sidebar, it reports the viewport as 566px wide, because it mistakenly believes the sidebar is open.

barna-isaac:
I'm happy with the changes otherwise. @sjd210 , if you'd prefer, I could also create a separate card for fixing the bug, and then this can be merged immediately.

sjd210:
@barna-isaac On closer look, the problem isn't what I assumed from your description - but it still has a simple enough solution that I'd still rather just include it here.

The problem was the same `deviceSize` being used both to decide WHEN the sidebar appears and WHAT the device size is when the sidebar is present. At a md-lg range, there was no way to make these both make sense simultaneously - so we must record device size separately when including the sidebar and not. I've made a new `useAbsoluteDeviceSize()` hook for this, which should be used as `deviceSize` for areas that change relative device size (just the My Ada sidebar right now, but this could eventually change).

_Note: This changes the effective breakpoint for sidebars very slightly. Fortunately (md + 220 = 988px) and (lg = 992px) screens are of comparable enough size that this makes practically no difference, and if anything it's suitable that this now actually does align exactly with a breakpoint._

barna-isaac:
@sjd210, I think the new `absoluteDeviceSize` hook is a good idea, and I did wonder whether it was confusing to use `useDeviceSize` in the sidebar, which itself affects the value returned by `useDeviceSize`. I also quite like the new `pixelsToDeviceSize` function, which nicely cleans up some duplicated logic!

I have noticed one functional change compared to what we currently have on adacomputerscience.org, and I wonder if this is intentional. This only effects smaller screens that still get the horizontal sidebar, eg., a viewport that uses a width of 1010px. On any screen that gets the horizontal sidebar (eg: https://adacomputerscience.org/assignments), the sidebar first appears as expanded, but clicking any link now collapses it. I think this behaviour is only meant for the dropdown-style vertical navigation that we show on mobile-sized screens (eg: width 604px), but with the changes here, the horizontal sidebar also gets it.

I think this is caused by the `onclick` handler in `MyAdaSidebar`.

barna-isaac:
I also like that now that we've moved Ada to use `lg` breakpoints, `SidebarLayout` explicitly toggles between column and row layouts using JS.

However, another piece of changed behaviour is that as we change the viewport width to 988 from 987, the sidebar still does not yet appear, but the site shifts to the layout with margins that we previously started using at the same time as we started showing the sidebar. I think this is quite subtle, and the site could use the margins at this point anyways, but we probably shouldn't introduce an extra layout. Still, if you don't think this is worth fixing, I'm okay merging the PR as is. 

I think this might be a side-effect of the new sidebar breakpoints?

sjd210:
@barna-isaac On your first point, I just missed that the breakpoint had previously been on `below['sm']` before rather than `below['md']` and so over-corrected up two breakpoints rather than just one 🤦. Good spot

---

Your second point seems to be some pre-existing strange behaviour. If you go to the live site right now, and switch between 987/988px, you'll see some elements/margins shift around when relying on bootstrap, even outside the My Ada pages (e.g. the text at the top of the homepage, the margins on the question finder) and then some FURTHER changes happening at 992px when we've defined breakpoints ourselves (e.g. the background image of the question finder appearing _(which incidentally is appearing in front of the search bar, for somewhat related reasons)_ ).

This wasn't introduced by this branch, but is a lingering from the initial sidebar implementation here: https://github.com/isaacphysics/isaac-react-app/blob/032ab40d2a9b3051df8a039205f208c7290cdb98/src/scss/cs/isaac.scss#L298-L306

Ideally we should be using the default `$grid-breakpoints` on most pages, and then some edited ones only on pages with the sidebar, but these being SASS variables, we can't simply override them since they apply only at compile-time.

I think the next simplest solution is to reset this variable to its default values, then manually review all the breakpoint-specific bootstrap classes within any pages in the My Ada section and increase them where relevant, but we could also redefine all the breakpoint-based bootstrap ourselves and use a system like we've implemented for darkmode to switch between these (which could even allow adapting to whether or not the sidebar is expanded).

~~The current state is also changing the widths at which layouts shift, which is causing some other oddities, so I'm not a fan of it. That said, this is pre-existing and definitely sufficiently unrelated to this branch so I think this is fine to merge for now, but I'll make a card to fix this and it should be dealt with soon.~~

Looking a little further into this, it's a bigger issue than I had initially realised. Since I'm about to be away for ~2 weeks, I think the best thing for now is for us to just merge the accessibility warning work without worrying about the strangely wrapping ALVI title, then deal with the entire My Ada/breakpoints issue later all in one batch.

barna-isaac:
@sjd210, I haven't realised this was pre-existing, although I've now checked the adacomputerscience.org landing page and you're correct that it too shifts layouts at 987/988px. Keen to here what the bigger issue is that you've eventually uncovered.

I do agree the card has exceeded its original scope, and a return to that is a good idea. I'll review just the original changes and merge those, thanks for the reset.

barna-isaac:
Looks all good!

To be safe, I've checked what would show if a question had accessibility tags without an `icon` attribute, and the result would be a warning that looks like:

Screenshot 2026-06-08 at 12 00 31

Whereas normally it is:
Screenshot 2026-06-08 at 12 01 26

Approved, merging.

PR #2171, opened 2026-05-26
Test fixes for flaky question types VRT

jacbn:
Attempts to fix the /questions VRT that has been flaky for some time. I have rerun the VRTs 5 times on the most recent commit with 100% success.

Not sure which combination of these changes actually fixed it, but from repeatedly looking at diffs it seems that preferring a `<Spacer/>` over `ms-auto` fixed the DnD test and ensuring Cypress starts from the top of the page (which `cy.scrollTo('top')` does _not_ accomplish 🙄) while all elements were unfocused before taking a screenshot gave the most general consistency improvement, seemingly fixing issues at the bottom of the page. I could believe this fixes things as said bottom of the page was where the viewport was before it was taking the screenshot – if there is an issue with the pre-screenshot scrolling in Cypress, pre-jumping to the top would fix this.

The other changes are improvements regardless of whether they helped, so I'm happy to keep everything here.


axlewin:
These changes all seem sensible, hopefully this will finally fix the VRT problems 🤞 

The only issue I've found is with the `Button` changes in `IsaacQuestion` - the "Check my answer" button is now smaller and oddly aligned:
image

axlewin (approved):

PR #2181, opened 2026-06-01
Fix flaky /questions VRT

jacbn:
Makes changes to the `/questions` VRT to reduce the flakiness of the tests.

After significant testing, it seems to me that the issue at the very least _involves_ a bug with Cypress' `cy.screenshot` image stitching. On the original VRT, the "erroring" question type was always near the bottom of the (admittedly very large) screenshot. On removing said question type, the error simply moved up to the next-bottom question type, suggesting nothing wrong with the questions themselves but something with the image stitching.

As a workaround, I have moved each question type into an individual file. I've left the old approach around as it would be much quicker to use that, but this approach may come with its own advantages too (e.g. more easily see which question type has broken).

There are additionally changes to the report icon and figures; I did note that the _type_ of VRT failure I was seeing after implementing these (but before splitting into separate files) had changed, in that the height of the generated image was different rather than changing "in-place". This seems to be an improvement for debugging in future so I am happy to keep.


src/app/components/elements/ReportAccordionButton.tsx
```diff
@@ -38,13 +40,17 @@ export const ReportAccordionButton = ({pageId, sectionId, sectionTitle, sectionI
...
-        aria-label="Report a problem (opens in new tab)"
-        title="Report a problem (opens in new tab)"
-        onClick={() => {
-            logAccordionReport();
-            window.open(`/contact?preset=contentProblem${getContactFormParams()}`, "_blank");
-        }}
-    />;
+    return <div className="d-flex w-100 justify-content-end mb-4 px-4">
+        <button
+            className="accordion-icon btn-action bg-transparent btn-blank p-0 wf-2 hf-2 vertical-center"
```
axlewin: `accordion-icon` doesn't exist anymore. This button seems to look ok without it anyway, though?

src/test/pages/__image_snapshots__/ada/Question types' regression test page (part 1) should have no visual regressions #0.png
axlewin, line 1: Do we need these "part 1/2" baselines?

jacbn: Ah, yeah, forgot these. I wanted to keep the original big one in case we ever bring the `it.skip`ped test back but these won't ever be used again.

axlewin (approved):

PR #2193, opened 2026-06-04
Add "back to book" button on gameboards

axlewin:
If a gameboard is accessed from a book page, show a "Back to book section" button in the sidebar that links back to the book page.


src/app/components/elements/BookPage.tsx
```diff
@@ -49,7 +48,7 @@ export const BookPage = ({ page }: { page: IsaacBookDetailPageDTO }) => {
             <div className="mt-3 mb-7 list-results-container p-2">
                 <ListView
                     type="gameboard"
-                    items={convertToALVIGameboards(page.gameboards as GameboardDTO[])}
+                    items={convertToALVIGameboards(page.gameboards as GameboardDTO[], window.location.pathname)}
```
sjd210: I really don't like us embedding so much of one url into another url. The slashes make it difficult to figure out where the actual pages url ends and this one begins at a glance, and it just makes things unnecessarily long.

We could certainly slice the `/books/` from the front since these are only ever meant to link to book pages so we definitely know to include it. We could maybe then also split the `?book=$book/$section` into `?book=$book&section=$section` although then that's a tradeoff between clarity and length so I'm not 100% sold. I'll leave that to you.

src/app/components/elements/list-groups/ListView.tsx
```diff
@@ -166,16 +166,17 @@ export const QuizListViewItem = ({item, isQuizSetter, useViewQuizLink, ...rest}:
     />;
 };
 
-type ALVIGameboard = GameboardDTO & {type?: string};
+type ALVIGameboard = GameboardDTO & {type?: string} & {linkedBookPath?: string};
```
sjd210: Any reason we call it `bookPagePath` in other files but `linkedBookPath` here? It feels easier to track use if we stick with one consistent name

sjd210 (changes requested):
I wasn't here for the tech meeting discussion, but I'd imagine that content also want the button to still be on the question deck when navigating to a question, then back to the deck. There are a few different links that do this, but it should mostly just be propagating the url parameter in those different places too, which seems reasonable.

Also this definitely wasn't requested, but I imagine this would be also be useful for concept pages which are also very commonly linked from book pages. Maybe something to mention?

src/app/components/elements/sidebar/GameboardContentSidebar.tsx
sjd210: Currently navigating using the bottom-page links will maintain the url parameters, but these sidebar links don't. We'll need to add `linkedBookUrlParams` as a prop to `QuestionLink`, but otherwise this should be quite straightforward.

src/app/components/pages/Gameboard.tsx
```diff
@@ -45,7 +45,10 @@ export const Gameboard = () => {
     const queryArg = user?.loggedIn && !isTeacherPending(user) ? undefined : skipToken;
     const {data: assignments} = useGetMyAssignmentsQuery(queryArg, {refetchOnMountOrArgChange: true, refetchOnReconnect: true});
     const thisGameboardAssignments = isDefined(gameboardId) && isDefined(assignments) && isFound(assignments) && (assignments.filter(a => a.gameboardId?.includes(gameboardId))) || undefined;
-    const bookPagePath = new URLSearchParams(location.search).get("book") ?? undefined;
+
+    const urlParams = new URLSearchParams(location.search);
+    const linkedBookSection = [urlParams.get("book") ?? "", urlParams.get("section") ?? ""];
+    const linkedBookSectionUrlParams = `?book=${encodeURIComponent(linkedBookSection[0])}&section=${encodeURIComponent(linkedBookSection[1])}`;
```
sjd210: If there is no `book` or `section` url parameter, this should be empty. Currently its `?book=&section=`.

sjd210 (changes requested):
Hate to be feature-creeping this - but for consistency, we probably _should_ also apply this to wildcards. 

`Generic` and `Concept` pages as wildcards in a gameboard have a `GameboardContentSidebar` for us to propagate this through, plus it'll need to be added to the `ShortcutListViewItem`.

axlewin:
> we probably _should_ also apply this to wildcards.

~~Hmm. I'm not against this in principle, but I'm not sure it works equally well for all wildcard types - for example where the wildcard is a generic content page we'd end up with something like this, which looks a bit odd.~~ (Ignore me, I was using the wrong sidebar...yes, we ought to do this for wildcards too.)

sjd210 (approved):

PR #2194, opened 2026-06-04
Remove routes to now-unpublished pages

axlewin:
Also remove `/chemistry` since this goes to a different landing page now.


sjd210 (approved):
A thought: Perhaps some of the old, now irrelevant routes from the other closed PR now *should* be unpublished and then we get rid of their routes - but a decision would have to be made there and this is definitely responsible anyway

PR #2195, opened 2026-06-04
Restore margin to "check my answer" outside accordions

axlewin:
Fixes a recent regression that made this button full-width on small screens.


axlewin:
(this will presumably break every single question type vrt, sorry!)

PR #2198, opened 2026-06-08
Combine Set Assignments & Schedule into Manage Assigned Work

jacbn:
Combines Set Assignments and the Assignment Schedule into one page for universal assignment / test management, called Manage Assigned Work (`/assigned`).

Note that this is a feature flagged branch over the My Saved Decks branch, which is *not* itself a feature flagged branch. We ought to keep both branches separate for now, then when merging MSD, merge this into main after.


barna-isaac:
@jacbn, I think this can go into `main`, and we don't need to hold up the release until we can merge Part 3 (although I hope we can merge that too until next week).

### Switch off
With the feature switch disabled (its default state on non-dev environments), I can only see two changes (both are Isaac-only).
1) On the Set Assignments page (which we'll deprecate for Isaac), we now show "Assign" buttons on the cards instead of "Assign / Unassign" buttons (regardless of whether the assignment has been set already).
2) On a few places with styled selects, such as the country selector or the wildcard picker on the question deck builder, the dropdown now gets a dark background when in dark mode.

That's it, up to you whether 1) needs to change (I don't think so).

### Switch on
With the feature switch enabled, teachers get the new Set & manage work (`/assigned`) page, which combines the old Assignment Schedule (`/assignment_schedule`) and Set assignments` (/set_assignments`) pages. It also shows tests, which Assignment Schedule didn't.

I have one major finding regarding the page, and a little nit-pick. However, since the feature is behind a switch, I leave it to you whether you address any of these on this PR or in part 3. (I also suspect you might have already addressed these in part 3.)

The major finding is about test cards. Clicking the title tries to take the user to the quiz attempt page. However, because the user is a teacher rather than a student, these pages will be typically unavailable to them, so these pages now just show errors.

The nitpick is just about the URL `/assigned`, which feels awkwardly incomplete? Maybe we could just do the obvious thing and use `/set_manage_work`?

### Next steps
I still need to look at why the VRT's failed. Seeing that you've reworked the jest tests on the Part 3 branch, I'm planning to just limit these to Ada. This way, they shouldn't break any builds on `main`.

barna-isaac:
@jacbn , please tell me whether there are any changes you'd like to make, having considered the comments I've made above. Please also tell me whether you approve of the commit where I've disabled some jest tests, and the changes to VRT baselines.

If you approve, I think this is ready to merge!

jacbn:
Thanks for spotting the test link! Have fixed.

> The nitpick is just about the URL /assigned, which feels awkwardly incomplete? Maybe we could just do the obvious thing and use /set_manage_work?

We were going to ask teachers at the symposium for a new name, so it probably won't be set / manage work – perhaps we can hold off on this thought until then?

barna-isaac (approved):
Thanks for the fix. I've a few more findings, but will put these under Part 3 as they only affect the new `/assigned` page, which is still feature flagged as of this PR.

PR #2201, opened 2026-06-09
Fix Ada sidebar container overflows

jacbn:
Fixes overflow caused by changes to the Ada sidebar container.

Before (overflows to the right):
image
After:
image

---

Also fixes the bad float position on the share link causing certain pages without toptext to appear too small, just by hiding it for non-content pages.

Before:
image
After:
image


sjd210 (approved):
I think it's fine to stop showing action buttons on non-content pages for Ada (we were already quite inconsistent on which of these pages showed them), but for the record I think that the bad float position was just because you were loading Ada frontend with Isaac content - so the content fragment was missing.

I thought about this before, and this saves us from the scenario where the content in the fragment gets emptied without us being consulted, or if it fails to load - so its at least still a little helpful for the desired case anyway.

(+ the My Ada overflow fix is definitely a win!)

PR #2207, opened 2026-06-11
Dark mode additions vol 2

jacbn:
Fixes the remaining documented issues with dark mode.

If any *additional* dark mode failures are found during testing, we probably ought to fix them here this time – but this by no means is required to be the last set of dark mode changes.


axlewin (commented):
Nice, I think this is pretty much ready to un-feature-flag 👍

A few minor outstanding issues I came across while testing:
- The (i) icon on the email verification banner isn't visible
- The contrast of the red 'Overdue' text on dashboard assignment cards is too low
- On the markbook, the colour code box for "Not attempted" still blends in with the background (I realise this isn't straightforward to fix by just editing the colour, but it looks a bit odd; maybe a border around the box like we have in light mode would work?)

Otherwise I think this is good to merge.

axlewin (approved):

PR #2208, opened 2026-06-12
Use modified /questions endpoint to search bookmarks

jacbn:
Rather than relying on front-end filtering of the entire list of bookmarks, instead use the [modified](https://github.com/isaacphysics/isaac-api/pull/799) `/questions` endpoint to allow our ES filtering to run this in precisely the same manner as for other uses of the endpoint.


axlewin (approved):
Looks good.

Not sure if you want to hold off on merging this until the corresponding API changes have been through a release cycle; maybe since this'll only affect people with the feature flag on we don't need to wait.

PR #2212, opened 2026-06-17
Fix Ada sidebar breakpoint inconsistency

jacbn:
Attempts to resolve the sidebar issues described in https://github.com/isaacphysics/isaac-react-app/pull/2168, caused by the sidebar using a modified page-width-obtaining function that compensates for the sidebar existing on the page – when it obviously should not since it *is* the sidebar!

This is very similar in approach to https://github.com/isaacphysics/isaac-react-app/pull/2168/changes/25ae05eb42c7e4b10b77784faa6965302b81a66b, but fixes a bug where `shouldIncludeSidebar` was expecting the sidebar to be `>=992px` when it should be `>=(768px + $sidebar-width)` = `>=988px`. This explains the buggy behaviour at precisely 988-991 pixels on the original.

This is, however, intentionally less fully-featured; for example, `useDeviceSize` does not change according to whether the sidebar is shown (i.e. `...&& Object.values(MyAdaTabs).some(tab => window.location.pathname.includes(tab.url)) && ...`). This does mean that all Ada pages are affected by the modified widths, not just those that have a sidebar.

The reason for not doing this is that right now, Bootstrap `d-md-x` etc classes can only represent exactly one, non-changing value owing to SCSS being compiled, and if we allowed `useDeviceSize` to mean two different things, there would be inconsistency between Bootstrap and e.g. `above`/`below`. I believe sitewide consistency between these is too fundamental *not* to cause a slew of bugs if we do not keep it as is. 

Of course, if there was a way for Bootstrap's meanings to differ in different situations, I would be all for it. I am not particularly keen to spend much time on it, however, as BS have announced containerised Bootstrap will be coming in v6 which would solve all of this, and we'd only be reinventing the wheel (very awkwardly, since we can't modify BS ourselves...).

### DIFF
--- a/src/app/services/device.ts
+++ b/src/app/services/device.ts
@@ -34,7 +34,7 @@ const descDeviceSizes = [DeviceSize.XXXL, DeviceSize.XXL, DeviceSize.XL, DeviceS
 
 export const useDeviceSize = () => {
     const getSize = (): DeviceSize => {
-        const shouldIncludeSidebar = isAda && window.innerWidth >= 768;
+        const shouldIncludeSidebar = isAda && window.innerWidth >= 988;
         const width = window.innerWidth - (shouldIncludeSidebar ? 220 : 0);
         if (width >= 1800) return DeviceSize.XXXL;
         else if (width >= 1400) return DeviceSize.XXL;
@@ -77,6 +77,20 @@ export const useDeviceHeight = () => {
     }, []);
 
     return windowHeight;
+};
+
+export const useFullSidebarLayout = () => {
+    const shouldUseFullLayout = (): boolean => window.innerWidth >= siteSpecific(768, 988);
+
+    const [fullSidebarLayout, setFullSidebarLayout] = useState(shouldUseFullLayout);
+
+    useEffect(() => {
+        const handleResize = () => {setFullSidebarLayout(shouldUseFullLayout());};
+        window.addEventListener('resize', handleResize);
+        return () => window.removeEventListener('resize', handleResize);
+    }, []);
+
+    return fullSidebarLayout;
 };
 
 export const useNavbarExpanded = () => {
--- a/src/app/services/device.ts
+++ b/src/app/services/device.ts
@@ -80,7 +80,7 @@ export const useDeviceHeight = () => {
 };
 
 export const useFullSidebarLayout = () => {
-    const shouldUseFullLayout = (): boolean => window.innerWidth >= siteSpecific(768, 988);
+    const shouldUseFullLayout = (): boolean => window.innerWidth >= siteSpecific(992, 988);
 
     const [fullSidebarLayout, setFullSidebarLayout] = useState(shouldUseFullLayout);
 
--- a/src/app/components/elements/layout/SidebarLayout.tsx
+++ b/src/app/components/elements/layout/SidebarLayout.tsx
@@ -1,7 +1,7 @@
 import React, { useContext, useEffect } from "react";
 import { Col, ColProps, RowProps, Offcanvas, OffcanvasBody, OffcanvasHeader, Container, Accordion, AccordionItem, AccordionHeader, AccordionBody } from "reactstrap";
 import classNames from "classnames";
-import { above, isPhy, siteSpecific, useDeviceSize } from "../../../services";
+import { isPhy, siteSpecific, useFullSidebarLayout } from "../../../services";
 import { mainContentIdSlice, selectors, sidebarSlice, useAppDispatch, useAppSelector } from "../../../state";
 import { ContentSidebarContext, SidebarContext } from "../../../../IsaacAppTypes";
 import { AffixButton } from "../AffixButton";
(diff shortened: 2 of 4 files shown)


src/app/services/device.ts
```diff
@@ -34,7 +34,7 @@ const descDeviceSizes = [DeviceSize.XXXL, DeviceSize.XXL, DeviceSize.XL, DeviceS
 
 export const useDeviceSize = () => {
     const getSize = (): DeviceSize => {
-        const shouldIncludeSidebar = isAda && window.innerWidth >= 768;
+        const shouldIncludeSidebar = isAda && window.innerWidth >= 988;
```
sjd210, line 37: This still doesn't align the device size with the bootstrap breakpoints properly. Currently at an `innerWidth` between 768-987px, `useDeviceSize()` returns MD whereas `$grid-breakpoints` returns SM. Since we want them to align, we should probably just use different values directly for Isaac/Ada in each `if` case instead of this logic that overlaps sizes in a way our css does not.

Right now this causes some breakages for elements that use both bootstrap and `useDeviceSize()` in that width range (e.g. the difficulty summary not appearing on question ALVIs)

sjd210 (changes requested):
I still think this system is odd for prioritising sidebar-based screenwidths on a site that doesn't currently use sidebars on most pages, but there's definitely potential we start using them more so I won't advocate for tearing this apart (not just yet, at least). The bootstrap v6 containerisation would of course be great here, but it doesn't look like that's coming any time remotely soon.

That said, the behaviour in this branch is definitely an improvement over the current implementation so I'm definitely happy for it to go in except for one more bit of misalignment (see below).

---

There are still certainly some issues from the original implementation of things looking odd at screenwidths they weren't intended for (e.g. the bottom margin for the `#cta-container` on the homepage gets very tight), which I'm collating a list of - however they're not breaking anything so once I've finished investigating I'll make that into a separate PR.

jacbn:
> I still think this system is odd for prioritising sidebar-based screenwidths on a site that doesn't currently use sidebars on most pages

I do agree that it's odd, but with the current setup we only have two options: 
- prioritise pages that don't use a sidebar, meaning pages _with_ a sidebar use a dense, e.g. 4-column layout when a 2-column layout would better suit;
- prioritise pages that do use a sidebar, meaning pages _without_ a sidebar use a sprase, e.g. 2-column layout when a 4-column layout would fit.

The former often looks buggy; if the elements were designed with some _unenforced_ minimum suitable width, titles are not given enough breathing room and everything looks cramped (example: the 4-wide `IconCard` layout on `/dashboard`). Worse, if elements were designed with an _enforced_ min-width, they genuinely overflow. The latter, while not making the best _use_ of space, does ensure elements never overflow and will certainly have enough breathing room. I prefer the non-"buggy" approach, and I did at one point try to explain this to Ada content and they agreed, insofar as they understood it.

---

> we should probably just use different values directly for Isaac/Ada in each if case instead of this logic that overlaps sizes in a way our css does not.

I'm not sure I fully understood what you were trying to say here, but to me the output of `useDeviceSize` is correct here but `$grid-breakpoints` is wrong. i.e., we should have

```
xs: 0-575px
sm: 576-767px
md: 768-(991+$sidebar)px
lg: (992+$sidebar)-(1200+$sidebar)px
etc
```

In this way, the site remains at `md` all the way from 767px, through the sidebar expansion, and all the way to ~1212px, when we can be certain the content has certainly surpassed the 992px threshold and thus should be displayed at `lg`.

We then make use of the new, alternative method to calculate when the sidebar should expand/collapse, since this has no relation to the breakpoints.

I've made a few slight changes to accommodate this. Does this work as you'd expect now? Or do you disagree with how this should work?

sjd210 (approved):
> prioritise pages that don't use a sidebar, meaning pages with a sidebar use a dense, e.g. 4-column layout when a 2-column layout would better suit;

But the 4-column layout only exists because of the specific classes/styling that we have used, right? I would've thought it would make more sense to put some more effort in up-front to restyle things e.g. changing the `row-cols-lg-4` in the page's `IconCard` layout you mention to `row-cols-xl-4`. Am I missing something here (perhaps I'm not getting what you mean by "enforced" min-width)?

---

I don't think I explained myself clearly enough about the mismatch, however this resolves the problem in a consistent way so enough of it got through that it doesn't matter too much.

This along with the new hook fix the main issues I was complaining about in the PR where I brought this up, so this is good to go and we can bare anything else in mind for the future.

jacbn:
Do you mean you'd have preferred to restyle all the xx-[md/lg/etc]-xx props on the pages with a sidebar to effectively "shift" the breakpoint of each up so that it can absorb the 220px lost page space to the sidebar?

It would work if done right, but there are about 15 or so fully-fledged pages (lots of markbook ones) that would all need looking at. There's also the then-inconsistency in the code where if you copy or reuse a component from a sidebar page onto a non-sidebar page the breakpoints will be one step too large (c.f. board rows, tests rows, when we switch to ALVIs universally, those, ...). And finally, if content decide to use the sidebar elsewhere, there's very little in the way of "free" extensibility; you'd have to change all the breakpoints on all the new components/pages.

I just prefer the consistency with this approach. I know it makes layouts _slightly_ more sparse on tablet, but it's still entirely usable.

PR #2217, opened 2026-06-19
Fix reorder/parsons question selected item focus-outlines 

sjd210:
Currently a selected item will have its focus outline around the original item position, rather than the offset position used to indicate its selection.

This is because of an interaction with `position: fixed` causing the position of the item itself to not change while its children do. We can therefore just move the outline to the moved child to fix the visual issue.

### DIFF
--- a/src/scss/common/questions.scss
+++ b/src/scss/common/questions.scss
@@ -428,6 +428,17 @@ figure .inline-container {
         }
       }
     }
+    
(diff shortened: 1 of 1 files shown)


jacbn (approved):
The box outline is indeed more closely tied to the item now. I presume your other branch will fix not being able to indent in parsons questions?

If so, LGTM + happy to merge.

sjd210:
> I presume your other branch will fix not being able to indent in parsons questions?

Yep, already working on it!

PR #2221, opened 2026-06-22
Add sorting mechanism for total parts in Detailed Marks

jacbn:


sjd210 (approved):
Makes sense and works. LGTM 👍

PR #2222, opened 2026-06-22
Improve inline cloze item alignment on Safari

jacbn:
This feels fairly safe as this should only modify where on the line the cloze region generates. I haven't noticed any major differences on non-Safari browsers regardless.


barna-isaac:
I've tested on Firefox, Chrome and Safari, and confirm the added class makes no change on Firefox and Chrome, and fixes the alignment on Safari 🎉.

barna-isaac (approved):

PR #2226, opened 2026-06-24
Add accessibility buttons to reorder/parsons questions (+ simplify code)

sjd210:
The main objective of this branch is to make reorder/parsons questions more accessible. This has been done by:
- Adding reorder buttons (similar to the new gameboard builder) to the left side of a draggable item, allowing them to be clicked as a non-dragging pointer alternative
- Reconfiguring existing indent controls to work with keyboard controls (i.e. appearing when an item is focused, being interactable with space/enter)
- Adding list switcher buttons to the right side of a draggable item (alongside indent controls), allowing them to be clicked to move items from one list to another

While doing this, a mild refactor of these question types has also been completed so they use shared components rather than duplicated logic. This can all been found in `ParsonsDraggableItem.tsx`

---

This has all currently been done with the assumption that we will transform the appearance of these questions for all users, but this could also be put behind the same **"Enable non-dragging alternative inputs"** accessibility setting as the DnD changes for logged-in users. I think this is probably unintrusive enough to go through without the setting (and certainly the indent controls fix should regardless)


jacbn:
I've made a small change to adjust the accessible names of various elements when using VoiceOver for doing this (feel free to adjust if you want!). One thing I've noticed though is that we now have two approaches to moving items between Available Items and Your Answer, in both the DnD interface's keyboard accessibility text and the new button. The DnD interface is quite intense; it adds a whole multi-sentence string of text after you focus any draggable item (along the lines of "press space to start a drag. you can then use the arrow keys to..."). I don't know that having both is necessarily _bad_, but do I wonder if it's moreso confusing for new screenreader users?

src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -1,41 +1,39 @@
...
-    const [ currentIndent, setCurrentIndent ] = useState<number | null>(null);
-    const [ currentMaxIndent, setCurrentMaxIndent ] = useState<number>(0);
-    const [ currentDestinationIndex, setCurrentDestinationIndex ] = useState<number | null>(null);
-
+    const [draggedElement, setDraggedElement] = useState<HTMLElement | null>(null);
+    const [initialX, setInitialX] = useState<number | null>(null);
+    const [currentIndent, setCurrentIndent] = useState<number | null>(null);
+    const [currentMaxIndent, setCurrentMaxIndent] = useState<number>(0);
+    const [currentDestinationIndex, setCurrentDestinationIndex] = useState<number | null>(null);
     const canIndent = (!isDefined(doc.disableIndentation) || !doc.disableIndentation) && !readonly;
```
jacbn, line 55: We're using this condition as a check for whether we can indent relative to its parent in most cases, not just whether indenting is enabled at all. Your new buttons that use the new `canDecreaseIndentation` / `canIncreaseIndentation` don't use this directly so don't have this issue, but the old ways still allow invalid indenting of e.g. the topmost element through various means, though I'm finding it quite inconsistent in whether it works. I know this was an existing bug but would it be a quick thing to fix?

src/app/components/elements/ParsonsDraggableItem.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+interface ReorderButtonsProps {
+    index: number;
+    items: Immutable<ParsonsItemDTO>[];
+    setItems: Dispatch<SetStateAction<Immutable<ParsonsItemDTO>[]>> | ((items: Immutable<ParsonsItemDTO>[]) => void);
+    isParsons?: boolean;
+    currentIndent?: number | null;
+}
+
+const ReorderButtons = ({index, items, setItems, isParsons, currentIndent}: ReorderButtonsProps) => {
```
jacbn, line 134: There is a little issue with these in that it's now very easy to create invalid indentation by just reordering an indented item with these buttons. We had some prevention in the past whereby if you try to drag an item up from beneath another, its indentation level is limited to at most one below the new parent (although this, also, seems to work inconsistently). It's hardly a breaking issue, but given all the work we're doing on e.g. the indentation buttons to prevent invalid reorderings, this does seem a little inconsistent.

jacbn (commented):
I've made a few changes to the code to improve accessibility readouts and a little dark mode bug, but in terms of functionality this seems to work great. There are some small issues I've found below, but looking really good!

src/app/components/content/IsaacParsonsQuestion.tsx
sjd210: I've addressed both this condition's bug and the the issue with reorder buttons below in the same way, which is by checking the validity of indentations in the first place (before even setting their attempt). Trying to handle lots of individual cases of items moving between indentation levels/lists in different ways felt quite unwieldy. This way no matter how an invalid list would be formed, it gets corrected to its closest valid list.

sjd210:
@jacbn Hmm, I hadn't noticed the `@hello-pangea/dnd` built-in aria description. It is rather clunky including it on every item like that, so I've written our own as a one-off hidden note above the items similar to what we use for DnD/cloze questions. I wasn't sure how much to include because of the now two possible focus-based input schemes, but it still seems cleaner than being on every individual item when I tried navigating with NVDA.

jacbn (approved):
Looking good! Thanks for doing this 😄

PR #2228, opened 2026-06-24
Set & manage all work types

jacbn:
Extends the work done in `manage-assignments` to include tests in the schedule-like `Set / manage work` page.


barna-isaac:
@jacbn, I'll start posting findings so you have more time to address them.

❗️ The new `/assigned` page does not work for tutors. This is a regression. Tutors used to be able to access the `/set_assignemnts` and `/assignment_progress` pages, but not `/set_tests`. With the unification, they lost access to managing assignments too.

barna-isaac:
❗️ The new `/assigned` page also needs fixes for mobile? This is what it looks like on a large and a small iPhone:
| iPhone 14 Pro Max    | iPhone SE |
| -------- | ------- |
| Screenshot 2026-07-01 at 14 59 45   |  Screenshot 2026-07-01 at 14 59 57   |

barna-isaac:
❗️ On the `/assign` page, user-created decks (and possibly all other decks too?) always show in physics purple (and using the Physics tag), even if they only contain questions from a single other subject.

Filtering by subject only works on tests. I've not been able to get any user-created decks to show up when a subject filter was active. The same works fine on `/set_assignments`, even for user-created decks.

barna-isaac:
❗️ On the `/assign` page, we display an empty star icon, even when the deck has been bookmarked. The icon remains empty even if I try to bookmark the deck through this page (although the success message is shown and the deck then gets added to "My Saved Decks").

src/app/components/elements/ManageAssignedCards.tsx
```diff
@@ -58,6 +60,7 @@ export const ManageTestCard = ({quizAssignment}: {quizAssignment: QuizAssignment
         usageDisplay={{type: "progressLink"}}
         openAssignModal={() => openAssignModal(quizAssignment)}
         cancel={() => cancel(quizAssignment)} 
+        openSetFeedbackModeModal={() => openSetFeedbackModeModal(quizAssignment)}
```
barna-isaac, line 63: ⬇️ Not a huge deal, but what's gained by invoking `useManageQuizAssignments` out here and passing down its props, versus just invoking it within `TestCard`? I have the same question about `ManageAssignmentCard` and `GameboardCard`.

barna-isaac:
With `/practice_tests` renamed, we'll need to update the content at https://editor.isaacphysics.org/edit/master/content/general_pages/fragments/help_toptext_tests_student.json and https://editor.isaacphysics.org/edit/master/content/general_pages/fragments/help_toptext_tests_teacher.json so these link to the new `/view_tests` page. (There's also an event pod from 2025 but I don't think that needs updating.)

Also, we've redirected `/practice_tests` to `/view_tests` on Ada, but not on Science, is this intentional? (The path still being in place on Ada, it's less urgent there, but we should probably update Ada content as well.)

barna-isaac:
⬇️ I don't think the crumbs are too important, but when a student views feedback for a past attempt, the crumbs used to show `My tests > [name of test] Feedback`. They now show `View tests > [name of test] Feedback`, even though it's not possible to view a past attempt's feedback from the `View tests` page.

I think this is why, from the Ada suite, a testcase in `QuizAttempt` still fails.

src/app/components/elements/SaveBoardButton.tsx
```diff
@@ -17,15 +17,14 @@ export const SaveBoardButton = (props: SaveBoardButtonProps) => {
...
+    const isLinked = useMemo(() => board.savedToCurrentUser, [board]);
     const [justLinked, setJustLinked] = useState(false);
-    const isLinked = useMemo(() => board.savedToCurrentUser || justLinked, [board, justLinked]);
 
     const linkBoard = useCallback(() => {
         if (!user || !board) return;
         setJustLinked(true);
         void dispatch(saveGameboard({
             boardId: board.id ?? "",
-            boardTitle: board.title,
```
barna-isaac, line 28: do I correctly understand we used to give custom names to each gameboard we've saved? If so, it's a relief you've fixed this!

src/app/components/pages/assignment_progress/AssignmentProgressGroup.tsx
```diff
@@ -221,7 +221,7 @@ export const AssignmentProgressGroup = ({user, group}: {user: RegisteredUserDTO,
                                 </div>
                             : <div className={classNames("d-flex flex-column m-2 p-2 hf-12 text-center gap-2 justify-content-center", siteSpecific("bg-neutral-light", "bg-cultured-grey"))}>
                                 <span>You haven&apos;t {activeTab === "assignments" ? "set any assignments" : "assigned any tests"} yet.</span>
-                                <strong><Link to={activeTab === "assignments" ? PATHS.SET_ASSIGNMENTS : "/set_tests"} className={classNames("btn btn-link", {"fw-bold": isPhy})}>
```
barna-isaac, line 224: I know `/set_assignments` is still around (although not linked anywhere else), but I imagine we'll eventually remove that page completely (for Science, at least). Shouldn't we direct teachers to `/assigned` instead of `PATHS.SET_ASSIGNMENTS`?

And shouldn't we be using a constant (potentially under `PATHS`) instead of `/view_tests`?

barna-isaac:
⬇️ On the `set_tests` page, Ada used to have a dedicated layout for xs screens, which we've now lost. I think the old layout looked a little better, but the new layout isn't too bad either. On the bright side, the new layout is the same thing adopting to a smaller screen, while the old layout was a dedicated different view for mobile, and it's good we're getting rid of these.

| old    | new |
| -------- | ------- |
|  Screenshot 2026-07-02 at 17 05 10  | Screenshot 2026-07-02 at 17 05 22  |

barna-isaac:
❗️ When viewing a question deck, the star icon shows full and sparkly, even when the deck is not saved to "My Saved Decks". Clicking the button asks whether to un-save it, and a message appears about the deck having been removed, even though it's never been saved in the first page. Clicking the star button again still just tries to un-save it.

The regression is also present on `/main`, but not the live site. Maybe I've missed it while reviewing Part 2?

Also, I think it might be that it always show as full for the user that's created it, and never show as full for a user who's just tried saving it.

src/app/components/pages/Gameboard.tsx
```diff
@@ -141,14 +156,23 @@ export const GameboardContents = ({gameboard, linkedBookSectionUrlParams}: Gameb
...
+                    {user.loggedIn
+                        ? <Button color="keyline" block onClick={() => {
+                            void dispatch(saveGameboard({
+                                boardId: gameboard.id ?? "",
+                                user,
+                            }));
+                        }}>
+                            {siteSpecific("Save this deck", "Save to My quizzes")}
+                        </Button>
+                        : <Button color="keyline" block tag={Link} to={"/login"} onClick={() => {
```
barna-isaac, line 168: This will never show, because on line 144, user will be falsy for an anonymous user (I think?). At least I haven't been able to view the "Log in to save this deck" when viewing the page while logged out.

src/app/components/pages/Gameboard.tsx
```diff
@@ -112,11 +117,21 @@ interface GameboardContentsProps {
...
     const dispatch = useAppDispatch();
+    const {openAssignModal} = useSetAssignment(gameboard);
     const user = useAppSelector(selectors.user.orNull);
     const contentSummary: ContentSummaryDTO[] = gameboard?.contents?.map(q => { return {...convertGameboardItemToContentSummary(q), state: q.state}; }) || [];
     const wildCard: ContentSummaryDTO = {...gameboard?.wildCard, type: SEARCH_RESULT_TYPE.SHORTCUT, tags: [], className: "wildcard-list-view"} as ContentSummaryDTO;
     const displayQuestions = (gameboard?.wildCard && gameboard && showWildcard(gameboard)) ? [wildCard, ...contentSummary] : contentSummary;
 
+    const [searchParams] = useSearchParams();
+
+    useEffect(() => {
```
barna-isaac, line 128: ⬇️ This will show the dialog for a student, or even somebody anonymous. (A teacher or tutor might mistakenly share the URL).

src/app/components/pages/Gameboard.tsx
```diff
@@ -141,14 +156,23 @@ export const GameboardContents = ({gameboard, linkedBookSectionUrlParams}: Gameb
...
             : gameboard && !gameboard.savedToCurrentUser && <Row>
                 <Col className="mt-4" sm={{size: 8, offset: 2}} md={{size: 4, offset: 4}}>
-                    <Button tag={Link} to={`${PATHS.ADD_GAMEBOARD}/${gameboard.id}`}
-                        onClick={() => setAssignBoardPath(PATHS.SET_ASSIGNMENTS)}
-                        color="keyline" block
-                    >
-                        {siteSpecific("Save this deck", "Save to My quizzes")}
-                    </Button>
+                    {user.loggedIn
+                        ? <Button color="keyline" block onClick={() => {
```
barna-isaac, line 160: Doesn't this just duplicate the star button? And what if the deck's been saved already?

src/app/components/pages/GameboardBuilder.tsx
```diff
@@ -218,7 +219,13 @@ const GameboardBuilder = ({user}: {user: RegisteredUserDTO}) => {
                 title: `${siteSpecific("Question deck", "Quiz")} ${gameboardId ? "created" : "creation failed"}`,
                 body: <GameboardCreatedModal resetBuilder={resetBuilder} gameboardId={gameboardId} error={error}/>,
             }));
-            if (gameboardId) setSubmissionAttempted(false);
+            if (gameboardId) {
+                setSubmissionAttempted(false);
+                void dispatch(saveGameboard({
```
barna-isaac, line 224: why did we need this? The newly created board already appeared on "My question decks", even without it...

src/app/components/pages/Overview.tsx
```diff
@@ -77,7 +77,7 @@ export const TeacherOverview = () => {
                     title: "Tests",
                     icon: {name: "icon-school", color:"secondary"},
                     bodyText: "Set a test curated by the Ada Computer Science team.",
-                    clickUrl: "/set_tests",
+                    clickUrl: "/view_tests",
```
barna-isaac: It's technically possible to set a test from the `/view_tests` page too, but why change this link on Ada? Are we also removing the old `/set_tests` page from Ada?

src/app/components/site/phy/HeaderMenuPhy.tsx
```diff
@@ -26,9 +26,10 @@ export const HeaderMenuPhy = (props: NavProps & {toggleMenu: () => void}) => {
             <div className="my-3"/>
         </>}
         {isOffcanvas && <NavigationSection topLevelLink to="/" title={"Home"}/>}
-        <NavigationSection topLevelLink to="/about" title={deviceSize === "lg" ? "About" : "About Isaac"}/>
+        {(isOffcanvas || above["xl"](deviceSize)) && <NavigationSection topLevelLink to="/about" title={"About Isaac"}/>}
```
barna-isaac, line 29: After the changes in this file, the "About Isaac" menu point is now missing from the top menu for certain screen sizes (eg: width 998px), even when all other menu points are still shown. Before this change, the About Isaac menu appeared and disappeared at the same time as everything else in the top menu, and could be found in the collapsed "Menu" section on smaller sizes. (Although to save space, the caption changed from "About Isaac" to just "About")

src/app/components/site/phy/RoutesPhy.tsx
```diff
@@ -88,17 +85,15 @@ export const RoutesPhy = [
...
-    <Route key={key++} path="/assigned" element={<FeatureFlagWrapper flag={FeatureFlag.MANAGE_ASSIGNMENTS} onUnset={<NotFound />} onSet={
-        <RequireAuth auth={isTutorOrAbove} element={<ManageAssignments />} />} />
-    } />,
+    <Route key={key++} path="/assigned" element={<RequireAuth auth={isTutorOrAbove} element={(authUser) => <ManageAssignments user={authUser} />} />} />,
 
     // Teacher test pages
-    <Route key={key++} path="/set_tests" element={<RequireAuth auth={isTeacherOrAbove} element={(authUser) => <SetQuizzes user={authUser} />} />} />,
-    <Route key={key++} path="/set_quizzes" element={<Navigate to="/set_tests" replace />} />,
+    <Route key={key++} path="/set_tests" element={<Navigate to="/view_tests" replace />} />,
+    <Route key={key++} path="/set_quizzes" element={<Navigate to="/view_tests" replace />} />,
```
barna-isaac, line 92: - redirected `/practice_tests` -> `/view_tests` on Ada, but no redirect for Sci, is this intentional?
- `/set_tests` now just redirects to `/view_tests`, but `/set_assignments`, `/assignment_schedule` are still available through the URL. Why remove one and not the other? Are you planning to remove these later?

src/app/state/slices/api/gameboardApi.ts
```diff
@@ -98,9 +98,9 @@ export const gameboardApi = isaacApi.injectEndpoints({
                 method: "POST",
                 params: {title: newTitle},
             }),
-            invalidatesTags: ["AllGameboards"],
+            invalidatesTags: (_, error, {boardId}) => error ? ["AllGameboards"] : ["AllGameboards", {type: "Gameboard", id: boardId}],
```
barna-isaac, line 101: I don't understand the changes in this file. Before the change, `linkUserToGameboard` didn't change the cache when the request failed, and purged the parts cached under "AllGameboards", as well as the cached entry for the board that was modified. Now, it always purges "AllGameboards", and only purges the specific entry for the board if there's not been an error. Why do you think the error affects AllGameboards, but not the specific board?

barna-isaac (changes requested):
Thanks for making these! I think teachers will be happy we've unified managing tests and assignments in a single page, as well as the ability to see tests on a timeline. I like the new green card banners, and I hope we establish a consistent pattern for using them down the line. I hope this can go out as soon as possible!

With that in mind, I maybe shouldn't have left this many comments. I've marked with ❗️any showstopper bugs that I think are important to fix. I think https://github.com/isaacphysics/isaac-react-app/pull/2228#issuecomment-4867922122 is the most important, as this affects something already on main. For everything else (and indeed, even these): please don't treat any of these as things that _must_ be changed -- the comments are just thoughts I had at the time of reading this, and it's up to you to decide whether they're helpful! For the questions I've asked: if you think the answer is trivial (or just doesn't matter), I'm okay not receiving an answer.

I think we should aim to get this into this week's release, and leave any non-essential fixes to the following weeks.

Two Ada unit tests still fail because of https://github.com/isaacphysics/isaac-react-app/pull/2228#issuecomment-4866062270, and I didn't touch these because I wasn't sure whether you wanted to update the tests or change the site behaviour.

src/app/components/elements/ManageAssignedCards.tsx
jacbn: I appreciate for this specific instance of `TestCard` (and indeed there is only one such instance for now) we could, but the reason for doing this is to more closely mirror `<GameboardCard/>`, in that it may not come from an assignment. If we were to put it in the inner component as you suggest, which represents e.g. a _gameboard_, where would we get the ID for a related _assignment_ from? We would need to pass down the assignment ID from here instead, but that doesn't save us anything compared to this approach. It also bloats `GameboardCard` with a bunch of code that isn't to do with gameboard at all; to me it makes sense to separate all the e.g. assignment logic into the assignment _wrapper_ around the `GameboardCard`.

src/app/components/elements/SaveBoardButton.tsx
jacbn: IIRC this was usually undefined so it didn't really matter, but yes we shouldn't have been using this here.

src/app/components/pages/assignment_progress/AssignmentProgressGroup.tsx
jacbn: I hadn't noticed this specific case, but there are actually a number of places that still link to `/set_assignments` (see usages for `PATHS.SET_ASSIGNMENTS`) and given the time constraints and scope creep on this PR I didn't want to change them all. It's a little annoying because `/assigned` isn't a direct replacement for `/set_assignments`.

What I'll do is adjust this specific link, but make a card to adjust the others. Set Assignments is still a valid link at the moment so if someone does end up there it isn't the end of the world for now.

src/app/components/pages/Gameboard.tsx
jacbn: An anonymous user isn't falsy, it's an object with `{ loggedIn: false }`. I can make it appear by just going to a gameboard in an incognito window?

src/app/components/pages/Gameboard.tsx
jacbn: I suppose so!

src/app/components/pages/Gameboard.tsx
jacbn: Yes, but it's what was there already!

It's the only real action a user can take on that page, so I don't think duplicating it is a bad thing.

src/app/components/pages/GameboardBuilder.tsx
jacbn: It appeared on _My question decks_ in the old world, yes, but we've changed the logic now for _My saved decks_ so that not all boards you create are forcibly and irreversibly linked to your account. We still want them to default to being saved, but we now have to manually call this to do so.

src/app/components/pages/Overview.tsx
jacbn: I replace-all'd this link 🙈. Good spot!

src/app/components/site/phy/HeaderMenuPhy.tsx
jacbn: This is intentional and I confirmed with the content team that this is what they wanted. They really wanted a new `Tests` item in the header and something simply had to go to make space for that at tablet sizes. We decided that the about us link is in the footer anyway, so it wasn't too bad to lose it for this very specific case.

src/app/components/site/phy/RoutesPhy.tsx
jacbn: As mentioned above for Set Assignments, there are still valid links to these pages (and likely more in the content) and replacing them all in the timeframe / scope of this PR isn't really feasible. I'm keeping the pages around at least for now so that links I've missed still work.

src/app/state/slices/api/gameboardApi.ts
jacbn: It **needs** to remove **both** AllGameboards **and** {type: "Gameboard", id: boardId} when linking a user to a board so that we don't cache a board whose saved-to-user status has been modified. If there's an error, nothing happened so we don't really need to invalidate anything, but it was there before so I suppose I kept it. I don't see any harm in always invalidating both if you'd prefer?

(I don't remember adding this, though; was this from a weird merge?)

jacbn:
I can't reply to comments so in future it's more helpful to add them somewhere in the code if there is more than one! But I'll do my best:

> On the /assign page, user-created decks (and possibly all other decks too?) always show in physics purple

This is unfortunately a limitation of the new way we load saved decks; we don't have the full gameboards, only LiteGameboards from the API which don't load the content and thus we can't work out their tags. I've made the default neutral rather than physics, and I would like to change this to be full objects at some point, but perhaps less of an issue now. 

> On the set_tests page, Ada used to have a dedicated layout for xs screens

Yes, fair enough. We are, however, using ALVIs now, which are _much_ more consistent and really ought to be being used everywhere anyway. I'm happy to take a look later at improving the styling, but I do think we should stick with this approach.

> When viewing a question deck, the star icon shows full and sparkly, even when the deck is not saved to "My Saved Decks". Clicking the button asks whether to un-save it, and a message appears about the deck having been removed, even though it's never been saved in the first page. Clicking the star button again still just tries to un-save it.

I can't recreate this at all. Could you show me? Bear in mind some fixes here require backend changes: https://github.com/isaacphysics/isaac-api/pull/805 and it may be that this fixes your issue anyway.

barna-isaac:
@jacbn has confirmed the share, save buttons are intentionally hidden from "My Gameboards", "My assignments" on mobile. Looks good to me!

barna-isaac (approved):

PR #2231, opened 2026-06-26
Remove "Coming soon!" from GCSE Bio physiology

sjd210:


jacbn (commented):
Were we okay with the QF not moving non-disabled options to the top? It might be that this is what content want for consistency with other finders, but to me it doesn't make sense to interleave disabled and non-disabled options; I'd personally add a sortBy isDisabled ascending.

Image

sjd210:
Content haven't indicated anything about order, but I think grouping disabled options together is a good and unintrusive idea so I'll just do that 👍

PR #2232, opened 2026-06-26
Feature/anvil message forward

barna-isaac:
This forwards marking messages from the Anvil app towards the Isaac backend. This completes the work on recording answers! The PR is ready for review now, but please hold off releasing until the backend (https://github.com/isaacphysics/isaac-api/pull/798) as well as config changes (https://github.com/isaacphysics/isaac-sops-config/pull/23) are merged and released. Without the backend changes, we would flood the user's screen with error messages each time they submitted an answer, although just in this one app: https://isaacscience.org/pages/app_page_mental_maths_overall.

To test this locally, apply the config and api changes locally, and visit the http://localhost:8004/pages/app_page_mental_maths_overall page. Watch out, as there is also a migration to apply. Once you've clicked submit, the answers should be recorded to your local database.

### DIFF
--- a/src/app/components/content/AnvilApp.tsx
+++ b/src/app/components/content/AnvilApp.tsx
@@ -1,6 +1,6 @@
 import React, {RefObject, useContext, useEffect} from 'react';
 import {AnvilAppDTO} from "../../../IsaacApiTypes";
-import {selectors, useAppSelector} from "../../state";
+import {selectors, useAppSelector, usePostAnswerMutation} from "../../state";
 import {AccordionSectionContext, QuestionContext} from "../../../IsaacAppTypes";
 import {selectQuestionPart} from "../../services";
 import { AnvilCookieHandler } from '../handlers/InterstitialCookieHandler';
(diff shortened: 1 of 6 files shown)


src/app/components/content/AnvilApp.tsx
```diff
@@ -67,26 +67,29 @@ export const AnvilApp = ({doc}: AnvilAppProps) => {
...
-        const data = e.data;
-
-        if (iframeRef.current && (data.fn == "newAppHeight")) {
-            (iframeRef.current as HTMLIFrameElement).height = data.newHeight + 15;
-        }
-    };
+    const [postSkillsAnswer] = usePostSkillsAnswerMutation();
 
     useEffect(() => {
+        const onMessage = function(e: any) {
```
jacbn: We can certainly narrow the type a little here; if all the available types were known we could put that in place of this `any` too?
```suggestion
        const onMessage = function(e: MessageEvent<any>) {
```

src/app/components/content/AnvilApp.tsx
```diff
@@ -67,26 +67,29 @@ export const AnvilApp = ({doc}: AnvilAppProps) => {
...
+            if (iframeRef.current && e.source !== (iframeRef.current as HTMLIFrameElement).contentWindow) {
+                return;
+            }
+
+            const data = e.data;
+
+            if (iframeRef.current && (data.fn == "newAppHeight")) {
+                (iframeRef.current as HTMLIFrameElement).height = data.newHeight + 15;
+            } else if (iframeRef.current && user?.loggedIn && e.origin === `https://${doc.appId?.toLowerCase()}.anvil.app` && data.type === 'SUBMISSION_MARKED') {
```
jacbn: What is the `e.origin` check for here? Why do we not do it for `newAppHeight`? Perhaps we would rather return immediately at the start if this is not the case?

jacbn (approved):
Two tiny points, but this would be fine to merge.

src/app/components/content/AnvilApp.tsx
barna-isaac: it makes sure the message is indeed coming from our anvil app. On received messages, this is the primary security feature suggested by [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html).

I hadn't included it for the resize checks, since that "just does a layout adjustment" rather than do something critical (like send a post request). Still, for consistency, I'm now doing the same check for that too.

barna-isaac:
@jacbn , could you have another look? I've addressed both issues you've raised. Besides this, @jsharkey13 has suggested a change to the content model, where we now have a new `SkillsApp` type, which includes the `AnvilApp`. This solves the issue you raised about unique id's without us needing to modify the existing anvil apps, and this new content object can also hold other metadata later, like what the subskills are.

jacbn (approved):
Looks good to me!

PR #2233, opened 2026-06-26
Banner cleanup

jacbn:
- Makes a distinction in the file hierarchy between **Banners** (full-width, top of page) and **Alerts** (container width, in-page);
- Refactors existing, useful banners into DismissibleBanner / DismissibleCookieBanner for consistency, for easier creation of new ones
- Deletes unused banners.

My approach here evolved a little inconsistently, so apologies for rewriting much of the banner creation logic in the last commit (04a1dec). I do think this approach is much cleaner than the hooks approach I tried before.


src/app/components/navigation/IsaacApp.tsx
```diff
@@ -93,11 +90,8 @@ const RootLayout = () => {
         <SiteSpecific.Header />
         <Toasts />
         <ActiveModals />
+        <SiteBanners />
         {/* TODO: turn notification banners into a useBanners hook or similar; c.f. REVISION_CHALLENGES – we could reuse the auto-expiry logic */}
```
sjd210: I'm assuming this comment is now unnecessary? We've refactored banners with a useSiteBanners hook, so I'm going to remove it - feel free to add it back if I'm missing something

sjd210 (approved):

PR #2240, opened 2026-07-06
Add accessibility warnings to several new places

sjd210:
Accessibility warnings should now be on:
- Generic pages
- Gameboard builder ALVIs
- TopicSummaryLinks (links ON topic pages, not links TO topic pages)

when the associated page has `access:warning` or `access:motor`. This also allows other content summary tags like `nofilter` or `deprecated` to show, which is also generally useful.

Also does some very minor cleanup to `TopicSummaryLinks.tsx` and `Topic.tsx` by removing all site-specific references since these components are already Ada-specific.

### DIFF
--- a/src/app/components/elements/list-groups/TopicSummaryLinks.tsx
+++ b/src/app/components/elements/list-groups/TopicSummaryLinks.tsx
@@ -5,12 +5,8 @@ import {
     audienceStyle,
     DOCUMENT_TYPE,
     documentTypePathPrefix,
-    isAda,
-    isPhy,
     isIntendedAudience,
     makeIntendedAudienceComparator,
-    notRelevantMessage,
-    siteSpecific,
     stringifyAudience,
     useDeviceSize,
     useUserViewingContext
@@ -19,7 +15,7 @@ import {Link} from "react-router-dom";
 import {selectors, useAppSelector} from "../../../state";
 import classNames from "classnames";
 import {Markup} from "../markup";
-import { ListGroup, ListGroupItem, Button, UncontrolledTooltip } from "reactstrap";
+import { ListGroup, ListGroupItem, Button } from "reactstrap";
 import { Spacer } from "../Spacer";
 
 export function TopicSummaryLinks({items, search}: {items: ContentSummaryDTO[]; search?: string}) {
@@ -29,9 +25,8 @@ export function TopicSummaryLinks({items, search}: {items: ContentSummaryDTO[]; 
 
     return <ListGroup className="mt-4 link-list list-group-links">
         {items
-            // For CS we want relevant sections to appear first
+            // We want relevant sections to appear first
             .sort((itemA, itemB) => {
-                if (!isAda) {return 0;}
                 return makeIntendedAudienceComparator(user, userContext)(itemA, itemB);
             })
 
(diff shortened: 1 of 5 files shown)


src/app/components/elements/list-groups/TopicSummaryLinks.tsx
```diff
@@ -42,37 +38,33 @@ export function TopicSummaryLinks({items, search}: {items: ContentSummaryDTO[];
...
 
             // Render items
-            .map((item, index) => {
+            .map((item) => {
                 const audienceString = stringifyAudience(
                     item.audience, userContext,
                     isIntendedAudience(item.audience, userContext, user)
                 );
 
+                const documentType = (item.type && ["isaacQuestionPage", "isaacFastTrackQuestionPage"].includes(item.type)) ? DOCUMENT_TYPE.QUESTION : DOCUMENT_TYPE.CONCEPT;
```
axlewin: I'm surprised we don't already have a helper method for this check to avoid hardcoding the page type strings, but indeed the only relevant method seems to be `isAQuestionLikeDoc` and I don't think it's worth changing the typing of that just to neaten this up very slightly.

src/app/components/elements/list-groups/TopicSummaryLinks.tsx
```diff
@@ -42,37 +38,33 @@ export function TopicSummaryLinks({items, search}: {items: ContentSummaryDTO[];
...
-                                </UncontrolledTooltip>
-                            </div>}
+                        <div className="title p-3 ps-6 d-block d-sm-flex">
+                            <Markup encoding={"latex"}>
+                                {item.title}
+                            </Markup>
+                            <ContentPropertyTags 
+                                className="ps-sm-2"
+                                deprecated={item.deprecated}
+                                supersededByPath={item.supersededBy ? `/questions/${item.supersededBy}` : undefined}
```
axlewin: Is `/questions` a safe assumption?

In practice there are currently no superseded concepts on Ada so it makes no difference here, but there's at least one on Isaac which suggests superseding concepts is supported. We could have a `documentType` check here just to be safe?

axlewin (approved):
Small question about a `supersededBy` path, but once you either address that or decide it's not actually a problem this looks good to merge 👍

src/app/components/elements/list-groups/TopicSummaryLinks.tsx
sjd210: Huh, there are exactly two on Isaac but they don't fully behave as expected anyway. They do get blocked from being filtered by non-staff, and an ALVI will include the pill - but we don't have anything appearing on the concept page itself 🤔 

I think generally concept pages can just be edited directly, whereas editing question question pages can break progress - so there's almost no need for superseding for concepts. Anyway for this case, it's simple enough to re-use the `documentType` and think no further, so I'll just do that.

src/app/components/elements/list-groups/TopicSummaryLinks.tsx
sjd210: I wasn't really happy with how I did this either. It was because `documentTypePathPrefix` was expecting a `DOCUMENT_TYPE`, but it's much smoother to just cast the `string` to it instead instead. An item without a proper type isn't able to get past previous checks in the code, so it's safe enough to do.

PR #2243, opened 2026-07-07
Gameboard builder visual enhancements

sjd210:
Several small things:

- Allow subtitles to appear below title on `flatLayout` ALVIs (for the deck builder)
- Allow question ID to be checked using the deck builder preview button tooltip
    - I tried a few different ways for this. Including the ID directly or a new hover region was messy-looking, and adding a tooltip to the title was too distracting. This seemed best, and definitely works for the desired use-case
    - (I also changed the table builder row title tooltip to include it, so the tooltips still match)
- Use a css `icon` for drag-indicator so that it becomes white and hence visible in dark mode
- Since the page uses a unique ALVI type, expand cypress test so these ALVIs are visible (and with a slightly larger width to accommodate all the main builder ALVI characteristics)

(PR recreated from https://github.com/isaacphysics/isaac-react-app/pull/2242 because rebasing got messy 🫣)


jacbn (approved):
This is great and we should get this in, but the point I was trying to make with content still stands here – there really isn't much point in a `flatLayout` if we're going back on the decision to make everything as thin as possible. Including the subtitle under the title practically doubles the height of rows that didn't have that before, so the original arguments like "it's too tall to drag things around" that made us change this in the first place will only come back. 

But, if it's what they want now 😮‍💨. There is no noticeable difference in height if we remove `style="flat"` from the builder's list view, so having it (and all the `flatLayouts` in ALVI) is quite pointless now. We _could_ clean everything up, but maybe it's worth keeping for a little while in case they have another change of heart.

---

I can't remember what the use case was for the IDs, and whether it was just checking or copying – but if we did need to copy to, I do think _clicking_ is a distinct action from _dragging_ so we could e.g. copy the ID to the clipboard if you click the title, or similar. We could also e.g. only enable this for staff, as I doubt teachers will use it much.

sjd210:
@jacbn I didn't see/hear your original point with content, but I both agree that subtitles are definitely required enough that they should be shown on the builder ALVIs, and also that a "flat" layout that's not really flat is counterintuitive - however I think that this was already an issue with the stacked stage/difficulty list, especially for questions with 3+ difficulties, so it's not anything altogether new.

Since we just had a bunch of requests come through after the original builder rework, I'd rather wait before simplifying the code again to make sure we're definitely settled on a design.

---

The request for IDs was indeed simply for checking: _**"I usually do a check that I've added the right questions in the right order by mousing over the questions I've added to the deck and comparing the ids to my list of ids"**_. 

I don't _expect_ there's a need to copy IDs, but equally I wouldn't be surprised if one came up. In terms of implementation, I think clicking and dragging aren't quite the *same*, but for an action that you can't easily undo like overwriting the clipboard I'd rather they weren't so similar. Maybe that's a situation for a custom context menu? Where we could also have options for e.g. opening the question in a new tab, or checking question properties that aren't otherwise shown on this ALVI (status, field, etc.)? We've not tried anything of the sort before, but maybe its worth a go.

sjd210:
Also FYI, I'm merging this but added one more small thing that I noticed while looking back. Topics are appearing by themselves on builder ALVIs, and we should be using aliases for topics that do this since their regular title may be too vague e.g. Power -> Electrical Power. This also affects the table in the builder modal.

PR #2248, opened 2026-07-09
Feature/skills on my progress

barna-isaac:
Starts showing skills attempts on my progress. I'll deploy this, as well as the necessary backend changes to dev, so you can try this.

Key considerations, for now:
- just shows results from the mental maths app, which is hardcoded
- only loads skills attempts when the skills tab has been selected, and each time the tabs are toggled, it fetches again. There's no spinner yet.
- there's an empty state for when there haven't been any attempts. It tells users about the maths practice app.
- should look nice on mobile as well as desktop!

### DIFF
--- a/src/IsaacApiTypes.tsx
+++ b/src/IsaacApiTypes.tsx
@@ -410,6 +410,7 @@ export interface AnvilAppDTO extends ContentDTO {
 export interface SkillsAppDTO extends ContentDTO {
     anvilApp?: AnvilAppDTO
 }
+
 export interface ChemicalFormulaDTO extends ChoiceDTO {
     mhchemExpression?: string;
 }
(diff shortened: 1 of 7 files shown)


jacbn (commented):
For the sake of time I've directly made a number of edits to this, rather than raising them all as comments – but please still double-check over them. The big ones are:
- Refactor `ActivityGraph` to fix an ESLint hooks warning around `selectedDates` causing too many rerenders, and to regenerate the graph on window resize;
- Rework tabs to more directly allow the pattern of "show all tab headers, but only render each tab when you're _on the tab_; normally, the `<Tabs/>` component renders all of them (and therefore runs all the hooks / network requests) but `d-none`'s them, but now with `renderHiddenTabs={false}` you can avoid doing this _while using the expected tab pattern_.
- Tidy up the DOM structure by making better use of ReactStrap's `Row` / `Col` classes.

There are a few other smaller changes too, but those are mostly stylistic. I think I'm happy for this to go in now, but as I say please double check!

PR #2249, opened 2026-07-10
Hide unnecessary sidebar element on Ada quiz feedback

axlewin:
We need this for indenting the attempt footer properly on sci (since the rest of the page has a sidebar) but it isn't needed on Ada and just shows up as a white box.


jacbn (approved):
LGTM!

PR #2250, opened 2026-07-13
Set metadata on a quick question page as "Correct"

sjd210:
All question pages should have markable questions on them, so this is not really valid content - but they display as "Correct" from a GameboardItem in gameboards and their sidebars (or any other API sources like the QF), so we ought to be visually consistent with that and change how its calculated for question metadata.


src/app/state/selectors.tsx
```diff
@@ -27,7 +27,8 @@ export const selectors = {
     questions: {
         getQuestions: (state: AppState) => state?.questions?.questions,
         allQuestionsCorrect: (state: AppState) => {
-            return !!state && !!state.questions && state.questions.questions.every(q => !!q.bestAttempt?.correct);
+            // Invalid question pages without any markable questions (e.g. just a quick question) should still have a "correct" status for consistency
+            return !!state && (!state.questions || state.questions.questions.every(q => !!q.bestAttempt?.correct));
```
jacbn, line 31: I think the all-correct behaviour supersedes them so this wouldn't change anything visually at the moment, but surely this logic should apply to `anyQuestionCorrect` and `allQuestionsAttempted`, if for nothing else than the same consistency we are trying to achieve with the changes to `allQuestionsCorrect`?

jacbn (commented):
Working as expected, just have a small point about code consistency we could consider?

src/app/state/selectors.tsx
sjd210: I did consider this, so if you had the same thought then that's enough consensus that I'll do it 👍

jacbn (approved):

PR #2252, opened 2026-07-14
Show error on test save failure

jacbn:
Show a toast when the attempt fails to submit. Does **not** enable any option to reattempt, but rather asks the user to reload the page (all *prior* attempts should be saved, so the effort of building a new feature to achieve the same thing feels excessive), with some help text to remember to save their answers elsewhere.


axlewin (approved):
LGTM!

PR #2255, opened 2026-07-16
Improved /account page blocking

jacbn:
We have several issues in the past regarding the account page preventing saving owing to a field being invalid in a different tab.

For example, you could:
- Reset the first name to blank (invalid)
- Switch to a different tab and attempt to make an account change
- This will not work.

Admittedly, an option becoming invalid is unlikely, but accidental changes are not – it wouldn't be obvious that e.g. a name change has occurred if you weren't on the tab.

As a solution to these, we would instead prefer that there be a warning if you attempt to switch tabs with unsaved changes. If you proceed regardless, the tab states should all reset, so that if you make adjustments to another tab, the first set of changes cannot block the changes.

### DIFF
--- a/src/app/components/pages/MyAccount.tsx
+++ b/src/app/components/pages/MyAccount.tsx
@@ -182,6 +182,19 @@ export const MyAccount = ({user}: AccountPageProps) => {
 
     const formSpecificTabs = [ACCOUNT_TAB.passwordreset, ACCOUNT_TAB.teacherconnections];
 
+    const [clearUnsavedChanges, setShouldClearUnsavedChanges] = useState(false);
+    const resetState = useCallback(() => {
+        setUserToUpdate({...(editingOtherUser ? userToEdit : user), password: ""});
+        setMyUserPreferences(userPreferences);
+        setEmailPreferences(userPreferences?.EMAIL_PREFERENCE);
+        setUserContextsToUpdate(userToUpdate.registeredContexts?.length ? [...userToUpdate.registeredContexts] : [{}]);
+        setAttemptedAccountUpdate(false);
+        setSaving(false);
+        setCurrentPassword("");
+        setNewPassword("");
+        setShouldClearUnsavedChanges(true);
+    }, [editingOtherUser, userToEdit, user, userPreferences, setEmailPreferences, userToUpdate.registeredContexts]);
+
     useEffect(() => {
         setEmailPreferences(userPreferences?.EMAIL_PREFERENCE);
         setMyUserPreferences(userPreferences);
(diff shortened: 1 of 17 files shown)


src/app/components/pages/MyAccount.tsx
```diff
@@ -305,26 +336,20 @@ export const MyAccount = ({user}: AccountPageProps) => {
...
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [activeTab]);
-
     return <PageContainer id="account-page" className="mb-7"
         pageTitle={
             <TitleAndBreadcrumb currentPageTitle={pageTitle} icon={{type: "icon", icon: "icon-account"}} className="mb-3"/>
         }
         sidebar={siteSpecific(
-            <MyAccountSidebar editingOtherUser={editingOtherUser} activeTab={activeTab} setActiveTab={setActiveTab}/>,
+            <MyAccountSidebar editingOtherUser={editingOtherUser} activeTab={activeTab} setActiveTab={safelyChangeTab}/>,
```
sjd210, line 352: This `safelyChangeTabs` is firing twice, so whenever the confirm dialog is cancelled - it immediately comes straight back up.

It turns out that this is kinda a pre-existing issue with `StyledTabPicker` (although one that doesn't matter for most instances, since duplicating a `navigate` or `setState` makes no difference).
https://github.com/isaacphysics/isaac-react-app/blob/4a5d2e7d8a38c5eaef5feda88f67672d6e2a8b62/src/app/components/elements/inputs/StyledTabPicker.tsx#L72-L75
When we use an `onClick`/`onKeyboardDown` system for `StyledTabPicker` like `MyAccountSidebar` does, the `onClick` gets set to the above `Label` AND activates the `Input` within so sends two events. 

Some `StyledTabPicker` instances use `onInputChange` directly, which avoids this - so we should probably either change all instances to that site-wide OR we can protect against the issue within `StyledTabPicker` by passing `onClick` as a prop directly to the `Input` and not to the `Label` (as is done via `{...rest}`). Either works 🤷‍♀️

sjd210 (changes requested):
Also, an interesting quirk that I don't think matters: Changing data in one tab and then navigating away to another tab with a hash anchor will still cause the "You have unsaved changes on this page" dialog on trying to click away, despite the user now being on a different page without that data.
We don't surface the hash anchors on the page, so it would have to be done intentionally, and we still fail safely - so if a user puts themself in that position, so be it.

src/app/components/pages/MyAccount.tsx
jacbn: I think this is almost right – but the input activating isn't doing anything as it doesn't have an event handler. The `Label` element is, however, repeating the `onClick` method because of the input; [this post](https://stackoverflow.com/a/76931227) describes how onClick is hitting both the label and the input, but the input's one is getting propagated up to the label so that one fires twice. Your fix works because `onInputChange` is explicitly not applied to the `Label` but is to the `Input`.

To prove this, a third alternative solution (not suggesting we use this, just as proof of concept) is to stop the propagation of the onClick and onChange events inside the input:
```tsx
<Input ... onClick={(e) => e.stopPropagation()} onChange={(e) => {e.stopPropagation(); onInputChange?.(e);}}... />
```

I don't really know if we should be running event handlers in the input or the label – as the post describes, events that hit the label are also sent to the input if they are tied, but if they are nested events will also propagate from the input to the label, so it seems arbitrary which we choose. Perhaps the input is the more logical place for them; they shouldn't fire twice if existing there, I think.

jacbn: I've made a cleanup branch targeting this one [here](https://github.com/isaacphysics/isaac-react-app/pull/2259); it's quite a large set of changes so figured it would be best reviewed separately? Do let me know your thoughts.

sjd210: That Stack Overflow post is also how I found out what the issue was. I kinda just... didn't fully explain it while reviewing, just giving the gist (although I certainly could've linked it too) 🙃 My bad

Reviewing the cleanup branch now

sjd210 (approved):

PR #2257, opened 2026-07-20
Visual revisions for accessible reorder/parsons questions

sjd210:
Most notably, this PR adds a new `useSingleList` property to reorder/parsons questions to enable a mode where there is only a "Your answer" list that all items start in, and which must be sorted in place (rather than moving items from "Available items" first). This can only be used for questions where all items are required in the answer (unlike e.g. [combined_transformations_r1](https://isaacscience.org/questions/combined_transformations_r1)), and any previous attempts at the question will be displayed only if they already included all items. It should otherwise be fully backwards-compatible.

Other smaller changes:
- Adds a reorder question VRT alongside the existing parsons one. This question uses the `useSingleList` property.
  - The regression test page will also be updated once both a content cycle has occurred to let the property be used universally, and the playwright tests have been updated to allow both question types to work.
- Removes `expansion-layout` from Isaac for these questions, and for Ada only when `useSingleList` is on
- Make list swap button always visible on two-list layout (Parsons indent buttons are still on-hover to preserve space for small items)
- Remove `<pre>` from Reorder questions to reinstate correct font and wrapping behaviour
- Disable reorder buttons in "Available items" column

Also see https://github.com/isaacphysics/isaac-content-editor/pull/128 and https://github.com/isaacphysics/isaac-api/pull/812


src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -162,23 +170,23 @@ const IsaacParsonsQuestion = ({doc, questionId, readonly} : IsaacQuestionProps<I
         </div>
         <Row className="my-md-3">
             <DragDropContext onDragEnd={onDragEnd} onDragStart={onDragStart} onDragUpdate={onDragUpdate}>
-                <Col md={6} className="parsons-available-items">
+                <Label className="visually-hidden" id="item-section-info">
```
jacbn, line 173: One thing that stood out to me immediately on using the single list was how it really wasn't clear to a _sighted_ user how to interact with the question. The alt-text is fine and very clear, but for a sighted user, the only visible text is **Your answer**; there's perhaps a little over-reliance on knowing that you have to reorder the items and that that's what the icons represent. Remember that we will have 11yos answering these!

Image

---

Looking at other types of question for some inspiration, we tend to use instructive placeholders (i.e.. "click here to...":
Image

Some questions pair these placeholders with explanatory text:
Image

...and if no direct input exists, e.g. in the sketcher, we still use explanatory text:
Image

The original dual-column reorder questions get away with this a little better because the empty right-hand column contains such a placeholder:

Image

I don't love that this description disappears if you start answering the question*, but it's not as bad as the single-column case which has no instructions at all. 

> *I understand that the placeholder disappears for other question types that use placeholders too, but for those there is only a single input box – and if that box is filled in, it is very clear that you have previously attempted this question. What we have for reorder questions is very custom, so the "obviousness" of previously having attempted the question is not there. To that end, making the instructions visible in all cases seems like it can't be a bad thing.

Given that there isn't really anywhere to _put_ a placeholder in the single-column case, I wonder if taking the chemistry / sketcher question approach of having some explanatory text is the way to go. Even if it's just a single sentence along the lines of "Drag these items to put them in the correct order" in small font.

As an aside, I wonder if the "Your answer" text even makes sense at all in the single-column case. There's nothing else to compare it to, so of course it's my answer?

src/app/components/elements/ParsonsDraggableItem.tsx
```diff
@@ -269,27 +277,23 @@ export const ParsonsDraggableItem = ({currentItem, index, items, setItems, inAva
...
                 aria-describedby={undefined}
             >
-                <ReorderButtons index={index} items={items} setItems={setItems} isParsons={isParsons} currentIndent={currentItem.indentation}/>
-                <pre>
-                    <Markup trusted-markup-encoding={"html"}>
-                        {currentItem.value}
-                    </Markup>
-                </pre>
+                <ReorderButtons index={index} items={items} setItems={setItems} isParsons={isParsons} currentIndent={currentItem.indentation} inAvailableItems={inAvailableItems} />
+                {isParsons && canIndent ? <pre className="item-text">{markupItem}</pre> : <div className="item-text">{markupItem}</div>}
```
jacbn: My understanding is that **all** Parsons questions are to do with code snippets, regardless of whether indenting is enabled or not. 

See https://adacomputerscience.org/questions/fileprog_17; since `disableIndentation` is set for this question, the `canIndent` field on this line is false and so this question is no longer using code blocks.

This change would make this the only (I think?) visual difference between a reorder question and a Parsons question with disabled indentation, which ought to have _something_ differentiating them – as otherwise why do we have two question types?

```suggestion
                {isParsons ? <pre className="item-text">{markupItem}</pre> : <div className="item-text">{markupItem}</div>}
```

src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -187,19 +195,19 @@ const IsaacParsonsQuestion = ({doc, questionId, readonly} : IsaacQuestionProps<I
...
                     <h4 className="mt-4 mt-md-0">Your answer</h4>
                     <Droppable droppableId="answerItems">
                         {(provided: DroppableProvided) => {
-                            return <div id="parsons-choice-area" ref={provided.innerRef} className={classNames("parsons-items", {[`ghost-indent-${currentIndent}`]: isDefined(draggedElement) && currentIndent !== null, "empty": !(currentAttempt && currentAttempt.items && currentAttempt.items.length > 0), "is-dragging": draggedElement})}>
-                                {currentAttempt && currentAttempt.items && currentAttempt.items.map((item, index) => 
-                                    <ParsonsDraggableItem key={item.id} currentItem={item} index={index} readonly={readonly}
-                                        items={attemptItems} setItems={setAttemptItems} canIndent={canIndent} isParsons
+                            return <div id={`${questionId}-parsons-choice-area`} ref={provided.innerRef} className={classNames("parsons-items", {[`ghost-indent-${currentIndent}`]: isDefined(draggedElement) && currentIndent !== null, "empty": !(attemptItems.length > 0), "is-dragging": draggedElement})}>
```
jacbn:
```suggestion
                            return <div id={`${questionId}-parsons-choice-area`} ref={provided.innerRef} className={classNames("parsons-items", {[`ghost-indent-${currentIndent}`]: isDefined(draggedElement) && canIndent && currentIndent !== null, "empty": !(attemptItems.length > 0), "is-dragging": draggedElement})}>
```

This makes non-indentable Parsons qs not show the indent border, which I think tidies them up a little.

src/scss/common/questions.scss
```diff
@@ -450,9 +450,9 @@ figure .inline-container {
       // Use negative z-index rather than `display: none` to hide buttons so they can still be selected using keyboard controls
       .hidden-buttons {
         position: absolute;
-        right: 2px;
+        right: 38px;
```
jacbn: The buttons here feel oddly positioned to me. I'd expect the distance to the top and bottom to roughly be the same as the distance to the right? Maybe half what it currently is?

Image

---

Also, being absolute means they overlap long lines and do have a funny interaction where you can't read the contents while hovering.

Image

I think though that this is necessary to avoid reflowing of content in all situations. I can't find a universal way around this, so perhaps we should just convince the content team not to make lines be too long.

jacbn (commented):
A few small points, but this is a nice improvement!

src/scss/common/questions.scss
sjd210: This 38px is meant to allow space for the list-swapping buttons when they exist, but I didn't consider the case of this new `useSingleList` world where they don't. In this case, yes I absolutely agree that they shouldn't be that far left.

src/app/components/content/IsaacParsonsQuestion.tsx
sjd210: I absolutely agree. The texts for the symbolic/graph sketcher questions don't _fully_ explain their interaction mechanisms, but they're complicated enough to need help modals within so that's a different issue with a different solution (and for which we still need to think about the user-specific storage solution for). 

Just a short sentence should suffice for this simpler question type.

src/app/components/elements/ParsonsDraggableItem.tsx
sjd210: I can agree that this is our _intention_ for Parsons questions, but we still have several (mostly older) questions like [bb_fusion_tokamak_q1](https://staging.isaacscience.org/questions/bb_fusion_tokamak_q1) or [gcse_maths_ch8_46_q1](https://staging.isaacscience.org/questions/gcse_maths_ch8_46_q1) that are using Parsons with `disableIndentation` despite not being code questions.

I think the issue here is that we originally just had Parsons questions, then Reorder questions were added later (2019 vs 2022 looking back at the Git history). Ada seems to have superseded their old Parsons questions that were non-code, but that hasn't happened on Isaac. Instead of what I've done here, we probably should enforce the code styling for Parsons questions, then notify the team that these questions have a content issue (and they can decide if it's a big enough deal to supersede them).

jacbn (approved):
LGTM!

PR #2259, opened 2026-07-20
StyledTabPicker onChange cleanup

jacbn:
See https://github.com/isaacphysics/isaac-react-app/pull/2255#discussion_r3615576153; this makes all usages of `StyledTabPicker` consistent such that they all use `onChange` when the tab is changed.

### DIFF
--- a/src/app/components/pages/Support.tsx
+++ b/src/app/components/pages/Support.tsx
@@ -148,7 +148,7 @@ export const Support = () => {
                 {Object.values(section.categories).map((category, index) => 
                     <StyledTabPicker
                         key={index} id={category.category} tabIndex={0} checkboxTitle={category.title} checked={categoryIndex === index}
-                        onClick={() => activeTabChanged(index)} onKeyDown={ifKeyIsEnter(() => activeTabChanged(index))}
+                        onChange={() => activeTabChanged(index)} onKeyDown={ifKeyIsEnter(() => activeTabChanged(index))}
                     />
                 )}
             </FAQSidebar>,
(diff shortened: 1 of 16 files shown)


src/app/components/pages/Support.tsx
```diff
@@ -148,7 +148,7 @@ export const Support = () => {
                 {Object.values(section.categories).map((category, index) => 
                     <StyledTabPicker
                         key={index} id={category.category} tabIndex={0} checkboxTitle={category.title} checked={categoryIndex === index}
-                        onClick={() => activeTabChanged(index)} onKeyDown={ifKeyIsEnter(() => activeTabChanged(index))}
+                        onChange={() => activeTabChanged(index)} onKeyDown={ifKeyIsEnter(() => activeTabChanged(index))}
```
sjd210:
```suggestion
                        onChange={() => activeTabChanged(index)}
```

sjd210: Should we also set `onKeyDown?: never` on the props? `onChange` already handles keyboard interactions, and I can't think of a situation where we'd want it to behave any differently with a keyboard than a click.

sjd210 (changes requested):
One small thing, but then I'm happy for both this and the base branch to be merged 👍

sjd210:
Also we're not getting a check for it since this isn't merging into main, but see https://github.com/isaacphysics/isaac-react-app/pull/2260 which shows that the sidebar isn't being closed properly at small screens

src/app/components/pages/Support.tsx
jacbn: Hmm. `onKeyDown` has a broader theoretical scope of actions tied to it than click,  so forcibly removing it does technically limit options. But I agree that in the case where all we're running is `ifKeyIsEnter` and then the same function as onChange, it's not worth having... and this case is _realistically_ all you'd ever use, so 🤷. You could argue either way. I'll exclude it in the type because I don't think we'll ever need it for anything else.

jacbn:
The VRT shouldn't be valid anymore (hopefully).

sjd210:
`onChange` being at least partially enforced seems like an improvement, and certainly should be fine if devs are following the pattern of other `StyledTabPicker` implementations. 👍

sjd210 (approved):

PR #2262, opened 2026-07-23
Signup overhaul: 11-12, SSO

jacbn:
- Allows 11-12 users to log in to Ada via SSO.
- Reworks the SSO flow to direct back into the email signup flow, such that required account info is not left to the RAI modal after signing up. 
- As a consequence, allows teachers to sign up directly on Ada with SSO, like they can for email
- Adjusts Ada user info collection to:
  - include required DoB, MM/YY only
  - require gender input
  - remove defaults for (non-assignment) email prefs; require users to pick values for news + events options
- Shifts user preferences before "Account created!" message (Ada)
- Adds "Account created!" message for Sci with graphic, to align the two flows more closely


src/app/components/pages/RegistrationAgeCheck.tsx
```diff
@@ -50,6 +53,8 @@ export const RegistrationAgeCheck = () => {
             () => navigate("/register"));
     };
 
+    const AGE_LOWER_LIMIT = siteSpecific(10, 11);
```
axlewin: Could we use this constant in `RegistrationAgeCheckFailed`, too? On Ada, selecting '10 or under' still shows 'we aren't able to offer accounts to students under 13 years old'.

src/app/components/elements/sidebar/SignupSidebar.tsx
```diff
@@ -4,21 +4,30 @@ import { confirmThen } from "../../../services";
...
+        }
     };
 
     return <ContentSidebar buttonTitle="Create an account">
         <div className="section-divider mt-4"/>
         <h5 className="mt-1">Create an account</h5>
         {/* Tabs are clickable iff their page could be reached with a Back button */}
         <StyledTabPicker checkboxTitle={"Sign-up method"} checked={activeTab === 0} disabled={activeTab > 2} onChange={() => (activeTab === 1 || activeTab === 2) && goBack("/register")}/>
-        <StyledTabPicker checkboxTitle={"Age verification"} checked={activeTab === 1} disabled={activeTab < 1 || activeTab > 2} onChange={() => activeTab === 2 && goBack("/age")}/>
+        <StyledTabPicker checkboxTitle={"Age verification"} checked={activeTab === 1} disabled={activeTab < 1 || activeTab > 2} onChange={() => activeTab === 2 && handleSSOBack("/age")}/>
```
axlewin: This might be something I broke in the merge with the other tab picker changes (apologies if so), but: on Isaac, selecting 13+ then navigating back with this sidebar tab doesn't work (`page not found: /age`).

src/app/components/pages/RegistrationAgeCheckSSOOnly.tsx
```diff
@@ -0,0 +1,62 @@
...
+                    <GoogleSignInButton knownRole="STUDENT" />
+                    {isPhy && <MicrosoftSignInButton knownRole="STUDENT" />}
+                    {isPhy && <SsoHelpLink />}
+                </div>
+                <Row className="justify-content-end">
+                    <Col sm={4} lg={3} className="d-flex justify-content-end">
+                        <Button className="mt-2 w-100" color="keyline" onClick={() => navigate(-1)}>Back</Button>
+                    </Col>
+                    <Col sm={6} lg={4}>
+                        <Button className="mt-2 w-100" color="solid" onClick={returnToHomepage}>{siteSpecific("Take me back", "Go home")}</Button>
```
axlewin, line 56: Incredibly minor: "Go home" matches neither the designs ("Try another way") nor the equivalent button for <10s ("Back to the site"), is this intentional?

src/app/components/pages/RegistrationAgeCheck.tsx
```diff
@@ -101,7 +103,7 @@ export const RegistrationAgeCheck = () => {
                             color="primary"
                             invalid={submissionAttempted && agePermission === undefined}
                         />
-                        <Label for="registration-age-check-under" className="ms-2 mb-0">Under {siteSpecific("10 years old", "13")}</Label>
+                        <Label for="registration-age-check-under" className="ms-2 mb-0">{AGE_LOWER_LIMIT - 1} or under</Label>
                         <FormFeedback>
                             Please make a selection.
```
axlewin, line 110: This makes the last bullet point wrap:

Image

src/app/components/elements/sidebar/SignupSidebar.tsx
axlewin: Ah, and similarly: clicking "Back" on the email signup details page  goes to `/register/student/details/age` which doesn't exist.

src/app/components/pages/MyAccount.tsx
```diff
@@ -315,7 +315,7 @@ export const MyAccount = ({user}: AccountPageProps) => {
         if (userToUpdate.loggedIn &&
             validateEmail(userToUpdate.email) &&
             allRequiredInformationIsPresent(userToUpdate, {...newPreferences, EMAIL_PREFERENCE: null}, userContextsToUpdate) &&
-            (isDobOldEnoughForSite(userToUpdate.dateOfBirth) || !isDefined(userToUpdate.dateOfBirth)) &&
+            (isDobOldEnoughForSite(userToUpdate.dateOfBirth) || (isPhy && !isDefined(userToUpdate.dateOfBirth))) &&
```
axlewin, line 318: This might have already been discussed, but how will this work for existing users that don't have a DOB set? Is it confusing if they just see "Please enter a valid date of birth" next time they update their details? Do we want a RAI-style modal like we did when country became compulsory on sci?

src/app/services/validation.ts
```diff
@@ -40,9 +40,9 @@ const isDobOverN = (n: number, dateOfBirth?: Date | number) => {
     }
 };
 
-export const isDobOverThirteen = (dateOfBirth?: Date | number) => isDobOverN(13, dateOfBirth);
-export const isDobOverTen = (dateOfBirth?: Date | number) => isDobOverN(10, dateOfBirth);
-export const isDobOldEnoughForSite = siteSpecific(isDobOverTen, isDobOverThirteen);
+export const isDobOldEnoughForSite = (dateOfBirth?: Date | number) => isDobOverN(10, dateOfBirth);
```
axlewin: Should this be 11 for Ada?

src/app/components/elements/inputs/DobInput.tsx
axlewin: Should this be 11 for Ada now, to allow SSO users to set their DOB?

src/app/components/pages/MyAccount.tsx
jacbn: I've added DoB to `validateRequiredFields` in `validate.ts`, which is used by the RAI to generate the required fields to show in that modal. So, a missing date of birth will indeed appear in the RAI, yep!

image

src/app/components/pages/RegistrationAgeCheckSSOOnly.tsx
jacbn: When I first made the designs it took you back to the start of the flow, so "Try another way" as an action made sense as it would let you try SSO after clicking. We since moved SSO options directly onto this page, so "Try another way" makes less sense now – hence why I changed it to something new. "Back to the site" duplicates the word "back" from before which doesn't help with clarity and is also less descriptive (where does it take you?). I think of all the choices I prefer "Go home", and I'd be happy to switch the under 10 flow to the same text for consistency.

axlewin (approved):

PR #2269, opened 2026-07-27
Help users find quiz footers

barna-isaac:
For the problem, see the ticket.

This lifts the position of quiz footers. I think our previous layout looked a little more elegant on large screens, but the new layout makes footers visible on shorter screens. I've discussed the solution with @jacbn and they agreed this should be the fix.

Here's what it looked before
Screenshot 2026-07-27 at 12 18 16

And here's what footers look with the changes in this PR:
Screenshot 2026-07-27 at 12 18 41

The previous layout was achieved by having two vertically stacked sidebar layouts. With the changes here, there's just one, and the footer is included in the main content.


jacbn (approved):
This is an improvement on the current quiz layout and does fix the required issue, so I'm happy for this to go in. 

However, I'm going to make a card to improve the quiz sidebar layout a little more widely in the general case to make it more consistent with the other pages on the site, as I think there is more that could be done. For example, I'm not keen on the footer existing as a child of `QuizContentsComponent`; this doesn't structurally make much sense, you would expect the footer to exist _after_ but _adjacent to_ the content, both inside a sidebar layout parent. The problem with this for quizzes is that the `QuizContentsComponent` _is_ the sidebar layout, which is the real cause of the inconsistency here.

In most other places, we use a PageContainer that wraps around everything, that takes a pageTitle and a sidebar prop. The main content then exists as a child to this container. I'd prefer quizzes follow this structure too, although as I say this is a larger task and I don't expect this to be done here.

PR #2270, opened 2026-07-28
Cleanup IsaacHints components

sjd210:
Outwardly, this allows hint titles to be displayed instead of "Hint 1", "Hint 2", etc. for Ada (and for print view), and switches its modal to an ActiveModal (which gives its the standard header with a bigger title and a 'X' button to close)

Code-wise, as-is `IsaacLinkHints` and `IsaacTabbedHints` would then both share logic for incrementing undefined titles, and recording hint views. Instead of extracting those functions and needing to deal with their props, we can instead combine the components into a new component `IsaacHints` with a `style` prop to control if displaying "modal" or "tabbed", which should generally simplify the hint code.


axlewin (approved):
Looks good; it's nice to have this code tidied up anyway (and to finally get rid of the extra blank space that was present for questions without hints!).

sjd210:
PR was already approved, but I noticed while I had the branch open that Sci quiz question padding looked baaad so I snuck in an extra quick fix. It's almost never used (the only published, non-hidden from teacher one I can find is eqn_practice_springs_a), but we should support it anyway.

PR #2272, opened 2026-07-29
Stop using subjectId as a secondary source of context

sjd210:
~~Pages are augmented with a `subjectId` when their tags are processed, however previously in cases of multiple subject tags (and for which there was no subject in the url), the ID would be whichever is the first in the `doc`'s tag list. The order of this tag list should not matter (and indeed does not appear to be preserved when indexed) - instead we should use the subject priority that is already used for `pageContext`: **[physics -> maths -> chemistry -> biology]**~~

~~This was leading to mismatched page and title icon themes, although this could have also been avoided with a consistent order of `pageContext?.subject ?? doc.subjectId` (rather than the other way around), or by not re-stating the subject unnecessarily in the `TitleAndBreadcrumb`. To help avoid anything similar coming up again, these have both also been addressed.~~

(Kept for posterity, but no longer relevant)

---

An old system for determining the subject that should be associated with a page was by calculating it when processing tags, then augmenting the doc with a `subjectId` corresponding to that calculated subject. This system was not at all sufficient for the redesigned Sci site with the `pageContext` system mostly replacing it, but we never fully got rid of this old system.

The specific issue that this is trying to solve is a mismatch between page theme and title hex theme. For question pages, the page theme had been determined by `pageContext`, whereas the hex theme was determined by `subjectId` - however `subjectId` did not have the subject prioritisation of `pageContext` for multi-subject pages, leading to this mismatch. Rather than make `subjectId` work properly and maintain both systems, I replaced any remaining uses of `subjectId` and removed all associated code to clean up the codebase. 

(+ allows subject colouring to work properly with the `icon-info` used on generic page ALVIs in Sci search)

### DIFF
--- a/src/IsaacAppTypes.tsx
+++ b/src/IsaacAppTypes.tsx
@@ -724,10 +724,6 @@ export interface Tag extends BaseTag {
     level: number;
 }
 
-export interface DocumentSubject {
-    subjectId?: string;
-}
-
 export interface Choice extends ChoiceDTO {
     correct?: boolean;
     explanation?: ContentBase;
--- a/src/app/components/elements/QuestionMetadata.tsx
+++ b/src/app/components/elements/QuestionMetadata.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { Col } from "reactstrap";
 import { IsaacQuestionPageDTO } from "../../../IsaacApiTypes";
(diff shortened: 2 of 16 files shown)


axlewin (approved):
This all looks sensible; it certainly fixes the original issue, and it should make it easier to see how subject colours are actually working now 👍

PR #2276, opened 2026-07-31
Make gender dropdown keyboard-accessible in Firefox

axlewin:
Having a `hidden` option prevents this dropdown from working properly in Firefox (it works fine with enter followed by arrow keys to scroll options, but if you press the spacebar to open the dropdown then it's impossible to select an option). For accessibility, this dropdown should work the same way as the others on this page.


barna-isaac (approved):
I've not been able to reproduce the bug on a mac, but @axlewin has shown on their computer this was indeed an issue, and also that it's now fixed.

I've checked and we're still not saving UNKNOWN to the database when somebody registers without making a selection.

PR #2277, opened 2026-07-31
Adjust figcaption background on Ada

jacbn:
Sets the background to match that of the figure background, to better


sjd210 (approved):
I'm not sure that I see the current figure caption backgrounds as an issue, even with the original thread's context - but this doesn't seem to make things any worse and the code is all okay, so I'm happy for this to go in based on others' opinions.

PR #2280, opened 2026-07-31
Adapt expansion-layout to content-controlled expanded-layout

sjd210:
Renamed to correspond with new `"layout": "expanded"` content property. 

Generally re-using the logic of `expansion-layout` for reorder/parsons questions, primarily as a toggle for accordion sections (adding a background to those sections similar to that of a `question-component`) and Ada pages, but it _can_ also be applied to expand the width of nearly any content block on Ada (although I don't confirm full compatibility, and don't provide toggles in the content editor). At the page-level, it reduces the size of margins, as had previously been hardcoded for **computer_science_stories**. Nested `expanded-layout`s will not have any effect.

---

While doing this, I also rewrote a few sections of the `phy/questions.scss` file that had been mangled over time from various contributions (e.g. rather than adding padding to components around submission buttons based on their presence, attach the padding to the buttons themselves). It should _hopefully_ now be easier to parse and edit while not having affected the layout of any questions (which I've checked quite extensively) - although I'd appreciate another set of eyes to see if anything did go wrong.

(also see https://github.com/isaacphysics/isaac-content-editor/pull/129)


src/app/components/content/IsaacQuestion.tsx
```diff
@@ -273,10 +272,10 @@ export const IsaacQuestion = ({doc}: {doc: ApiTypes.QuestionDTO}) => {
                         validationResponse={validationResponse}
                     />
                     : (!correct || canSubmit || (isFastTrack && (fastTrackPrimaryAction || fastTrackSecondaryAction))) && !locked &&
-                        <div className={classNames("submission-buttons d-flex align-items-stretch flex-column-reverse flex-sm-row flex-md-column-reverse flex-lg-row")}>
+                        <div className={classNames("submission-buttons d-flex align-items-stretch flex-column-reverse flex-sm-row flex-lg-row")}>
                             {isFastTrack 
                                 ? <FastTrackSubmissionButtons fastTrackPrimaryAction={fastTrackPrimaryAction} fastTrackSecondaryAction={fastTrackSecondaryAction} />
-                                : <Button {...checkAnswerButtonProps} className="mx-auto mt-3 w-100 w-sm-100 w-md-50 w-lg-50" type="submit">{submitButtonLabel}</Button>
+                                : <Button {...checkAnswerButtonProps} className="mx-auto w-100 w-sm-100 w-md-50 w-lg-50" type="submit">{submitButtonLabel}</Button>
```
jacbn, line 278: I think all the spacing changes here leave everything as they were, so fine with me. Just a note that FT questions are (still) a bit funky with lots of bottom padding, if you wanted to fix those too :) 

`/questions/ft_upper_q2b_stage1_mech_2018`:
Image
(dark mode is also broken for this – maybe it's best we make a card...)
(also, clicking the right button makes it look like you can edit the buttons contents – this one _does_ have a card already)

src/scss/cs/expanded-layout.scss
```diff
@@ -0,0 +1,20 @@
+@keyframes gutter-overflow {
+  0%  {width: 100%; @include make-col-offset(0, 8);}
+  60%  {width: 100%; @include make-col-offset(0, 8);}
+  100% {width: 150%; @include make-col-offset(-2, 8);}
+}
+
+@include media-breakpoint-up(lg) {
+  .expanded-layout:not(.expanded-layout .expanded-layout) {
+    @include make-col-offset(-2, 8);
+    width: 150%;
```
jacbn: I wonder if expanding the width beyond 100% is the correct way to go about this. I know it's how we used to do it, but with the addition of new content types that can be expanded, content only expands to the same outer width as each other provided they are the same inner width to begin with. Take this page (`/questions/999f1379-f03e-4767-9a4b-bc71b6df5cc4`), for example:

Image

All of the content is expanding to slightly different widths, which ends up looking a bit funny. 

It may be the case that this is okay, because the _full-page_ expanded layout is likely to cover for the case where multiple elements require expanding (and this works great, since all content is bound within the expanded-width container). If, however, there is a case for multiple expanded elements on a page, I would rather we mirror the approach taken by containers whereby each breakpoint has a fixed width that specifies the extent to which all expansion content expands.

See containers:

```scss
.container {
    @each $bp, $_ in $container-breakpoints {
        @include media-breakpoint-only($bp, $container-breakpoints) {
            width: auto;
            padding-left: map.get($spacing-container-margins, $bp);
            padding-right: map.get($spacing-container-margins, $bp);
        }
    }
}
```

We could mirror this via e.g.:
```scss
.expanded-layout:not(.expanded-layout .expanded-layout) {
    @each $bp, $_ in $expansion-breakpoints {
        @include media-breakpoint-only($bp, $expansion-breakpoints) {
            width: map.get($expanded-bp-widths, $bp);
            justify-self: center; // avoid negative left margin calculation
        }
    }
}
```
where `expansion-breakpoints` contains the subset of `grid-breakpoints` for which an expansion layout should be applied.

This should create something like this:

Image

jacbn (commented):
It's a bit tricky to test the interplay of all the padding changes at once, but the VRTs have been stable and I can't see any (newly) bad looking questions in my search, so I'm cautiously happy to have this go in. I did have a point about the way we do the actual expansion, but it may be that you feel we don't need this. Let me know what you think!

src/app/components/content/IsaacQuestion.tsx
sjd210: Should we even be showing the fasttrack options here at all when not in a fasttrack gameboard? Quick questions are programmed differently in terms of how they display, but looking back it feels like they were just missed when this change was made: https://github.com/isaacphysics/isaac-react-app/commit/660c782f52f9062f330d9931f5b18024ecd2892f.

Also unrelated to this PR obviously, but probably something to think about while we make a separate card.

sjd210: I've fixed the spacing, and incidentally while doing so the bug where it appears you can edit the button's contents - but for the more general fasttrack issues including the darkmode fix, I think we should be sharing logic with `IsaacQuestion` and fixing them both together. I'll make sure all these things are properly recorded in cards 🫡

src/scss/cs/expanded-layout.scss
sjd210: Yes this does seem sensible. I had only been imagining either expanded layout being applied to the whole page, to reorder questions on existing pages, or to multiple accordion sections on the same page (which if all at the section-level would expand to the same width) - but it does feel like we shouldn't be relying on that. This is much more future-proof 👍

src/scss/cs/expanded-layout.scss
```diff
@@ -1,50 +1,43 @@
...
+    $expanded-width: map.get($expanded-bp-widths, $bp);
 
-  .expanded-layout:not(.expanded-layout .expanded-layout) {
-      @each $bp, $_ in $expansion-breakpoints {
-          @include media-breakpoint-only($bp, $expansion-breakpoints) {
-              width: map.get($expanded-bp-widths, $bp);
-          }
-      }
-  }
+    .expanded-layout-container:not(.expanded-layout .expanded-layout-container) {
```
jacbn, line 24: I am finding the grid parent a little confusing and it could use some explanation. Does the grid container let you do things you can't do without it, or is this just future-proofing to ensure the parent is a valid display type? `justify-items` is equivalent to `justify-self` on child elements, which to me seems not to need a container at all as the parent (`content-chunk`) is just a regular `display: block` (which works with `justify-self` fine)? It may be that the parent isn't always `content-chunk`, in fairness, but I'm not 100% on that.

Alternatively, is it an issue you found with the nesting (`:not(.expanded-layout .expanded-layout)` etc)? I have a feeling we honestly don't need this at all with container-like approach, since all expanded layouts can be the same width regardless of nesting-ness.

jacbn (approved):
This does work fine so would be happy to merge it, but it would be nice to get some clarity on what the grid parent is for?

src/scss/cs/expanded-layout.scss
sjd210: First of all, the parent indeed isn't always a `content-chunk` - see accordion sections marked as `"layout": "expanded"` such as the first section of [your test page](http://localhost:8003/questions/999f1379-f03e-4767-9a4b-bc71b6df5cc4) where the content is a child of the `expanded-layout` rather than the other way around. Because of this, I'd like a consistent parent we can control the display type of.

Secondly, a regular `display: block` does not fully work with `justify-self` fine (as of yet) anyway. It seems like the [specifications of justify-self](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/justify-self#formal_definition) say that it should apply to this context, but the [browser support is limited](https://caniuse.com/wf-justify-self-block). I didn't actually know this while developing it, I just fortunately happen to use Firefox which is one of the ones lacking support so that caught it.

As for the `:not(.expanded-layout .expanded-layout)`, fundamentaly I think expanded-layouts shouldn't be nested in content anyway, but we ought to build defensively - and even with this container-style approach it stops some oddities like an expanded accordion section animation playing on an already expanded page. We _could_ apply some rules but not others, but I think it's easier to keep it uniform like this.

sjd210: I'll leave a comment describing why we're doing this before merging. If you're confused now, someone is bound to be in the future too.

PR #2281, opened 2026-07-31
Fix a broken Link object

ioiototm:
...and add more precise link in LLM Consent section


barna-isaac (approved):
Looks good, thanks for the fix!

PR #2282, opened 2026-08-03
show microsoft sign-in buttons on ada

barna-isaac:
This adds Microsoft as an SSO provider to Ada.


src/app/components/pages/AuthError.tsx
```diff
@@ -72,17 +73,19 @@ const AccountNotLinked  = ({ state }: { state?: State }) => {
...
     <p>If you need more help signing in, <ContactUs />.</p>
 </>;
 
-const SSOLink = ({ children }: { children: ReactNode}) =>
+const SSOLink = ({ children }: { children: ReactNode}) => isPhy && 
     <Link to="/pages/single_sign_on" aria-label="Link to Single Sign-On documentation">{children}</Link>;
 
 const ContactUs = () => <Link to="/contact" aria-label="Link to contact form">contact us</Link>;
+
+const site = siteSpecific("Isaac", "Ada CS");
```
sjd210: I'm not doing a full review, but while I'm having a quick look: Can we please use the `SITE_TITLE_SHORT` constant here instead? It has the same contents as this, but is more consistent and resilient to changes.

(+ same for the one in the test file)

jacbn: We already have this constant. It's `SITE_TITLE_SHORT`.

src/app/components/pages/AuthError.tsx
```diff
@@ -48,16 +48,17 @@ const AccountNotLinked  = ({ state }: { state?: State }) => {
...
+            in.
         </p>
         <ul>
             <li>
                 If you&apos;ve not yet enabled sign-in with {provider}, first log in with another method (e.g. email and
-                password). Then, on <Link to="/account#security" aria-label="My Account link">My Account</Link>, next
-                to &quot;{provider}&quot;, click &quot;Link&quot;. <SSOLink>Read more about signing in with {provider}.
-                </SSOLink>
+                password). Then, on <Link to="/account#security" aria-label="My Account link">{siteSpecific(
+                    "My Account", "My Ada > Account")}</Link>, on &quot;Security&quot;, next to &quot;{provider}&quot;, click &quot;Link&quot;.
```
jacbn:
```suggestion
                    "My Account", "your Account page")}</Link>, on &quot;Security&quot;, next to &quot;{provider}&quot;, click &quot;Link&quot;.
```

I think this reads a lot more nicely, particularly for screenreaders.

src/test/pages/AuthenticationSSO.test.ts
```diff
@@ -124,14 +125,14 @@ describe("Google SSO Authentication", () => {
...
+                        We've found an ${SITE_TITLE_SHORT} account with the email address from this Google account. However, the ${SITE_TITLE_SHORT}
                         account isn't configured to allow access to this Google account. You've either not enabled
-                        sign-in with Google on your Isaac account, or you used a different Google account to log in.`
+                        sign-in with Google on your ${SITE_TITLE_SHORT} account, or you used a different Google account to log in.`
                     );
                     expect(authenticationError.element).toHaveTextContent(dedent`
                         If you've not yet enabled sign-in with Google, first log in with another method (e.g. email and
-                        password). Then, on My Account, next to "Google", click "Link". Read more about signing in with
-                        Google.`
+                        password). Then, on ${siteSpecific("My Account", "My Ada > Account")}, on "Security", next to
```
jacbn:
```suggestion
                        password). Then, on ${siteSpecific("My Account", "your Account page")}, on "Security", next to
```

src/app/components/elements/modals/LoginOrSignUpModal.tsx
```diff
@@ -114,10 +114,12 @@ const LoginOrSignUpBody = () => {
...
-                        {isPhy && <div className="mt-2 mb-2">
+                        <div className="mb-2">
+                            <GoogleSignInButton/>
+                        </div>
+                        <div className="mb-2">
                             <MicrosoftSignInButton />
-                        </div>}
+                        </div>
                         {isPhy && <SsoHelpLink />}
                     </>}
```
jacbn, line 122: Can we clean up these loose divs?
```tsx
<div className="d-flex flex-column gap-2 mb-2">
    {isAda && <RaspberryPiSignInButton concise={true} />}
    <GoogleSignInButton/>
    <MicrosoftSignInButton />
</div>
```

src/app/components/pages/RegistrationStart.tsx
```diff
@@ -50,10 +50,12 @@ export const RegistrationStart = () => {
...
                             </div>}
-                            <GoogleSignInButton />
-                            {isPhy && <div className="mt-2 mb-2">
+                            <div className="mb-2">
+                                <GoogleSignInButton />
+                            </div>
+                            <div className="mb-2">
                                 <MicrosoftSignInButton />
-                            </div>}
+                            </div>
```
jacbn: As above

src/test/pages/AuthenticationSSO.test.ts
```diff
@@ -189,7 +193,7 @@ const authenticationError = {
     }
 };
 
-const dedent = (s: TemplateStringsArray) => s
-    .join('')
+const dedent = (strings: TemplateStringsArray, ...values: unknown[]) => strings
+    .reduce((acc, str, i) => acc + str + (values[i] ?? ''), '')
     .replace(/\s+/g, ' ')
     .trim();
```
jacbn: should this be here? it's really a `miscUtil`...

src/app/components/elements/modals/LoginOrSignUpModal.tsx
```diff
@@ -39,7 +39,7 @@ const LoginOrSignUpBody = () => {
         return null;
     }
 
-    return <Row id={"login-page"}>
+    return <Row id={"login-page"} className="mb-n3">
```
jacbn: This doesn't feel like the right way to solve this; we ought to only use negative margins sparingly and we can just remove the padding from the parent by fixing the type of `loginOrSignupModal` below.
```suggestion
    return <Row id={"login-page"}>
```

alongside

```tsx
export const loginOrSignUpModal: ActiveModalProps = {
    centered: true,
    bodyContainerClassName: "p-0",
    body: LoginOrSignUpBody
};
```

jacbn (changes requested):
This works, but I think there's a lot of cleanup we can do here to tidy up the DOM a little.

Just a note here that we'll also need to add this to the new signup flow page for 11-12 users who try to signup with email.

jacbn (approved):
LGTM. We discussed the modal-in-VRT hack and we couldn't find much better. We have at least extracted this to a new function so that we can use this elsewhere without needing to repeat the hack multiple times.

PR #2284, opened 2026-08-03
Fix verification banner link

ioiototm:


sjd210 (approved):
Straightforward enough. Looking good! 👍

PR #2287, opened 2026-08-04
Prevent autotranslation of LaTeX content

jacbn:
The `translate` attribute is applied to all children, so doesn't need to be nested.

### DIFF
--- a/src/app/components/elements/markup/latexRendering.ts
+++ b/src/app/components/elements/markup/latexRendering.ts
@@ -319,7 +319,7 @@ export function katexify(
                 }
 
                 // prevent browser auto-translation
-                katexRenderResult = katexRenderResult.replace('<span class="katex', '<span translate="no" class="katex');
+                katexRenderResult = katexRenderResult.replace('<span class="katex"', '<span class="katex" translate="no"');
 
                 output += katexRenderResult;
 
--- a/src/app/components/elements/markup/latexRendering.ts
+++ b/src/app/components/elements/markup/latexRendering.ts
@@ -319,7 +319,7 @@ export function katexify(
                 }
 
                 // prevent browser auto-translation
-                katexRenderResult = katexRenderResult.slice(0, 6) + ' translate="no" ' + katexRenderResult.slice(6);
+                katexRenderResult = katexRenderResult.replace('<span class="katex', '<span translate="no" class="katex');
 
                 output += katexRenderResult;
 
--- a/src/app/components/elements/markup/latexRendering.ts
+++ b/src/app/components/elements/markup/latexRendering.ts
@@ -318,6 +318,9 @@ export function katexify(
                         }"`);
                 }
 
+                // prevent browser auto-translation
+                katexRenderResult = katexRenderResult.slice(0, 6) + ' translate="no" ' + katexRenderResult.slice(6);
+
                 output += katexRenderResult;
 
                 index = match.index + match[0].length;


src/app/components/elements/markup/latexRendering.ts
```diff
@@ -318,6 +318,9 @@ export function katexify(
                         }"`);
                 }
 
+                // prevent browser auto-translation
+                katexRenderResult = katexRenderResult.replace('<span class="katex', '<span translate="no" class="katex');
```
sjd210: This works, but the way you've written it is quite odd to me. Is the hanging `"` and ending with `class="katex` meant to allow it to match other classes like "katex-html" too? If so, I don't think it's doing anything helpful.

KaTeX always starts with a `<span class="katex">` so this replace rule always hits that anyway, and I'd prefer we order the replacement the other way around as `<span class="katex" translate="no"` for the sake of being able to possibly add conditional properties using the same `<span class="katex"` search term in the future.

```suggestion
                katexRenderResult = katexRenderResult.replace('<span class="katex"', '<span class="katex" translate="no"');
```

sjd210 (changes requested):

src/app/components/elements/markup/latexRendering.ts
jacbn: > Is the hanging " and ending with class="katex meant to allow it to match other classes like "katex-html" too?

Yes, exactly this. There are many types of katex class generated by the above code, and I want to hit the outermost one. Your suggestion _won't_ hit `katex-display`, `katex-html` nor `katex-mathml` if any of those are ever outermost.

> KaTeX always starts with `<span class="katex">`

Does it? In the display case, it starts with `<span class="katex-display">`, which _looks_ to always contain precisely one inner `<span class="katex">`, but I don't know how we'd guarantee that. What if certain latex displays generate multiple? We could prefer `replaceAll`, I suppose.

Our previous modifications to the latex string also only use a single `replace`, though, so if it's broken here it's broken for those too. :shrug:.

jacbn: I'd be _okay_ to use your suggestion given we seem to rely on the assumptions required to make it work elsewhere, but I'd appreciate your opinion here.

sjd210: Hrm, well we'd been making the assumption before, and their documentation seems to indirectly imply it: e.g. https://katex.org/docs/api#server-side-rendering-or-rendering-to-a-string / https://katex.org/docs/error - but you've convinced me to check further.

We call katex using `katex.renderToString`, first using `output: "html"` to create `katexRenderResult`, then possibly using `output: "mathml"` to create `katexMathML` which is added as a child of `katexRenderResult` (as an alternative to an aria-label for screen-reader text).

Looking at their code: [renderToString](https://github.com/KaTeX/KaTeX/blob/5a5bf2060431aaed5eb3b549f6fab24f55e9e70a/katex.ts#L71) calls [buildTree](https://github.com/KaTeX/KaTeX/blob/5a5bf2060431aaed5eb3b549f6fab24f55e9e70a/src/buildTree.ts#L33) (via renderToDomTree). In the `output: "html"` case, the created span is "katex", and in the `output: "mathml"` case ([buildMathML](https://github.com/KaTeX/KaTeX/blob/5a5bf2060431aaed5eb3b549f6fab24f55e9e70a/src/buildMathML.ts#L283))  the created span can _either_ be "katex" or "katex-mathml" (and in both output cases they can optionally be wrapped with "katex-display", but importantly the above classes will still occur in the tree before any content). 

If in our code we were using the `output: "mathml"` directly as the output this could indeed be a problem, but since its only ever added as a child, and so the outermost element is always either `<span class="katex">` OR `<span class="katex-display><span class="katex">`, we are safe to use `<span class="katex"` as our replace term.

jacbn: Okay, sounds good. I don't love relying on a library not to change, but all our CSS and the other functions here would break if something changed and we'd definitely notice that – so any further precautions existing only here feel _too_ safe to be worth the time. Let's go with this. Thanks for taking a look!

sjd210: It seems fundamental enough that they at least wouldn't change it without that being a major release, and major releases of libraries _in general_ are prone to break things - so this doesn't seem too off-base to me.

PR #2289, opened 2026-08-05
Refactor inline code block styling in phy theme

ioiototm:


src/scss/phy/typography.scss
```diff
@@ -54,9 +54,11 @@ code {
   display: inline-block;
   color: unset;
   box-shadow: 0 0 10px 0 rgba(0, 0, 0, 0.08);
-  background: var(--color-neutral-white);
+  background: rgba(0, 0, 0, 0.1);
```
sjd210: The spacing and implementation here is good - you've done exactly what was asked on the card. That said, I have two problems with it on actually navigating the site that I think we can handle better.

1) This works very well on backgrounds with either a lot of colour, or no colour at all (white or grey) - but against a background with very light use of colour, this 10% black completely overpowers it. 

e.g. 

Image

> We'd slightly prefer this `availableUnits` to come through as a darker green against a light green background, but instead it's almost entirely grey.

2. Making the background slightly darker works well for contrast when the site is in light mode, but not for dark mode.  Our rough principle is that processes like this should be inversed for dark mode - so instead of making it 10% darker, it should be 10% lighter

e.g.

Image

> `availableUnits` is in a different font, but the background is practically identical to that of the surrounding text so it doesn't stand out aside from that.

---

I have a solution in mind that should fix both of these issues, but it gets in the weeds of how our colour system works so I won't ask you to implement it by yourself. We can call to go over this and do some pair programming instead.

sjd210 (changes requested):

sjd210:
@ioiototm Also see the failing test on this branch. In this case, it means that some pages have visually changed and we need to review whether or not we approve of that before merging. We can also go over this process on call.

ioiototm:
I fixed the background with `color-mix(in srgb, var(--color-neutral-900) 10%, transparent);` - after discussing it with @jacbn, we were looking through --subject-color-900 and --color-neutral-900. It seemed the subject colour would look a tiny bit stranger on different coloured backgrounds, such as the teacher notes and the partially correct answer background. 

--subject-color-900 in Teacher note and normal text field:
image

--subject-color-900 in partially correct:
image


--color-neutral-900 in Teacher note and normal text field:
image


--color-neutral-900 in partially correct:
image

--color-neutral-900 dark mode:
image
image


I think in this case the --color-neutral one works a bit better? This still has the light and dark mode change, but it's not tied to a theme.

I also made it so it should only affect code blocks.

sjd210 (approved):
I did consider `--subject-neutral-900` over `--subject-color-900` for the same reasons, but I was prioritising its appearance on a lightly coloured subject background over that of the much less common case of a non-subject coloured background in a subject context (e.g. other than one beta page, there are currently no teacher notes with an inline code block included).

That said, I don't think the grey background looks _bad_ in the component - especially considering we previously had a pure white background there anyway - so I'm willing to concede on that point. I'm sure we could create a more robust system that is incorporated into all background colours to give them more suitable offsets, but that sounds like far too much work for the use-case. This looks good enough 👍

(p.s. Remember to check the PR from the failing Visual Regression Test, and merge it into this branch if you approve the changes. I'll leave you to do that, so won't merge this PR yet)

PR #2291, opened 2026-08-07
Fix isDobOverN date validation logic

ioiototm:
Just changed it to return early, and then check today's date minus n with the date of birth.

Leap years would normalise to the next day so in this super niche scenario it would still work (if today is 29th of Feb, ten years ago wasn't a leap year, so it would normalise to 1st of March)


jacbn (approved):
LGTM!

PR #2293, opened 2026-08-07
Improvement/sticky sidebar

barna-isaac:
This implements sticky sidebars.

Here are some controversial choices I made:

1) I prioritised the case when the sidebar is short enough there’s no need for a scrollbar. In such a case, the page looks exactly as it did before. When there is a scrollbar, the sidebar’s contents is now narrower, to acommodate the width of the scrollbar. Maybe the reviewer knows a way to have the scrollbar eat into the padding we already have on the page? Unfortunately I couldn’t do this easily (considered relying on `scrollbar-gutter: stable`, but whether this takes up any space still depends on whether the os shows scrollbars).
2) when the main content becomes shorter, we loose scroll position and there’s a big jump. This might happen, eg. [on the Question finder](https://dev.isaacscience.org/questions), see attached video. I don't think there's an easy fix for this one: the browser remembers we're on the bottom of the page, but with the page a lot shorter, the bottom of the page becomes the footer, which doesn't have the scrollbar.
https://github.com/user-attachments/assets/480366b2-1030-4f96-80a8-bbde4d359f3c
3) For simplicity, I made every sidebar sticky. There are a few pages, like  "[News](https://dev.isaacscience.org/news)", where sticky sidebars feel unnecessary, as the sidebar doesn't really contain anything meaningful in the first place.

### DIFF
--- a/src/app/components/elements/layout/SidebarLayout.tsx
+++ b/src/app/components/elements/layout/SidebarLayout.tsx
@@ -65,11 +65,13 @@ export const ContentSidebar = (props: ContentSidebarProps) => {
     const sidebarContext = useContext(SidebarContext);
     if (!sidebarContext?.sidebarPresent) return <></>;
 
-    const { className, buttonTitle, hideButton, optionBar, ...rest } = props;
+    const { className, buttonTitle, hideButton, optionBar, children, ...rest } = props;
     return fullSidebarLayout
         ? siteSpecific(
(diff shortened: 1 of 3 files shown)


barna-isaac:
More weirdness:
- on question finder search box, focus outline is now clipped
- pages with unexpected jumps: any page that might become shorter in response to a sidebar action. Question Finders, Concept Finders, Test finder, "My tests" (and probably a lot more!)

PR #2295, opened 2026-08-12
Improvement/lighthouse accessibility changes

ioiototm:
This pull request covers most of the accessibility issues raised by @jacbn with Lighthouse on the 3rd of August. 

Most of the changes are small, h3 -> span/div, adding aria-labels, etc. 

The big change is in the MyAccount file, where a few helper functions have been added in order to make the tab selection accessible:

1. Once you have focus on the tab list, you can use left or right to travel, and you have to press space or enter to select. I haven't done automatically selecting the tabs since there is a possibility of a pop up in case of unsaved changes. 
2. Home/End put the cursor at the start or end of the tab list (although, on MacOS it's not working, since Mac does the keys as Command+Left/Right). 

This is following the standard at [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tablist_role#keyboard_interactions).


src/app/components/pages/MyAccount.tsx
```diff
@@ -362,12 +402,16 @@ export const MyAccount = ({user}: AccountPageProps) => {
         <ShowLoading until={editingOtherUser ? userToUpdate.loggedIn && userToUpdate.email : userToUpdate}>
             {user.loggedIn && userToUpdate.loggedIn && // We can guarantee user and myUser are logged in from the route requirements
                 <div className={siteSpecific("w-lg-75", "card")}>
-                    {isAda && <Nav tabs className="my-4 flex-wrap mx-4" data-testid="account-nav">
+                    {isAda && <Nav tabs role="tablist" aria-label="Account settings" className="my-4 flex-wrap mx-4" data-testid="account-nav">
                         {ACCOUNT_TABS.filter(tab => !tab.hidden && !(editingOtherUser && tab.hiddenIfEditingOtherUser)).map(({tab, title, titleShort}) =>
                             <NavItem key={tab} className={classnames({active: activeTab === tab})}>
```
sjd210: There's a slightly complicated case here. A `<ul>` (`<Nav>`) is expecting `<li>` (`<NavItem>`) as a direct child (for both accessibility and styling), and a `role="tablist"` expects `role="tab"` as a direct child (for accessibility), BUT `role="tab"` should be on the focusable element, which in this case is the `<a>` (`<NavLink>`) beneath this.

On its surface this is incongruous - we cannot satisfy all three of those cases at once. Fortunately, the aria [`role="presentation"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/presentation_role) solves this issue. This essentially skips the `<li>` from the accessibility tree so that it is just seen as a `"tablist"` > `"tab"`, while not removing it from the HTML tree so we retain `<ul>` > `<li>`.

(Note: `role="none"` is synonymous, but `"presentation"` seems more widely used so I think we should stick with that)

```suggestion
                            <NavItem key={tab} role="presentation" className={classnames({active: activeTab === tab})}>
```

src/app/components/elements/panels/UserProfile.tsx
```diff
@@ -57,7 +57,7 @@ export const UserProfile = (props: UserProfileProps) => {
 
     return <MyAccountTab
         leftColumn={<>
-            <h3>Account details</h3>
+            <h2>Account details</h2>
```
sjd210: This make sense logically to be `<h2>`, but visually it looks off - especially on Isaac where this makes this heading bigger than the actual `<h1>` page heading. 

I think we can keep the rank logic, but also maintain the styling. i.e.

```suggestion
            <h2 className="h3">Account details</h2>
```

Also, Lighthouse doesn't catch this because it only does an initial page load, but this `<h3>` is an issue for every tab on the My Account page. In reality we care beyond just the page load, so they should all have the same treatment as this.

src/app/components/site/cs/HeaderCS.tsx
```diff
@@ -91,9 +91,9 @@ export const HeaderCS = () => {
 
                         <NavigationSection className={"text-start-nav"} topLevelLink to="/contact" title={"Contact us"}/>
 
-                        <div className={"navbar-separator d-nav-none d-block"}/>
+                        <div className={"navbar-separator d-nav-none d-block"} aria-hidden="true"/>
 
-                        <div className={"ms-nav-auto"}></div>
+                        <div className={"ms-nav-auto"} aria-hidden="true"></div>
```
sjd210: This is unnecessary. An empty `div` contains no text or interactive children, so already provides no content to the accessibility tree.

src/app/components/pages/MyAccount.tsx
```diff
@@ -387,7 +431,7 @@ export const MyAccount = ({user}: AccountPageProps) => {
                         <TabContent activeTab={activeTab} key={clearUnsavedChanges ? activeTab : undefined}>
                             {/* the key above ensures that any state inside tabs is reset to initial values if activeTab is changed while accountInfo is dirty 
                                 (i.e. user has unsaved changes they do *not* want to commit) */}
-                            <TabPane tabId={ACCOUNT_TAB.account}>
+                            <TabPane tabId={ACCOUNT_TAB.account}{...tabPanelProps(ACCOUNT_TAB.account)}>
```
sjd210: I've got a couple suggestions about these props:
1) There's _slight_ redundancy, but I'd rather we keep these props actually local to their object, so one doesn't need to scroll back and forth to find out values when maintaining this code.
2) Since it's an enum value, just using the `ACCOUNT_TAB` directly is making the id contain a number rather than title, which isn't very descriptive. We'd rather it be intuitive which id belongs to which object.
3) Not here, but it would be nice if the `NavLink` has an `aria-controls` property linking to this id to make the connection two-way.

Something like this should do to cover the first two of these points for this pane specifically, but there are changes to make in several places if you want to implement this.

```suggestion
                            <TabPane 
                                tabId={ACCOUNT_TAB.account} id={`account-panel-${ACCOUNT_TABS[ACCOUNT_TAB.account].title.toLowerCase()}`} 
                                {...(isAda && {"role": "tabpanel", "aria-labelledby": `account-tab-${ACCOUNT_TABS[ACCOUNT_TAB.account].title.toLowerCase()}`})}
                            >
```

sjd210 (changes requested):
I think you've done a good job here, but unfortunately they were both a lot of things to find and some subtleties that need some more knowledge of accessibility/our codebase. 

If you've got any questions about the changes I'm requesting, please do feel free to drop a message and ask 😊

---

Quick list of individual violations I found on the reviewed pages when re-running lighthouse tests. I think it makes sense to also bundle fixes for them here:
- This [question finder button](https://github.com/isaacphysics/isaac-react-app/blob/32db1da7b2a203265743b1a93a89482f221a40e3/src/app/components/elements/panels/QuestionFinderFilterPanel.tsx#L143) on <=md screens (Note that the `<i>` within it is `aria-hidden`, but the button itself should still be labelled)
- Theses [h4s](https://github.com/isaacphysics/isaac-react-app/blob/648454abe6eed21370b867be704725147ee21c7b/src/app/components/content/IsaacReorderQuestion.tsx#L63-L85) for IsaacReorderQuestion (and the equivalent ones for IsaacParsonsQuestion)
- The [title tag](https://github.com/isaacphysics/isaac-react-app/blob/648454abe6eed21370b867be704725147ee21c7b/src/app/components/elements/modals/ActiveModal.tsx#L48) for modal headers (although be careful here because `undefined` sets them to `<h5>`. We need to set a non-`<h*>` tag manually and then ensure that `.modal-title` has correct styling by itself on both sites)
- All the `<h*>`s in [related questions](https://github.com/isaacphysics/isaac-react-app/blob/648454abe6eed21370b867be704725147ee21c7b/src/app/components/elements/RelatedContent.tsx#L62), including the fact that there are `<h4>` as direct children of a `<ul>` (`<ListGroup>`). Only `<li>` should be children here, so we should extract those.
- [Topic headers](https://github.com/isaacphysics/isaac-react-app/blob/ba364f6b9a657ecf4db77bf95040107b3bd5f1db/src/app/components/pages/AllTopics.tsx#L53)
- [Assignment card titles](https://github.com/isaacphysics/isaac-react-app/blob/ba364f6b9a657ecf4db77bf95040107b3bd5f1db/src/app/components/elements/MyAssignmentsContents.tsx#L69)

There are also definitely a lot more issues with _most_ other `<h*>` uses in the codebase, but not in the pages that were tested for the report - so I'm collating those for another card.
 
Also have you intentionally not addressed **"The user's focus is directed to new content added to the page - when questions are listed, focus remains on the "Apply filters" button"**? If so that's fine, I'd just like to make sure its recorded somewhere.

ioiototm:
I have started implementing the changes you requested, which I agree with all of them. 

I knew about the div being empty and still added the aria-hidden since the lighthouse report was complaining about that, but if you think it's being overly twitchy on that, then we can undo that change.

I've tried doing the account tab stuff, and I made it so it has a slug function, and replaces the space with "-" since if a tab contained more than a single word, the space would break the tag.

I went through all the user tabs and tried to fix the headings, I think I got everything.

I am starting the work on the other headings you mentioned, I haven't done anything with them. 

And for the final thing, the about the user focus, I haven't done that, it must have slipped my checks, I will look into that as well.

src/app/components/elements/panels/TeacherConnections.tsx
```diff
@@ -177,15 +177,15 @@ export const TeacherConnections = ({user, authToken, editingOtherUser, userToEdi
         }
     }
 
-    const SectionHeading = siteSpecific("h4", "h3");
+    const SectionHeading = siteSpecific("h3", "h2");
```
sjd210, line 180: Nothing that you need to change since I've already dealt with this in the broader <h*> tag PR , but I don't think it's quite right to have `<h2>`s semantically for Ada here. The `<h2>` of the page is "Connect to your teacher" (as you've put, and equivalent to each other account page), and the sections in the `rightColumn` are subheadings of that - so these should just be `<h3>`s on both sites.

You haven't affected the styling and it's no more semantically wrong than it was before so don't worry about it - I just want you to be aware.

src/app/components/pages/MyAccount.tsx
```diff
@@ -431,7 +430,12 @@ export const MyAccount = ({user}: AccountPageProps) => {
                         <TabContent activeTab={activeTab} key={clearUnsavedChanges ? activeTab : undefined}>
                             {/* the key above ensures that any state inside tabs is reset to initial values if activeTab is changed while accountInfo is dirty 
                                 (i.e. user has unsaved changes they do *not* want to commit) */}
-                            <TabPane tabId={ACCOUNT_TAB.account}{...tabPanelProps(ACCOUNT_TAB.account)}>
+                            <TabPane
+                                tabId={ACCOUNT_TAB.account}
+                                role={isAda ? "tabpanel" : undefined}
+                                id={isAda ? accountPanelId(ACCOUNT_TAB.account) : undefined}
+                                aria-labelledby={isAda ? accountTabId(ACCOUNT_TAB.account) : undefined}
```
sjd210: There's no reason that we need to be applying this pattern only to Ada. The items in Isaac's [MyAccountSidebar](https://github.com/isaacphysics/isaac-react-app/blob/main/src/app/components/elements/sidebar/MyAccountSidebar.tsx) are also tabs (displayed differently, but serving the same purpose) so we can apply the properties that `NavLink` uses in this file to each of those, and then simplify the props here by removing the ternary and making them the same across sites. 

(Don't worry about porting the keyboard navigation over there. It's more important that it's consistent with other sidebars than anything else. If we _were_ to change it, we should change all of them - which is a much bigger task)

src/app/components/site/cs/HeaderCS.tsx
```diff
@@ -91,9 +91,9 @@ export const HeaderCS = () => {
 
                         <NavigationSection className={"text-start-nav"} topLevelLink to="/contact" title={"Contact us"}/>
 
-                        <div className={"navbar-separator d-nav-none d-block"} aria-hidden="true"/>
+                        <div className={"navbar-separator d-nav-none d-block"}/>
 
-                        <div className={"ms-nav-auto"} aria-hidden="true"></div>
+                        <div className={"ms-nav-auto"}></div>
```
sjd210: > I knew about the div being empty and still added the aria-hidden since the lighthouse report was complaining about that, but if you think it's being overly twitchy on that, then we can undo that change.

Ah sorry, I see the issue now. Lighthouse was complaining because of using `<div>`s as direct children of a `<ul>` - which is fair enough, I was being unobservant. Removing the `aria-hidden` definitely wasn't the correct approach, but I don't think just marking them as hidden is quite right either - especially since we also have cases like this at L118 and L127 with the same problem.

Fundamentally the issue here is that these borders are attached to their own non-list elements when they're meant to correspond to the existing list elements anyway. We can take inspiration from our newer approach at [HeaderMenuPhy](https://github.com/isaacphysics/isaac-react-app/blob/main/src/app/components/site/phy/HeaderMenuPhy.tsx) in which only `NavigationSection`s exist at this level, and a border is created using a conditional `border-end` class on one of those elements. We could do a similar thing here e.g.
```js
<NavigationSection className={"text-start-nav me-nav-auto navbar-separator"} topLevelLink to="/contact" title={"Contact us"}/>
```
with `navbar-separator` restyled to be a bottom-border that only appears at >=nav screens

(+ alike for My Ada + I don't think the one at L127 is doing anything meaningful anyway so we can probably just remove it)

sjd210 (changes requested):
Good job with what was added. Everything is on the right track with `aria-label`s and such (I'm sure there are still some missing across the site, but I don't think we're too bad for it - and it's good to fix them where spotted).

Two more small-ish changes requested - neither of which are new to your current batch of commits, just things I didn't spot properly on the first go around that we still ought to address.

ioiototm:
The navbar div problem has been fixed, and the tabs and panels on Isaac have the tab roles. 

Only problem was that adding role tablist to the `<ul>` is giving an error, since you shouldn't have interactive roles on non interactive elements. And I think the Ada one, with Nav, is actually the same problem, but the linter doesn't catch `<Nav>` as being problematic?

sjd210:
Hrm, my screenreader seems to be processing the tablist > presentation > tab setup fine, so I'm not too worried and think your comment was the right solution. Maybe worth keeping in mind for the future if anything similar comes up, but I'm happy to merge this now 👍

PR #2301, opened 2026-08-13
Change sig figs feedback wording

axlewin:
Also increase the font size of the smaller feedback text.


barna-isaac (approved):
Looks good :).

I agree with your decision to rely on `rem`'s for setting the explanation's font size, even though it's inconsistent with how the size of the big "Correct" / "Incorrect" text is controlled, for which we use pixels. Other text on the page sizes using `rem`'s, so it's better to remain consistent with that.

PR #2303, opened 2026-08-14
Ensure font descenders are visible in Ada modals

barna-isaac:
This adds a VRT for the "Required Account Information" modal, and fixes a visual bug on Ada where font descenders were cut off. 
Before:
Screenshot 2026-08-14 at 11 21 24

After (the highlight is just an accident of how I took the screenshot):
Screenshot 2026-08-14 at 11 23 43

I've agonised quite a bit over this change. In `cs/modals.scss`, a comment above these rules complains our Ada input styles are currently a mess, and @jacbn has made the very valid observation that inputs shouldn't look different because they're on a modal. Indeed, we currently employ at least two input styles: On the My Assignment page, inputs have a 1px margin with a height of 38px, and within modals and on the My Account page, inputs are 45 px high and have a margin of 2px.

This inconsistency likely comes from taking our designs too literally, which also employ a variety of input styles (just on the designs for the "My Ada" page, some inputs are 48px high with a 1px border, some are 48px high with a 2px border, and other inputs are 56px high with a 2px border).

@jacbn has suggested getting rid of any custom styles that an input receives when it's within a modal. In the end, I've decided against this because I'm not sure which of our many input styles would be the correct new style (just deleting the rule would make these appear less consistent with, eg. the inputs from the My Account page). Although I originally suspected the 12px padding was just an artefact copied over from the designs (which don't show the modal inputs without any text so can't have been intentional about this choice), on a second thought I ended up keeping the horizontal padding. This pushes the element's text to the left, and saves a little space, which might be useful on mobile screens.

Unfortunately VRT's won't help standardise our input styles, but I'm hoping they'll at least help  avoid the kind of glaring mistake these screens originally had. 

We still have an unresolved issue with inputs where the chevron renders too close to the text, but the change here doesn't make that problem any worse.


jacbn:
> I've decided against this because I'm not sure which of our many input styles would be the correct new style

But surely picking one (whichever is your favourite!) and going with it is better than having lots of arbitrary styling on lots of different pages?

@sjd210 has been doing a lot of good, opinionated codebase cleanup recently. Perhaps they can be the arbiter here 😆

barna-isaac:
@jacbn , I've created two new cards for the issues I didn't resolve in this PR.
- https://trello.com/c/K7eS8PBj/6737-dropdown-chevron-intersects-with-text, for resolving a specific visual bug with dropdown chevrons,
- and https://trello.com/c/NQQpQ4z9/6736-consistently-style-dropdowns so we use a consistent style for dropdowns everywhere

sjd210:
I'll leave this comment on the new card too, but since I was summoned here:

I think we should be using a `solid 2px $gray-160 (#666)` border with a consistent `border-radius` across the board, because this is much more consistent with the rest of the site e.g. See the borders on the My Account nav items, the dividers on the homepage, the button styling (which to be fair is 3px not 2px, but deviating _less_ is good)
However, I think there's legitimate reason to have different heights for different inputs. In a form the inputs are the main focus so should be big enough to always see, but as filters at the top of a table they are a secondary focus and may not be used by everyone. It feels out-of-place to have them so big.

Our aim shouldn't be that all inputs look identical. It should be that consistent styling rules are applied, such that any difference is clearly intentional.

(Also the card just says dropdowns, but dropdowns should be consistent with other inputs used in the same context too!)

sjd210:
@ioiototm If you're waiting, I also think this PR itself is fine since we've agreed to fix other problems elsewhere.

That said, please do hold out on merging it until after the release has been completed anyway. We do a merge freeze between bringing up test and releasing the site, so that nothing that hasn't been regression tested goes in (except for possibly emergency fixes)

ioiototm (approved):

PR #2305, opened 2026-08-17
Use media-breakpoint-* mixins consistently (+ clean up mixins file in general)

sjd210:
We had three different ways of conditionally applying styling rules according to breakpoint values in the codebase:
- `@media (min-width: $VALUE)`
- `@include respond-above($bp)`
- `@include media-breakpoint-up($bp)`
(and equivalents for below and between)

Firstly, the `@media (min-width: $VALUE)` version does not adapt at all to our adjusted breakpoints for sidebar width on Ada, so are causing odd behaviour on pages such as the question finder. Similarly, `respond-above` COULD adapt to our adjusted breakpoints except it uses a separate list of values from the `$grid-breakpoints` list in `isaac.scss` which hasn't been adjusted, so may also cause problems.

Secondly, this inconsistency just makes it slightly harder to maneuver the codebase (e.g. searching for what styles change when switching from a sm to a md breakpoint, maintaining files with multiple different types of rules).

Thus, this PR replaces all instances of `@media (min-width: $VALUE)` and `@include respond-above($bp)` with the bootstrap-provided `@include media-breakpoint-up($bp)` (and equivalents).

> _The only complicated cases were:_
> - _The login/sign up modal - for which a min-height rule was used, but is no longer having an effect so could just be removed_
> - _The equality/graph sketcher modals - for which a non-breakpoint width was used for the question reminder box and has been replaced with the closest breakpoint, and some unnecessary `@media screen` rules were removed_

---

While removing the `media-breakpoint-*` rules from _mixins, some more light cleaning was done by finding and removing other unused mixins - specifically `ie()`, `out($colour, $width)`, `table-sticky()`, and `after-border()`.
I also moved the non-mixins calculated classes of `w(-{$bp})-{$size}` and `h(-{$bp})-{$size}` to a unified spot in `_utils`, where I think they are more suitably placed.

### DIFF
--- a/src/app/components/pages/Equality.tsx
+++ b/src/app/components/pages/Equality.tsx
@@ -1,17 +1,16 @@
 import React, {ChangeEvent, lazy, Suspense, useLayoutEffect, useRef, useState} from "react";
-import {Button, Col, Container, Input, Label, Row} from "reactstrap";
+import {Col, Container, Input, Label, Row} from "reactstrap";
 import queryString from "query-string";
 import {ifKeyIsEnter, isStaff, siteSpecific, jsonHelper, useModalWithScroll} from "../../services";
 import katex from "katex";
 import {TitleAndBreadcrumb} from "../elements/TitleAndBreadcrumb";
 import {useLocation} from "react-router";
 import {Inequality, WidgetSpec} from 'inequality';
-import {openActiveModal, selectors, useAppDispatch, useAppSelector, useGetSegueEnvironmentQuery} from "../../state";
+import {selectors, useAppSelector, useGetSegueEnvironmentQuery} from "../../state";
 import {EditorMode, LogicSyntax} from "../elements/modals/inequality/constants";
 import { Loading } from "../handlers/IsaacSpinner";
 import { GeneralFormulaDTO, InequalityState, SymbolicTextInput } from "../elements/inputs/SymbolicTextInput";
 import { initialiseInequality, sanitiseInequalityState } from "../../services/inequalityUtils";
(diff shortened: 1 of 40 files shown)


src/scss/cs/logo.scss
```diff
@@ -2,14 +2,14 @@
 .header-logo {
   float: left;
 
-  @include respond-above(nav) {
+  @include respond-above(lg) {
```
jacbn: We didn't change these?

src/app/components/pages/Equality.tsx
```diff
@@ -151,6 +154,9 @@ const Equality = () => {
                 </>}
             </Col>
         </Row>}
+        <Button onClick={() => dispatch(openActiveModal(loginOrSignUpModal))}>
+            Hello world
+        </Button>
```
jacbn: bin

src/scss/common/inequality-modal.scss
```diff
@@ -41,13 +41,13 @@
     }
 }
 
-@media screen and (orientation: portrait) {
+@media (orientation: portrait) {
```
jacbn, line 44: This _feels_ like it was intended functionality, but printing inequality certainly has never worked... :shrug: how odd!

jacbn (approved):
LGTM! I won't merge for this week owing to the event, but all good for next.

src/scss/common/inequality-modal.scss
sjd210: I'm not sure it is. If it was intended, it's odd that the screen-only rules were only applied in the exact same situations that other `@media` rules were anyway - and as you say, it's not like inequality actually looks any good printing regardless. I don't think much thought went into it.

PR #2306, opened 2026-08-18
Improve consistency of button styling

axlewin:
- Restyle fasttrack buttons such that the secondary action button is legible in dark mode.

- Use `solid`/`keyline` in more places, to simplify some site-specific styling & remove all instances of `outline` buttons on sci. In most cases there is no visual impact; where there is a visual change, the buttons should now follow the convention of main action = `solid`, alternative action = `keyline`.

- Remove CSS for sci `outline` buttons.

### DIFF
--- a/src/app/components/elements/inputs/ConfidenceQuestions.tsx
+++ b/src/app/components/elements/inputs/ConfidenceQuestions.tsx
@@ -159,7 +159,7 @@ export const ConfidenceQuestions = ({state, setState, validationPending, setVali
         {state === "initial" && <div className={"d-flex"}>
             <h4 className={classNames({"text-muted": disabled && isAda})}>{confidenceVariables?.title}</h4>
             <div className="ms-2 mt-n1 not-mobile">
-                <Button outline={isPhy} color="solid" className={"confidence-help"} size="sm"
+                <Button color="solid" className={"confidence-help"} size="sm"
                     onClick={() => dispatch(confidenceInformationModal())}
                 >
                     {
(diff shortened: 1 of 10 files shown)


src/app/components/elements/inputs/ConfidenceQuestions.tsx
```diff
@@ -159,7 +159,7 @@ export const ConfidenceQuestions = ({state, setState, validationPending, setVali
         {state === "initial" && <div className={"d-flex"}>
             <h4 className={classNames({"text-muted": disabled && isAda})}>{confidenceVariables?.title}</h4>
             <div className="ms-2 mt-n1 not-mobile">
-                <Button outline={isPhy} color="solid" className={"confidence-help"} size="sm"
+                <Button color="solid" className={"confidence-help"} size="sm"
```
sjd210, line 162: Leaving a note to say that this does have a slightly strange effect on this button, but confidence questions are unused on Sci and would already need visual fixes if we were to use them anyway - so I'm overlooking it.

sjd210 (approved):
Double checked over each changed button. They all seem sensible, so LGTM 👍 

(If we're not using them on Sci, we should probably be consistent with naming and remove them on Ada too, but that can be done later).

src/app/components/elements/inputs/ConfidenceQuestions.tsx
axlewin: Yes; this button just didn't look like a button at all before, so I figured at least making it more visible was a start, but presumably we'd want to restyle this whole component if it were to be used on sci anyway.

PR #2307, opened 2026-08-18
Create local storage-based state, use for markbook ordering

jacbn:
Creates a `useState` wrapper (very much along the lines of `useHistoryState`) that uses a previous value saved in localStorage, if such a value exists.

Uses this function initially for markbook ordering of groups and assignments, as per content request.


src/app/services/storage.ts
```diff
@@ -0,0 +1,24 @@
+import { useState } from "react";
```
barna-isaac, line 1: I think this file doesn't belong under `state/actions`? I think `state/actions` is for redux action creators, and this is a custom hook that doesn't touch redux at all (`useState` is for the component, not redux). I think this could just live in `src/services`? We already have a custom hook there (`useSessionExpired`). (I realize `history.tsx` also lives under `state/actions`, but I think that's an outlier too.)

src/app/services/progress.ts
```diff
@@ -28,8 +30,8 @@ export function useAssignmentProgressAccessibilitySettings({user}: {user: Regist
     const [colourBlind, setColourBlind] = useState(false);
     const [formatAsPercentage, setFormatAsPercentage] = useState(false);
     const [attemptedOrCorrect, setAttemptedOrCorrect] = useState<"ATTEMPTED" | "CORRECT">("CORRECT");
-    const [assignmentOrder, setAssignmentOrder] = useState<AssignmentOrderSpec>(AssignmentOrder.startDateDescending);
-    const [groupSortOrder, setGroupSortOrder] = useState<GroupSortOrder>(GroupSortOrder.Alphabetical);
+    const [assignmentOrder, setAssignmentOrder] = useLocalStorageState<AssignmentOrderSpec>(KEY.ASSIGNMENT_ORDER, AssignmentOrder.startDateDescending);
```
barna-isaac: I'm not sure what uses this! Did you intend to also persist the state of the dropdown on the screenshot? If so, that's unfortunately stored in `AssignmentProgressGroup.tsx:99`.

Image

barna-isaac (changes requested):
Tell me whether you agree the file would make more sense elsewhere, but this is a very elegant, react-y implementation, which will make it easier to persist more of these settings.

However, it might be that I misread what you were trying to do, but I think you meant to persist the order of assignments, and this still doesn't work.

src/app/services/progress.ts
jacbn: I saw a related `useState` and copied it without really trying if it worked :sob: It looks like that page is _supposed_ to inherit from the context but doesn't, so I've made it use that instead.

src/app/services/storage.ts
jacbn: Yes, all good points. Have moved this and `history` to services – but `services/history.ts` already existed, so have merged those.

barna-isaac (approved):
Looks good! In the meantime, @jsharkey13 has pointed out we'll still loose these when a user signs out, but I still think this is an improvement. If we want to persist the sort orders among a user signing in and out, I think we should modify the code part that clears local storage so it leaves these keys alone.

PR #2311, opened 2026-08-19
Remove border radius on sidebar list items (sci)

axlewin:
Prevents dividers between list items (e.g. on `/groups`) from curving at the edges.


jacbn (approved):
LGTM. I've added a few additional related fixes too, one for dark mode (to reduce the brightness of the separators) and one for mobile (to use a normal amount of padding around groups when no group is selected).

PR #2315, opened 2026-08-20
Support u13 RPF SSO flow

jacbn:
- Makes minor changes to the frontend to support the new `EmailVerificationStatus`, `EmailVerificationStatus.AGE_RESTRICTED`. 
- When set to this value, email, password and consent-related inputs are hidden and not required (and default to "no" for any consent).
- LLM questions have new copy for under 13s that prevents answering questions since OpenAI consent cannot be granted.


src/app/components/pages/MyAccount.tsx
sjd210: Annoyingly our `NavItem`s and `TabPane`s are correlated but whether they actually display is separately configured (it would be nice to fix that at some point) - so in this case the Notifications tab is displaying while its contents is not.

Easy solution:

```suggestion
                        {ACCOUNT_TABS.filter(tab => !tab.hidden && !(editingOtherUser && tab.hiddenIfEditingOtherUser) && !(isUnder13(userToUpdate) && tab.tab === ACCOUNT_TAB.emailpreferences).map(({tab, title, titleShort}) =>
```
although it would be _nice_ for it to be baked into `ACCOUNT_TABS` directly, like how `hiddenIfEditingOtherUser` is. That seems overkill for a single case, but maybe we could make `hidden` into a function rather than boolean, and then also remove `hiddenIfEditingOtherUser` as an extra field? 
I'd be happy enough with the easy solution for now, but if you've got ideas about expanding it, that'd be great.

src/app/components/elements/panels/UserProfile.tsx
sjd210, line 181: For an under-13 user, changing Date Of Birth can now trigger an irreversible account change in terms of email verification and the checks we now rely on around it.

Keeping it irreversible is probably sensible, as once the account has left `AGE_RESTRICTED` it now might have a valid email associated and most of our the visible changes are assuming that this is not the case - but we should probably have some kind of warning text for if the user is changing their DoB to a >=13y/o one so that it's not easy to make this transition without knowing that there will be other consequences.

sjd210 (changes requested):
Mostly looks good - nothing very major in terms of requested changes.

I also notice that all these PRs are drafts. Is that for the sake of making sure that everything is merged together properly (e.g. ensuring its done with the 11-12 signup flow) or are we waiting for a certain date to enable it? Just want to make sure I'm in the loop.

src/app/components/elements/panels/UserProfile.tsx
jacbn: We now can't do this (see other PR), so resolving for now.

sjd210 (approved):
Much cleaner in terms of how hidden tabs are handled for My Account, and I'm generally happier with the approach of making `AGE_RESTRICTED` accounts stay that way (or at least requiring an explicit route to upgrade them, which I think we could introduce eventually but is good for now without)

PR #2319, opened 2026-08-20
Use consistent heading element nesting for accessibility 

sjd210:
We have historically used heading elements `<h1>`-`<h6>` for styling purposes, however these elements are also important for navigation, especially for screenreaders.

According to the [WCAG standards](https://www.w3.org/WAI/tutorials/page-structure/headings/), in order to make this navigation work properly, headings should start with a single `<h1>` (after any navigation headings if they exist, which they don't for us), with sequentially increasing numbers for each nested subsection, and equal numbers for sections of the same rank/importance.

Most pages on both sites have violated this, so this requires quite a lot of changes to resolve, however all these changes are very straightforward. Each current `<h*>` element essentially falls into one of the following camps:
- Already fits into the hierarchy (e.g. `<PageTitle>` as our single `<h1>`) 
  - _no change_
- Should be a heading in the hierarchy, but is of the wrong rank (e.g. section titles on the My Account page)
  - `<h3>` -> `<h2 className="h3">`
- Should not be a heading in the hierarchy at all
  - `<h3>` -> `<div className="h3">`

The only exceptions to this are:
- Modals, for which there is [no consensus in WCAG or the HTML standard](https://medium.com/@web-accessibility-education/which-heading-level-should-dialog-modals-have-7b3df89437f0) for whether they should start with `<h1>` (since a modal represents its own document) or `<h2>` (since it is on the same wider page as a `<h1>` already). I have chosen to follow [Bootstrap's example](https://getbootstrap.com/docs/5.3/components/modal/#modal-components) and set the modal-title to `<h1>` and label accordingly from there, but either would do
- The `question-validation-response`, which was barely using `<h1>` for its styling anyway so a new `response-heading` class has been added in its place
- Some headings in the registration flow or My Account page that were using different elements per site. They now use the same element for navigation, but different `classNames` for styling. e.g. `<h2 className={siteSpecific("h4", "h3")}>`
- Headings tags have been wrapped around accordion header buttons. This is the [recommended structure](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).

---

(If there are any violations left, or for any that are generated in the future, they should now be individual cases that can be quickly fixed rather than a fundamental disregard for the heading structure. I checked manually for non-trivial cases, but can't guarantee this caught everything - especially for elements that _should_ be marked as headings but did not already use a `<h*>` tag).


ioiototm (commented):
I was testing the markdown at /markdown and [this worked well](https://github.com/isaacphysics/isaac-react-app/pull/2319/changes/846f1df372f738435164084b9409f67cf9c7a9e7) but when you have a # symbol, even if it's a code block, it gets replaced to a div 

````md
# Testing the heading

```python
test = sum (10,10)
# Shouldn't be heading
```
<h5>Raw HTML </h5>

This is a different Title
====

> # Quoted Heading

````

This code would erroneously put a div where the # in the python code block is. 

A solution Claude suggested was this:

```js

MARKDOWN_RENDERER.renderer.rules.heading_open = function(tokens: Remarkable.HeadingOpenToken[], idx: number/* options, env */) {
    return `<div class="h${tokens[idx].hLevel}">`;
}

MARKDOWN_RENDERER.renderer.rules.heading_close = function(tokens: Remarkable.HeadingCloseToken[], idx: number/* options, env */) {
    return `</div>\n`;
}

```

and removing the 6 rules in the regex. They do work if you type it without a space, but then with a space it thinks it's a heading and replaces it.

That code also does Setext and blockquotes the same way.

sjd210:
@ioiototm First of all: in my defence, I hadn't got around to properly testing it yet - this PR isn't ready to review!

Anyway I did suspect that the repeated #, ##, etc rules weren't the best way to implement this, but couldn't figure out a better way with regex specifically - and wanted at least something in place while I tested navigating with a screenreader. I'm glad to see there is something better :)

This uses the same kind of structure as our link replacement rule, so it seems sensible - but I'll wait to get around to that proper testing phase before I can confirm that and commit it in.

sjd210:
N.B. I used the word "header" without thinking while working on the first half of this PR, so it's in the branch name/many of the commits but shouldn't be. Headings are titles to specific sections of the page. Headers are navigation blocks at the top of the page. They are not the same.

src/app/components/elements/Accordion.tsx
```diff
@@ -164,74 +164,78 @@ export const Accordion = ({id, trustedTitle, index, children, startOpen, deEmpha
...
-                classNames("stage-label d-flex align-self-stretch align-items-center p-2 justify-content-center ", audienceStyle(audienceString))
-            }>
-                {(above["sm"](deviceSize) 
-                    ? audienceString 
-                    : audienceString.replaceAll(",", "\n")
-                ).split("\n").map(
-                    (line, i, arr) => <>
-                        {line}{i < arr.length && <br/>}
-                    </>
+        <h3 className="mb-0">
```
jacbn, line 167: There is some CSS in `phy/accordions.scss` with:
```scss
.isaac-accordion {
  ...
  &:has(> .accordion-header.active) { 
    border-color: var(--nav-primary);
    ... 
  }
}
```
that is used to colour the accordion in the subject colour when active. This doesn't apply any more with this structure change!

jacbn: Presumably, also, everything inside this h3 is read out as the section name in the screenreader. Does this create situations where unnecessary detail is in the heading? Should the _entire_ button be the h3, or just the part title?

jacbn (commented):
The code looks all okay to me besides the one point below about accordions. It will be tricky to visually test every page to make sure the headings are consistent, but very few VRTs have changed and the ones that have look good so I say we go with it and fix anything that comes up later.

jacbn:
I've fixed the border colour bug, but haven't made any changes to the h3 structure – I will leave this for later as I'd appreciate your opinions on this too.

src/app/components/elements/Accordion.tsx
sjd210: The heading tag wrapping the entire accordion heading button is the recommended pattern by w3: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/, and this is required for the structure to work properly with an interactable element (e.g. NVDA doesn't recognise the headings if the h3 is just on the inner title, because of it being inside the button), but I agree that this creates unnecessary details being included as part of the heading.

These extra elements have relevant information and should still be readable by screenreaders, so we shouldn't just `aria-hidden` them - but this makes them hard to remove from the heading. There probably is a way around this, but not a straightforward one - I'll do some more thinking.

(The accordion border colour was definitely an issue though. Thanks for spotting it!)

PR #2321, opened 2026-08-20
Remove unused temp gameboard code

ioiototm:
The problem was that there was a button on Ada where you can generate a quiz based on a topic.

image

Clicking it would go to the quiz builder, but an error message would pop out on the side about a GET request, since the endpoint was removed. As this was already removed from the API, I removed the code from the frontend as well, the button and also the endpoint. 

image


jsharkey13:
@ioiototm can you fix the unused variables ESLint issues? Otherwise looks good to me (though I did help write it, I grant you 😅).

jsharkey13 (approved):

PR #2322, opened 2026-08-21
Build Cyber Explorers transition pages

jacbn:
Adds:
- A landing page for the CyEx => Ada transition;
- A new section in the topics page
  - (todo: replace tab picker's `useHistoryState` with `useLocalStorageState` once that's merged)
- A new slice on the homepage
- UK Govt & TF logos on the footer

TODO:
- UK Govt & TF logos on homepage – *where* is still being decided on by content;
- Replace slogan on CyEx landing page with text from DSIT (to come)
- Fix "Read more" link on Cyber Explorers block on the new landing page


barna-isaac:
I'll start collecting review findings here. First, on the "TechFirst" landing page, the following two links don't work.
- https://adacomputerscience.org/pages/202609_techfirst
- https://staging.adacomputerscience.org/pages/202609_techfirst
I realise these are content-controlled, but shouldn't we at least put some placeholder in there so it's not a 404? Do the content teams know they need to create these pages?

barna-isaac:
I might be overstepping here, but "In collaboration with Funded by UK Government" is not a coherent sentence. If we have to use the "Funded by UK Government" asset, could we not just remove "In collaboration with"? Or swap the two logos so it becomes "In collaboration with TechFirst Funded by UK Government"?

Screenshot 2026-09-15 at 11 19 58

jacbn:
I would love to, but `In collaboration with` was requested specifically by the Ada team. We also cannot swap the order of the logos as this is laid out specifically in TechFirst's own brand guidelines.

jacbn:
> the following two links don't work

Indeed, it is on content to publish pages at those links. They should be ready by tomorrow.

btw, can you add review comments as an actual review? I can't reply to comments.

src/app/components/pages/AllTopics.tsx
```diff
@@ -112,5 +143,43 @@ export const AllTopics = () => {
...
+            onActiveTabChange={(n) => {
+                setStageTab(TOPICS_STAGES[n-1]);
+            }}
+            renderHiddenTabs={false}
+        >
+            {{
+                ["11 to 14 years"]: <TopicsListing tagCols={partition(ks4Tags, s => !s.comingSoonDate || ks4Tags.indexOf(s) <= 4)} age="11-14" />,
+                ["14 to 19 years"]: <TopicsListing tagCols={partition(coreAdvancedTags, s => s.title.charAt(0) <= "D")} age="14-19" />
+            }}
+        </Tabs>;
```
barna-isaac: I think this is the line whose semicolon renders at the bottom of the AllTopics page, although I might be wrong. The page definitely does show a semicolon at the bottom that it shouldn't be.

src/app/components/pages/AdaCyberExplorersResources.tsx
```diff
@@ -0,0 +1,173 @@
...
+                    </ImageBlock>
+                    <TextBlock>
+                        <h2>What is Ada Computer Science?</h2>
+                        <h3 className="font-size-1-25 mb-3">One place to learn computing, from ages 11 to 19.</h3>
+                        <ul>
+                            <li>Free computing resources, from practical activities for ages 11 to 14 to structured computer science learning for ages 14 to 19</li>
+                            <li>Classroom tools to group students into classes, assign activities, and track learners&apos; progress</li>
+                            <li>Expert content developed by the Raspberry Pi Foundation in collaboration with partners including the University of Cambridge</li>
+                        </ul>
+                        <div className="d-flex gap-3">
```
barna-isaac: I think rather than breaking into three lines, the buttons should just each take up a full line? This is what similar button pairs do on the home page and on the student and teacher landing pages.
Image

public/assets/cs/decor/teacher-5.png
barna-isaac, line 1: I know these are just given to us, but the `teacher-5` image is even lower-res than the resources we were given for the student and teacher landing pages. This is all the worse because these images are shown really large on mid-sized screens (and tablets also tend to bi high-res displays).

Here's the "alpha" decoration on the new image
Image
And here's the same decoration on the teacher landing page.
Image

src/app/components/pages/AdaCyberExplorersResources.tsx
```diff
@@ -0,0 +1,173 @@
...
+                            <Button href="/">Discover more of Ada CS</Button>
+                            <Button outline className="bg-white" href="/register">Create an account</Button>
```
barna-isaac: Also shown for logged-in users, which doesn't make a lot of sense? The teacher and student landing pages usually hide these or show a different link if you're logged in.

barna-isaac (approved):
I've fixed the stray `;`, and I didn't find anything major. I've left a few comments, but these are all things we can fix later.

public/assets/cs/decor/teacher-5.png
jacbn: Not much we can do about this unless we can get higher-resolution ones from the design team. I remember going down the rabbit hole trying to find the base image before and noone could find them.

barna-isaac:
Thanks for the fixes, looks great! Merging now.

PR #2329, opened 2026-08-24
Update our dynamic mentoring block with 26-27 group info

jacbn:
I have also done some content work in splitting the `_current` page up into the 3 parts that content have now requested.


barna-isaac (approved):
Looks great, thanks.

PR #2332, opened 2026-08-25
Fix `useHistoryState` race condition

jacbn:
The main fix here is the addition of `flushSync: true` inside the `navigate` call. 

Consider a scenario with two distinct values being managed by `useHistoryState` that update consecutively. In an ideal world, an initial location-state (i.e. the collection of all states at a given `history.location`) S1 is modified with an additional key from the first call, creating S2, which is then modified again with a second new key, creating S3. S3 is, of course, intended to have both new keys.

However, if a state change occurs to both states simultaneously*, since `navigate` is an asynchronous function it is possible for both `navigate` calls to start with a location-state reading of S1. This means S3 will not include the key created in S2, meaning not all state changes are recorded in the history.

> *In the issue that raised this, the second state was updated in a `useEffect` with a dependency on the first.

The `flushSync` option inside `navigate` forces an immediate DOM update after the navigation. This prevents batching of state updates, meaning the location is updated after **each state update**, preventing a situation where two calls can read the same stale state.

---

The remaining changes in the PR clean up the equality checking between two objects, ensuring that:
- If a state-changing function is passed in to `setHistoryState`, the computed value of this is used rather than the function itself;
- `isEqual` is preferred over `===`, providing better object equality checking.

### DIFF
--- a/src/app/services/history.ts
+++ b/src/app/services/history.ts
@@ -1,6 +1,7 @@
 import type {TypeGuard} from "@reduxjs/toolkit/dist/tsHelpers";
 import {useCallback, useEffect, useRef, useState} from "react";
 import {useLocation, useNavigate} from "react-router-dom";
+import isEqual from "lodash/isEqual";
 
 function prepareHash<T>(defaultState: T, typeGuard: TypeGuard<T>, hash: string) {
     const hashText = hash.replace("#", "");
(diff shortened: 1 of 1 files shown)


sjd210 (approved):
This seems to work and I'm happy for it to go in, but I think we should take heed of React's caveats about `flushSync`:
> - flushSync can significantly hurt performance. Use sparingly.
> - flushSync may force pending Suspense boundaries to show their fallback state.
> - flushSync may run pending Effects and synchronously apply any updates they contain before returning.
> - flushSync may flush updates outside the callback when necessary to flush the updates inside the callback. For example, if there are pending updates from a click, React may flush those before flushing the updates inside the callback.

Since we're only using this on `navigate`, the potential for major consequences seems low (e.g. interrupting pending `Suspense` typically doesn't matter if you're leaving the page anyway), it's just probably something to pay attention to if we do notice performance problems any time soon.

PR #2333, opened 2026-08-25
Ada 11-14 navigation: code suggestion

jacbn:


barna-isaac (approved):
That's better, I hated we needed to hardcode the list of pages twice.

The fact that we now get the navigation even without the query parameter is different to how navigation works on any other concept page, but I don't think this matters too much. It's also still possible to embed these concept pages in a topic, and the pages would still inherit the navigiation from the topic.

PR #2334, opened 2026-08-25
Add checkbox for email when a student finishes test

ioiototm:
Only Isaac for now, but it works just as well on Ada - maybe we want to discuss whether Ada also could have a use for it

The text for the checkbox can be changed if current text is not good or descriptive enough


jacbn (approved):
LGTM. I wondered whether it may be the case that we want to modify this setting on a quiz that already exists. The `updateQuiz` endpoint in the backend would need to allow this and we'd need a way to do this in the frontend; we think though that, while a good idea, this can be done separately as an extension. We've made a card for now, and happy to merge this in the meantime.

PR #2338, opened 2026-09-01
Fix bad save deck state caching on detail pages

jacbn:
Uses correct `rtkq` caching tags in book detail and revision detail pages. Paired with the API etag PR, this fixes issues with not appearing to save / unsave decks on these pages.

### DIFF
--- a/src/app/components/pages/SubjectOverviewPage.tsx
+++ b/src/app/components/pages/SubjectOverviewPage.tsx
@@ -7,8 +7,9 @@ import { PageContextState } from "../../../IsaacAppTypes";
 import { convertToALVIGameboard, CustomListViewItemProps, ListView, ListViewCards, transformItemsForCustomListView } from "../elements/list-groups/ListView";
 import { LandingPageFooter } from "./SubjectLandingPage";
 import { DifficultyIcon } from "../elements/svg/DifficultyIcons";
-import { GameboardDTO } from "../../../IsaacApiTypes";
 import { PageMetadata } from "../elements/PageMetadata";
+import { useGetGameboardByIdQuery } from "../../state";
+import { ShowLoadingQuery } from "../handlers/ShowLoadingQuery";
 
 const SubjectCards = ({context}: { context: PageContextState }) => {
     const deviceSize = useDeviceSize();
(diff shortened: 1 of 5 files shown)


sjd210 (approved):
I do need to read up more on RTKQ, but this much makes sense to me and definitely seems to work.

I noticed that subject overview pages were also not showing saved status. This was a slightly different issue (us not loading the gameboards at all), but for the sake of practice I've fixed it myself and committed that. I'd appreciate you having a look in counter-review.

src/app/components/pages/SubjectOverviewPage.tsx
```diff
@@ -76,26 +77,21 @@ const SubjectCards = ({context}: { context: PageContextState }) => {
...
+        maths: "sample_maths_questions",
+        physics: "sample_phy_questions",
+        chemistry: "sample_chem_questions",
+        biology: "sample_bio_questions"
     };
 
-    return items[subject].length > 0 ? <ListView className={className} type="gameboard" items={items[subject].map(convertToALVIGameboard)} /> : null;
+    const gameboardByIdQuery = useGetGameboardByIdQuery(gameboard_ids[subject]);
+
+    return gameboard_ids[subject]
```
jacbn: This works great! `maintainOnRefetch` is proving its value :) probably my favourite addition to the codebase...

The only thing I'd say is that it feels odd to have a query that relies on `gameboard_ids[subject]` being defined, then _afterwards_ have a conditional where we don't render contents if it is undefined. The query is still being executed and **can still error** prior.

There are really 3 solutions here, really up to you which you prefer:

1. Have faith in TS type-checking; `subject` _must_ be a valid key of `gameboard_ids` by its type definition. You don't really need the conditional here.

2. If we are going to deal with the case where `gameboard_ids[subject]` can be undefined, we'll want to make the query parameter deal with this too so it can't error prior; `skipToken`s are our most common way.

```tsx
    // skipToken needs importing
    const gameboardByIdQuery = useGetGameboardByIdQuery(gameboard_ids[subject] ?? skipToken);

    return gameboard_ids[subject] ? ...
```

3. As an extension to 2, if you'd rather not have two conditionals with the same condition, you can make more extensive use of the tools inside `ShowLoadingQuery` with `uninitializedPlaceholder`; note though that there isn't a way to automatically reuse the error message, mainly because it being uninitialised _isn't always an error state_.

```tsx
    const gameboardByIdQuery = useGetGameboardByIdQuery(gameboard_ids[subject] ?? skipToken);

    return <ShowLoadingQuery
        ...
        uninitializedPlaceholder={<>...</>}
    />;
```

jacbn (commented):
The query is great, just a little point about after it!

src/app/components/pages/SubjectOverviewPage.tsx
sjd210: Hrmm well looking in comparison to the book/revision detail pages, they fail a lot more gracefully by just not showing the gameboards section at all if they aren't loaded. For those pages and this one, the page as a whole is still functional without that section (unlike e.g. My Assignments where the page is fully broken so there really should be an error), so this feels sensible to me rather than alarming the user with a warning they don't _really_ need to worry about.

It seems like a combination of `uninitializedPlaceholder={RenderNothing}` and adding `ifNotFound={RenderNothing}` handles this case best, so I'll go with that

jacbn:
LGTM

PR #2340, opened 2026-09-01
Use consistent coloured alt text on white image backgrounds

sjd210:
Image backgrounds are always hardcoded to white, but text colour (for alt text when the image is missing) is dependent on the page context. This could make it also be white in e.g. in a `validation-response-panel`/dark mode. It is now hardcoded to black so it is always readable.

Some cleanup while I'm working here anyway: There were two `figure` blocks in this scss file. I combined these into one, removed the unnecessary `!important`s, and removed the unused Nomensa `img-thumbnail` class.


jacbn (approved):
LGTM.

I did some investigation on using lightness themes on specific components, e.g. figures, to force light mode and stop us worrying about places that need to be hard-coded and others that don't. This would fix the issue we were seeing [here](https://github.com/isaacphysics/isaac-react-app/pull/2159#discussion_r3797142707) too. Unfortunately, there's a fairly big problem that I've written up into a card in tech queue – I'm not really expecting us to do this, but worth keeping the investigation notes. To keep a long story short hard-coding the odd thing like this is far simpler than any solution we might otherwise come by, so happy to proceed here.

PR #2342, opened 2026-09-01
Add box alignment to front end drop box

ioiototm:
Boxes are aligned based on the selection in the content. Defaults to left for backwards compatibility.


sjd210 (approved):
Also tested with edge cases of drop-zones of `left: 0` and `left: 100`, and with very small drop-zones, which seem to be behaving sensibly 👍

PR #2350, opened 2026-09-04
Inline feedback improvements

jacbn:
Two fixes to inline feedback:
- Fixes an issue where two competing `::after`s were mixing into one, preventing the feedback marker appearing on numeric unit buttons;
- Brings back the feedback box for fully-correct inline questions only if they contain at least one part with correct answer feedback; auto-jump to the first of these


src/app/components/content/IsaacInlineRegion.tsx
sjd210: This `useEffect` is needed for the inline feedback state to display correctly on loading the page. It's mostly duplicated logic from the below one used for when the question is submitted, so we can probably clean that duplication up while adding the `firstCorrectPartWithFeedback` state here.

sjd210 (changes requested):

jacbn:
I don't love that the "just answered" state is not the same as the "coming back to a question" state – unanswered boxes are yellow in the former, black in the latter – but I *think* the issue there is that you can't tell whether something has been submitted there or not when you first come to a question page, so not much can be done.

sjd210 (approved):
I think coming back to a submitted question with unanswered boxes is a niche enough case that I'm not worried about a mild visual difference. 

LGTM 👍

PR #2352, opened 2026-09-07
Improve horizontal layouts for pods

jacbn:
Makes a number of improvements to pods when in horizontal layout.
- Will no longer fill all available space, preventing scaling issue with event pods on subject landing page
- Aligns news and event pods' horizontal layouts much more closely
- Simplifies construction of subject landing page content; no more needlessly nested row/col layouts


src/scss/phy/cards.scss
sjd210: These `event-pod-hex`s are aligned with the top of the pod, not with the top of the image; this looks odd. This was fine before, but not with the padding of a `pod-flush`.
They align just fine with the top of the image by default due to their placement in the DOM, so we can just remove this line and restore them to the same visually as before.


```suggestion
```

sjd210 (changes requested):

src/scss/phy/cards.scss
jacbn: I'm actually more tempted to give the image's container `<a />` `position: relative` to achieve this same effect. We can still remove the `top: 0` too since after this change it doesn't do anything, but the main additional purpose is to make the `left: 7%` reference the width of the image rather than the width of the pod as a whole. Without this, the hex appears almost halfway into the image on flush pods.

PR #2354, opened 2026-09-07
VRT issues: fix circular deps, non-enqueued `cy`

jacbn:
I have run this 10 times on the most recent commit – it failed once in `News.cy.tsx`, but I can't see any reason for this and also can't reproduce it. It would be helpful to know if this kind of failure is specific to News (in which case it is probably a problem with our code), or can happen on any VRT arbitrarily (likely a problem with Cypress), but without more reproductions I can't determine which. This is, however, significantly more consistent than it is on main, so I'm happy to call this an improvement and merge it.


barna-isaac (approved):
At first I was a little scared seeing the changes to the live login page, but I now understand it's just that the service and the inputs are now extracted to a separate file.

I've resolved the merge conflict (I updated the extracted inputs with an accessibility change that got applied to `LogIn` in the meantime), and think this would be good to merge now, although I'll hold on merging until the release is done.

The `.then()` chaning is a little unintuitive for me, and I'm not sure it actually makes a difference, as the `if` always evaluates to the same value. However, they're also now on main, and they can't hurt too much, so I'm inclined 
to leave them (and maybe later we can see if these can be removed).

PR #2355, opened 2026-09-09
fix bug where ada sidebar didn't reach page bottom

barna-isaac:
Fixes a bug where, for small content sizes, the ada sidebar didn't extend all the way to the bottom of the content container.

The sidebar's parent (the main content area) is sized using flexbox. This means a) it's sized so it fits
the content, and b) it's sized to take up any excess space not taken up by the header or footer. 
In order for the sidebar to run along the main content, this means we cannot just set on it `height: 100%`,
as the calculation would be recursive. I also didn't have the courage to set `display: flex` or
`position: relative` on the main content, which would have been a requirement for anything less hacky.

Instead, I used a workaround on the sidebar that's strictly local. The white background overflows a container 
whose height is 0. Elements are painted in DOM-order, which is just right for us: the white
background covers up the grey from main, and the buttons and footer cover up the white background.
The height here is arbitrary, but should be safe. The width of the background is defined explicitly, which
was easy to do since the sidebar's width is also explicitly defined.
        
I also extracted the size of the collapsed and expanded sidebar to sass variables.

### DIFF
--- a/src/scss/cs/sidebar.scss
+++ b/src/scss/cs/sidebar.scss
@@ -1,10 +1,7 @@
-.sidebar {
-    background-color: $cs-white;
-}
+$expanded_width: 220px;
+$collapsed_width: 69px; // 2*20px padding + 24px icon + 5px border
 
 :not(.offcanvas-body) > .sidebar {
-    background-color: $cs-white;
-
     @include not-reduced-motion {
         transition: all 0.3s ease-out;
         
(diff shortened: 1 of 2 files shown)


src/scss/cs/sidebar.scss
```diff
@@ -1,43 +1,77 @@
...
+        //
+        // In order for the sidebar to run along the main content, this means we cannot just set it to `height: 100%`,
+        // as the calculation would be recursive. I also didn't have the courage to set `display: flex` or
+        // `position: relative` on the main content, which would have been requirement for anything less hacky.
+        // 
+        // Instead, I used a workaround on the sidebar that's strictly local. The white background overflows a container 
+        // whose height is 0. Elements are painted in DOM-order, which is just right for us: the white
+        // background covers up the grey from main, and the buttons and footer cover up the white background.
+        // The height here is arbitrary, but should be safe. The width of the background is defined explicitly, which
+        // was easy to do since the sidebar's width is also explicitly defined.
```
sjd210: The solution is good, but this comment is far too long! You can leave a link to the PR or card with this in-depth explanation in it, but ideally in the code itself someone should be able to skim through and understand what its doing rather than needing to read multiple paragraphs of text for context first.

I think the only relevant details here that someone might need while changing code around this are a) this is a hacky alternative to setting `display: flex` on the main content, b) the DOM-order, and c) the height being arbitrary. Everything else is useful context for a reviewer, but not the right place to keep it long-term.

sjd210 (changes requested):

sjd210 (approved):

PR #2357, opened 2026-09-10
Fix print layout and pagination across browsers

ioiototm:
Printing on Safari on a bigger screen was broken, as it would print the content in a thin line:

image

The main fix was adding padding and margin to the page.

image

After fixing that, I realised there was a lot of empty space in some questions, depending on how you print it - fixed by making  the accordion body a block and removed the page break inside, so it prefers to be printed continuously, made the questions also preferably continuous, and removed the `@page` calculation altogether, as that made pages look too big (as per the live version of the screenshot).

I tested it mainly on Isaac, and a bit on Ada afterwards.


jacbn (approved):
I anticipate we will receive a number of complaints about this from our teams and some teachers about how printing has changed, so we should perhaps be prepared for changing this again in the future. But we won't know until we try, and this does improve Safari's printing, so cautiously happy to push for now.

PR #2359, opened 2026-09-11
Use correct submission method for fasttrack inlines

jacbn:
Inline questions should be submitted via `submitInlineRegion`; FastTrack questions were using a different submit button that did not have the `onClick` override that used to be in place for inline regions on the main button. This PR refactors the flow a little so all question attempts pass through the same `onSubmit` handler, ensuring both regular and FT inline questions submit using the same logic.


barna-isaac (approved):
Looks good! Also changes the focused element after submission for an inline question that's not fast-track, but I don't see that being a bad change (pressing Tab after submit previously landed on the flag, it now lands on the feedback navigation).

PR #2365, opened 2026-09-14
Adjust Poppins font ascent/descent overrides

jacbn:


barna-isaac:
will update the font instead (https://github.com/isaacphysics/isaac-react-app/pull/2370/changes), which seems to eg. the login page cutting off font descenders (without intoducing the issues we see here, eg. text becoming too high on tags)

PR #2368, opened 2026-09-14
Hotfix/ada descenders fix v2

barna-isaac:
This upgrades Ada's bundled `poppins` font to to v24, which I've downloaded from
https://gwfh.mranftl.com/fonts/poppins?subsets=latin,latin-ext, and removes a workaround that attempted to fix problems with the font version we previously used.

I'm doing this to resolve
- issues where descenders were cut off on the log-in page, in Chrome, a side-effect of `poppins-fixes`, illustrated:
Screenshot 2026-09-14 at 17 11 52

- issues where text is was not vertically centered in Safari, because the directives used in `poppins-fixes` were never supported there, illustrated:
Screenshot 2026-09-14 at 17 13 24.

## The changes
The vast majority of this commit is VRT updates. Besides these, the PR contains:
- in `/public/assets/cs/fonts`, 8 updated version of the bundled `.woff2` poppins fonts. I didn't update the `.woff` versions (didn't know how to get new versions for these), but these are only used as fall-back by old browsers anyways.
- an update to `/src/scss/cs/fonts.csv`, which reverts `poppins-fixes`. These were intended to fix problems with the previous font version, which caused text to appear out of alignment on input boxes (among others). @jacbn has today noticed these fixes were also responsible for text descenders getting cut off. Since the fix didn't work in Safari, the second screenshot also illustrates what Chrome would have looked like without these. With the updated version, Chrome no longer needs these fixes (also eliminating the descenders bug), and Safari also now gets vertical alignment on inputs.

## Learnings from VRTs
It's a new font, so there are slight layout changes all over the site. I think these are mostly improvements, and for the few regressions, I think it's worth accepting these for the gain that we're no longer trying to write our css against a font that's badly spaced. Having gone through the VRTs, here are the changes I noticed
- headers get pushed down by a few pixels
- text now correctly center-aligns vertically on buttons (left: before, right: after)
Screenshot 2026-09-14 at 17 36 31

- there are a few examples where a checkbox and its label are better aligned. Eg, on the "My Progress" page
Screenshot 2026-09-14 at 17 42 09.
On a few pages, the new alignment feels worse.
Screenshot 2026-09-14 at 17 43 19.
Crucially, there's no change to the alignment of text and its label on multiple choice questions.
- The font's `line-height: normal` value has changed from 1.2 to 1.5. For an example, see the "Remember me" text on the Log In dialog. 1.5 is the new value recommended by accessibility standards, although for a checkbox and its multi-line label, this does mean the label is now wider than the checkbox (eg: the log-in page, on a narrow screen)
  Screenshot 2026-09-14 at 17 45 57


ioiototm:
This looks good, I think it makes sense to update the fonts, rather than fighting them with css. I checked [this](https://caniuse.com/?search=woff2) for support of the .woff2 file, and from what I see, basically almost everything supports it, only super old (8+ years versions) don't support it (also, if your browser doesn't support woff2, then I would assume there are bigger problems on the website than some fonts). I also checked Plausible, as it has browser versions of users, and from all I could see, we support basically everything out users use.

Not sure if it's worth thinking about removing some of the old file type fonts, since almost everything supports .woff2, and if by some reason the user is using a super old browser, I think in that case it should just load a system font in its place (if no .ttf is found) - if that is a thing we're willing to accept.

There are also converters of woff2 to woff online, not sure how useful or helpful that would be if we want that.

The other minor thing about the 1.5 line-height, I read the [WCAG](https://www.w3.org/WAI/WCAG22/Understanding/text-spacing.html#:~:text=Author%20Responsibility) bit about text spacing, and it sounds like it's more of a "the user must be able to set the content to that" and not for us to set it necessarily. So as long as we don't hinder the user from doing that, 1.2 is fine, although if it's already 1.5 I think that is also fine.

ioiototm (approved):

barna-isaac:
@jacbn and I have decided to do a one-off fix for just the Log-In page (+ log in modal) only, so the "Remember me" text always shows as a single line.

jacbn (approved):
I'm mostly happy with this. I believe the new font file still has some issues with certain browsers, notably with the line height – this image is with `line-height: normal` on Chrome, and the text should definitely *not* be overflowing the generated text height:

Image

Different browsers generate different heights for the text and these are quite different from each other. I can see this causing some painful issues in the future.

Image

(safari)

I can't see any now, but so long as you're happy to go with this and deal with whatever issues we might face in the future, I'm okay with it.

barna-isaac:
I think the comment above was motivated by a misunderstanding. 

On Chrome, with the updated font on this branch, I'm seeing
Screenshot 2026-09-23 at 10 55 56

In fact, it's on the live site that I see
Screenshot 2026-09-23 at 10 54 57

sjd210 (approved):
There will inevitably be some slight variation between how font and their bounding boxes are rendered in different browsers/OSs (e.g. the height I'm seeing for the same element on all the browsers I'm trying is 24px rather than 25px), but I agree that I don't think it gets as extreme as the case Jaycie was positing, and that was probably just from the live site.

Otherwise, I agree with others that this seems sensible and looks good. The VRTs are clearly doing their job and didn't find anything, nor could I while poking around the site, and it makes sense to me that we use an up-to-date version of the font that fixes some of the problems we've been having. I don't think this is particularly risky (and worse case we give it a go, and revert it if we need to) LGTM 👍

PR #2372, opened 2026-09-15
Replace lookbehinds for compat with old Safari

jacbn:
An issue as raised by a ticket whereby old versions of Safari do not support lookbehinds in regex. Simply replaces the lookbehinds with equivalents.


src/app/components/elements/inputs/SymbolicTextInput.tsx
```diff
@@ -132,7 +132,7 @@ export const symbolicTextInputValidator = (input: string, editorMode: EditorMode
         }
     }
 
-    const decimalsOutsideCompound = input.match(/(?<![A-Za-z])\d+\.\d+\s*([A-Z]|$)|[A-Za-z]\d+\.\d+\s*$/);
+    const decimalsOutsideCompound = input.match(/(?:[^A-Za-z])\d+\.\d+\s*([A-Z]|$)|[A-Za-z]\d+\.\d+\s*$/);
```
jacbn: The output here is different, but the truthiness is the same – and this is all we use!

src/app/components/elements/inputs/SymbolicTextInput.tsx
```diff
@@ -132,7 +132,7 @@ export const symbolicTextInputValidator = (input: string, editorMode: EditorMode
         }
     }
 
-    const decimalsOutsideCompound = input.match(/(?<![A-Za-z])\d+\.\d+\s*([A-Z]|$)|[A-Za-z]\d+\.\d+\s*$/);
+    const decimalsOutsideCompound = input.match(/(?:[^A-Za-z]|^)\d+\.\d+\s*([A-Z]|$)|[A-Za-z]\d+\.\d+\s*$/);
```
jacbn, line 135: The output here is different, but the truthiness is the same – and this is all we use!

jsharkey13 (approved):

PR #2374, opened 2026-09-16
Auto-reload for users with stale chunks post-release

jacbn:
When we do a release, our Vite bundler creates new versions of different code chunks, each "signed" with a hash in their filename that updates with each change. If a user comes to our site before a release, the compiled JS they receive will use the *old* names. If this remains cached in their browser while we do a release, we will no longer host files under those old names; this means all lazily-loaded files *can* cause a client error given the right circumstances. Indeed, our stats show a small number of these errors across a huge variety of pages.

This PR aims to resolve the issue by causing a reload on the page before displaying the `<ClientError />`. To avoid infinite reload loops we must of course limit this to only errors we are certain a full reload will fix, which these dynamic import failures will as it causes a full refresh, bypassing the browser cache and requesting new versions of the files from us.

We already did something similar for ChunkLoadErrors; this merely expands on this. You can test it by going to a page with a link to a page that would require additional code to be loaded (e.g. most question types), then turning off your dev server; clicking that link will no longer show a client error, but instead cause a reload, probably showing a *browser error* instead as the _new_ request can't connect to the server.


barna-isaac (approved):
Since this just adds a few more error messages to an already established pattern, I approve. I've also checked manually the refresh doesn't accidentally happen outside of these cases, which I think is the worst thing that could go wrong here, and it doesn't.

I think the intuition that this will help most users is correct, although I do think there might have been edge cases where the error being shown was better. For example, the refresh might cause data loss that just showing the error page would not have. I think this would be relevant rarely enough that doing the refresh is still an improvement.

PR #2376, opened 2026-09-17
Fix dragging indentation behaviour for Parson questions

ioiototm:
When doing Parsons questions, there would be a set of steps you could do that made dragging indentation not work.

If you transferred all the cards from left to right with the button on them, and then tried to indent them by dragging, you would need to move the card up and down for it to update and realise it can be indented.

Same would happen if you drag them manually, but put them all at the top one after the other.

This fix basically just checks in `onDragStart` if you are in the answer column or not the first element, and sets the correct variables - the `onDragUpdate` wouldn't fire unless you move the card vertically, which explains why sometimes it wouldn't show the indentation unless you first moved it vertically.


src/app/components/content/IsaacParsonsQuestion.tsx
```diff
@@ -124,13 +124,23 @@ const IsaacParsonsQuestion = ({doc, questionId, readonly}: IsaacQuestionProps<Is
         setDraggedElement(draggedElement);
         setInitialX(choiceElement && choiceElement.getBoundingClientRect().left);
         setCurrentIndent(draggedElement?.className.match(/indent-([0-3])/g)?.map((match) => parseInt(match.split('-')[1]))?.[0] || 0);
+        if (initial.source.droppableId === 'answerItems' && initial.source.index > 0) {
+            setCurrentDestinationIndex(initial.source.index);
```
jacbn, line 128: These lines are the only changes that fix the bug; the rest are redundant and only end up complicating the variable flow, so I would rather get rid of them. You also don't need the `initial.source.index > 0` check, because if it is 0, the existing `previousItem` calculation in `onMouseMove` will be undefined and you'll get the behaviour you want anyway (`currentMaxIndent` set to 0). You _do_, though, need a `canIndent` check so that we aren't setting this for non-indentable questions.

In all, this is everything you need (after `setCurrentIndent(...)` in `onDragStart`):
```tsx
if (canIndent && initial.source.droppableId === 'answerItems') {
    setCurrentDestinationIndex(initial.source.index);
}
```

The reason you don't need the other lines is that:
- The `setCurrentMaxIndent` you had below this (L129-130) is still running in `onMouseMove` – this is just repeated logic;
- And while both the changes to L132-133 and L142-143 do reset the variable values between multiple drags and so there is an argument to keeping them in, they would be reset again anyway on the next `onDragStart` / `onMouseMove` so again these don't end up affecting anything.

jacbn (changes requested):
This does work, but I think a lot of it is redundant for what it achieves!

src/app/components/content/IsaacParsonsQuestion.tsx
ioiototm: It is simpler and it still does work, I was over correcting and adding too many unneeded checks

PR #2379, opened 2026-09-18
Hide chem text entry beta feature for staff on Ada

axlewin:


src/app/components/elements/panels/UserBetaFeatures.tsx
```diff
@@ -44,7 +44,7 @@ export const UserBetaFeatures = ({ displaySettings, setDisplaySettings, consentS
                 <p id="revision-helptext">{`This feature lets you answer questions ${siteSpecific("that you have answered before, without seeing your old answer.", "again, even if you've answered them before.")} It's useful if you are reviewing a topic before a test or exam.`}</p>
             </WithLinkableSetting>
 
-            {(isPhy || isStaff(user)) && <WithLinkableSetting id={"show-chem-text-entry"}>
+            {isPhy && <WithLinkableSetting id={"show-chem-text-entry"}>
```
jacbn, line 47: I think using `isAdmin` would be *my* preferred option, but I do have a bad habit of testing Ada with Sci content and it probably shouldn't be up to me. The use cache is niche enough that this seems fine.

jacbn (approved):

PR #2381, opened 2026-09-21
Unhide links to first batch of DCMS content

jacbn:
Updates the links to the first set of DCMS content (Cyber bits) and removes the "Coming soon" tags.


src/app/components/elements/cards/IconCard.tsx
```diff
@@ -14,6 +15,7 @@ export interface IconCardContentProps {
     clickUrl?: string;
     onButtonClick?: () => void;
     buttonText?: string;
+    includeTitleInAriaLabel?: boolean; // if true, will tie the link / button to the title via aria-describedby; use if the provided button text is accessible by itself
```
jsharkey13:
```suggestion
    includeTitleInAriaLabel?: boolean; // if true, will tie the link / button to the title via aria-describedby; use if the provided button text is not accessible by itself
```
?

PR #2384, opened 2026-09-21
Don't reset concept filters to subject tag on search

axlewin:
Fixes a bug where searching for e.g "force" on the A-Level Maths concept finder yielded no results until "mechanics" was explicitly selected, since the change in search term was triggering this useEffect which unnecessarily applied the maths tag and hid the specific physics questions that we want to appear in this question finder.


jacbn (approved):
LGTM, can't get this to break elsewhere.

PR #2395, opened 2026-09-23
Simplify and refine Ada related content lists

sjd210:
- Rather than define separate blocks for `<lg` and `>=lg` devices, use one block with different `className`s to reduce duplicated code.
- Re-align headings to be at the same level as their list items
- Stop displaying a heading for a section with no content (e.g. no empty "Outside your specification" section for those using All Stages/All Exam Boards)
- Display the "On your specification" heading on `<lg` screens (Looking back, I'm not sure why we didn't in the first place? This seems much clearer to me)


src/app/components/elements/RelatedContent.tsx
```diff
@@ -69,32 +70,19 @@ function renderQuestions(audienceQuestions: ContentSummaryDTO[], remainingQuesti
                     </Col>
                 </Row>
                 <hr/>
-                {/* Large devices - multi column */}
-                <div className="d-none d-lg-flex text-start">
-                    <div className="w-50">
-                        <div className="related-question-header h4">On your specification:</div>
+                <div className="d-lg-flex text-start">
```
barna-isaac: I'm glad you didn't rely on Bootstrap's column-row helpers, and would just like to point out, for the future, that I don't think we need the mediaquery-based `d-lg-flex` class either, and could just use `flex-wrap`. Probably something we should make a decision on together, depending on whether we'd like layout shifts to happen "as needed", or tied to breakpoints. Just a thought, not something that needs to change for this PR!

barna-isaac (approved):

barna-isaac:
The headings align much nicer with the content now, and I also agree the page looks better without headings displaying for empty sections. I'm also very happy it's now the same code regardless of screen size, thanks for this. I approve.

Here are the pages I used for testing: 
- https://adacomputerscience.org/concepts/hard_input?topic=hardware, to check that the "outside your specification" column disappears when empty 
- https://adacomputerscience.org/questions/_manual_regression_test, to see what it looks like on a question
- https://adacomputerscience.org/concepts/sql_dql, with a GCSE context, to see what it looks like when there's related content on as well as outside specification.

src/app/components/elements/RelatedContent.tsx
sjd210: Well technically we'd need `d-flex flex-wrap`. Just `flex-wrap` in a non-flex environment has no effect.

This case is already tied to breakpoints anyway per the several other classes below (e.g. `w-lg-50`), so I see no need to change specifically this top level, and regardless this element is only used in one context so there's not really any harm in doing so (in fact, having most elements shift around at consistent width-boundaries is quite useful for maintaining clear layouts, rather than possibly having to deal with a change at every pixel).

That said, for elements that _are_ used in a variety of contexts, I'm quite keen to move towards a `@container`-based approach (e.g. [cs-card](https://github.com/isaacphysics/isaac-react-app/blob/c4949f3860308a48de289841a2bd51b846724ca4/src/scss/cs/cards.scss#L42)) to have this "as needed" shift depending on the size of the element itself. ALVIs are in particular need, which we've got a card for already.

